PR #65, opened 2024-06-28
python(performance): ingestion request buffering + low hanging fruit

solidiquis:
## Context

It is very easy for folks to accidentally ingest single flows one-by-one in a hot-loop which interweaves serialization with streaming. This PR introduces `buffered_ingestion` to automate request buffering, only sending requests once the threshold is met. Example of how it is used:

```python
with ingestion_service.buffered_ingestion() as buffered_ingestion:
    buffered_ingestion.try_ingest_flows(*lots_of_flows)
    buffered_ingestion.try_ingest_flows(*lots_more_flows)
```

Contrast this with regular ingestion:
```python
ingestion_service.try_ingest_flows(*lots_of_flows)
ingestion_service.try_ingest_flows(*lots_more_flows)
```

Exiting the `with-block` will cause the remaining requests to flush, if there are any remaining. Users can manually flush as well.

This PR also caches the `IngestServiceStub`. Previously is was re-initializing it each time, but with the same persistent gRPC transport channel.

## Verification

- [Unit test](https://github.com/sift-stack/sift/pull/65/files#diff-f61dd4ad955b3436c1b1fa78bd83fd62b5b7302482baf58fb7ada734eb49cc93)

The following are benchmarks demonstrating the performance differences ingesting 50K data points to the Sift API. The benchmarks essentially take the following form:

```python
# Regular ingestion
for _ in range(50_000):
    ingestion_service.try_ingest_flows(...)

# Buffered ingestion
with ingestion_service.buffered_ingestion() as buffered_ingestion:
    for _ in range(50_000):
        buffered_ingestion.try_ingest_flows(...)

```

Screenshot 2024-06-27 at 3 34 24 PM


zachAtSiftStack (approved):
Soft review to get this moving out the door. LGTM.

philipwilcox-azimuth:
So for now this isn't async at all, just buffering to avoid network round trips between serialization calls?

solidiquis:
@philipwilcox-azimuth exactly. There will be support for async in the future, but for now this eliminates the create-and-ingest a single request in a hot-loop foot-gun.

PR #70, opened 2024-07-03
python(feature): automate gRPC retries

solidiquis:
## Changes

This PR adds support for automated retries using gRPC's retry mechanism should an RPC fail and return any of the following status codes:
- `UNKNOWN`
- `UNAVAILABLE`
- `ABORTED`
- `DEADLINE EXCEEDED`

These status codes were chosen on the basis that they are not explicitly set by Sift and often occur due to some external factor. Here is the retry policy that is introduced in this PR:

```python
  DEFAULT_POLICY: RetryConfig = {
      "methodConfig": [
          {
              # We can configure this on a per-service and RPC basis but for now we'll
              # apply this across all services and RPCs.
              "name": [{}],
              "retryPolicy": {
                  # gRPC does not allow more than 5 attempts
                  "maxAttempts": 5,
                  "initialBackoff": "0.1s",
                  "maxBackoff": "1s",
                  "backoffMultiplier": 2,
                  "retryableStatusCodes": [
                      StatusCode.UNKNOWN.name,
                      StatusCode.UNAVAILABLE.name,
                      StatusCode.ABORTED.name,
                      StatusCode.DEADLINE_EXCEEDED.name,
                  ],
              },
          }
      ]
  }
```

Users may be given the ability to customize retry policies in the future, but for now this is kept internal. The following is a link to the retry policy schema for the reviewer's convenience:

- [Retry policy schema](https://github.com/grpc/grpc-proto/blob/ec30f589e2519d595688b9a42f88a91bdd6b733f/grpc/service_config/service_config.proto#L136)

## Verification

The unit test in `python/lib/sift_py/grpc/transport_test.py` spins up a local test server that returns `UNKNOWN` error codes, recording number of failed attempts and ensuring successful attempts.

### DIFF
--- a/python/lib/sift_py/grpc/_retry.py
+++ b/python/lib/sift_py/grpc/_retry.py
@@ -28,9 +28,9 @@ class RetryPolicy:
                 "retryPolicy": {
                     # gRPC does not allow more than 5 attempts
                     "maxAttempts": 5,
-                    "initialBackoff": "0.1s",
-                    "maxBackoff": "1s",
-                    "backoffMultiplier": 2,
+                    "initialBackoff": "0.05s",
+                    "maxBackoff": "5s",
+                    "backoffMultiplier": 4,
                     "retryableStatusCodes": [
                         StatusCode.UNKNOWN.name,
                         StatusCode.UNAVAILABLE.name,
--- a/python/lib/sift_py/grpc/_retry.py
+++ b/python/lib/sift_py/grpc/_retry.py
@@ -1,0 +1,70 @@
+from __future__ import annotations
+
+import json
+from typing import Dict, List, TypedDict
+
+from grpc import StatusCode
+from typing_extensions import Self
+
+
+class RetryPolicy:
+    """
+    Retry policy meant to be used for `sift_py.grpc.transport.SiftChannel`. Users may have the ability to configure their own
+    custom retry policy in the future, but for now this is primarily intended for internal use.
+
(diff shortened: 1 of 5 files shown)


python/lib/sift_py/grpc/transport_test.py
```diff
@@ -0,0 +1,170 @@
+# ruff: noqa: N802
```
solidiquis, line 1: `ruff` doesn't like PascalCase for method names but that's coming from protobuf gen. This just ignores that particular rule.

python/lib/sift_py/grpc/_retry.py
```diff
@@ -0,0 +1,70 @@
...
+            {
+                # We can configure this on a per-service and RPC basis but for now we'll
+                # apply this across all services and RPCs.
+                "name": [{}],
+                "retryPolicy": {
+                    # gRPC does not allow more than 5 attempts
+                    "maxAttempts": 5,
+                    "initialBackoff": "0.1s",
+                    "maxBackoff": "1s",
+                    "backoffMultiplier": 2,
```
philipwilcox-azimuth, line 33: With a max of 5 retries we'll do:
0.1
0.2
0.4
0.8
1.0

for a total amount of time between first failure and last attempt of 2.5s by my count. I might bump this so we start earlier but also up a bit faster to give things more time to recover before that last one?

Like: 0.05s and 4x with a max of 5s:
0.05
0.20
0.80
3.2s
5s

which should give more like 9.25s before the final one, while still trying quickly for the first two.

solidiquis: yeah that sounds better. just updated

PR #72, opened 2024-07-05
python(feature): downloading telemetry

solidiquis:
Changes:

- Intoduces `sift_py.data` which contains tools to download telemetry for both regular and calculated channels.
- Adds support for `asyncio`. While `sift_py.data` is the only module using `async` at the moment, adding further async support for things such as ingestion is planned for the future.


zachAtSiftStack (approved):
Approval to get out the door.

PR #77, opened 2024-07-16
python(dependencies): extend dependency compatibility

solidiquis:
## Changes
- Lower minimum version requirements for some dependencies
- Implement gRPC synchronous interceptors in-house and removed `grpc-interceptor` dependency.
- Removed `aio` dependency as it's no longer in use.


zachAtSiftStack (approved):
Approving

zachAtSiftStack (approved):
Approving to get out the door.

PR #78, opened 2024-07-17
add pydantic models for the sift grafana plugin queries

bgfreeman7:
This adds pydantic models for the json queries in the Sift Grafana plugin.

### DIFF
--- a/python/lib/sift_grafana/sift_query_model.py
+++ b/python/lib/sift_grafana/sift_query_model.py
@@ -1,34 +1,37 @@
-from typing import Any, List, Optional
+from typing import List, Optional
+
+from pydantic import BaseModel, Field, model_validator
 from typing_extensions import Self
-from pydantic import BaseModel, Field,  model_validator
+
 
 # These are the models that are used to validate the input data for the Sift Grafana Plugin queries
 class Channel(BaseModel):
-    channelId: Optional[str] = None
-    channelIdentifier: Optional[str] = None
-    
-    @model_validator(mode='after')
-    def check_passwords_match(self) -> Self:
-        channelId = self.channelId
-        channelIdentifier = self.channelIdentifier
-        if channelId is None and channelIdentifier is None:
-            raise ValueError('channelId or channelIdentifier must be provided')
-        return self
-    
-class Asset(BaseModel):
-    assetId: Optional[str] = None
-    assetName: Optional[str] = None
+    channelId: Optional[str] = None # noqa: N815
+    channelIdentifier: Optional[str] = None # noqa: N815
 
     @model_validator(mode='after')
     def check_passwords_match(self) -> Self:
-        assetId = self.assetId
-        assetName = self.assetName
-        if assetId is None and assetName is None:
+        channel_id = self.channelId
+        channel_identifier = self.channelIdentifier
+        if channel_id is None and channel_identifier is None:
+            raise ValueError('channelId or channelIdentifier must be provided')
+        return self
+
+class Asset(BaseModel):
+    assetId: Optional[str] = None # noqa: N815
+    assetName: Optional[str] = None # noqa: N815
+
+    @model_validator(mode='after')
+    def check_passwords_match(self) -> Self:
+        asset_id = self.assetId
+        asset_name = self.assetName
+        if asset_id is None and asset_name is None:
             raise ValueError('assetId or assetName must be provided')
         return self
 
-class ChannelReference(Channel):    
-    channelReference: str = Field(..., pattern=r'^\$\d+$')  # Enforce format "${number}"
+class ChannelReference(Channel):
+     # Enforce format "${number}"
+    channelReference: str = Field(..., pattern=r'^\$\d+$')  # noqa: N815
 
 class AssetChannelQuery(Asset, Channel):
     pass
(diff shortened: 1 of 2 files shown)


solidiquis:
@bgfreeman7 did you happen to run both the linter and formatter? Seems like CI is just failing on linting.

```
$ source venv/bin/activate
$ dev lint-fix
$ dev fmt
```

python/lib/sift_grafana/sift_query_model.py
```diff
@@ -0,0 +1,56 @@
+from typing import List, Optional
+
+from pydantic import BaseModel, Field, model_validator
+from typing_extensions import Self
+
+
+# These are the models that are used to validate the input data for the Sift Grafana Plugin queries
```
solidiquis, line 7: Are these internal doc-comments or are they meant for the public? If they're meant for users, could you format it like in this [example](https://github.com/sift-stack/sift/blob/735ddca3ea26e22142c29518b5c32d819b968c5a/python/lib/sift_py/ingestion/service.py#L21)?

If just internal, feel free to just resolve this comment.

python/lib/sift_grafana/sift_query_model.py
```diff
@@ -0,0 +1,56 @@
...
+class Channel(BaseModel):
+    channelId: Optional[str] = None # noqa: N815
```
solidiquis, line 9: Could you change this to:

```python
class Channel(BaseModel):
    channel_id: Optional[str]
    channel_identifier: Optional[str]
    
    def __init__(self, channel_id: Optional[str] = None, channel_identifier: Optional[str] = None):
        ...
```

python/lib/sift_grafana/sift_query_model.py
```diff
@@ -0,0 +1,56 @@
...
+    channelIdentifier: Optional[str] = None # noqa: N815
+
+    @model_validator(mode='after')
+    def check_passwords_match(self) -> Self:
+        channel_id = self.channelId
+        channel_identifier = self.channelIdentifier
```
solidiquis, line 15: Not sure if there's a particular reason for camelCase, but if there isn't a particular requirement could we keep vars as snake_case?

python/lib/sift_grafana/sift_query_model.py
```diff
@@ -0,0 +1,56 @@
...
+        if channel_id is None and channel_identifier is None:
+            raise ValueError('channelId or channelIdentifier must be provided')
+        return self
+
+class Asset(BaseModel):
+    assetId: Optional[str] = None # noqa: N815
```
solidiquis, line 21: Same as comment above for `class Channel(BaseModel):`

python/lib/sift_grafana/sift_query_model.py
```diff
@@ -0,0 +1,56 @@
...
+
+class CalculatedChannelQuery(BaseModel):
+    name: str
+    asset: Asset
+    channelReferences: List[ChannelReference] # noqa: N815
+    expression: str
+
+class SiftQuery(BaseModel):
+    queries: List[AssetChannelQuery]
+    calculatedChannelQuery: Optional[CalculatedChannelQuery] = None # noqa: N815
```
solidiquis, line 47: Same as above. Good practice to define an initializer and do default args there :)

python/lib/sift_grafana/sift_query_model.py
```diff
@@ -0,0 +1,56 @@
+from typing import List, Optional
+
+from pydantic import BaseModel, Field, model_validator
```
solidiquis, line 3: This will also need to be added to `pyproject.toml` [here](https://github.com/sift-stack/sift/blob/735ddca3ea26e22142c29518b5c32d819b968c5a/python/pyproject.toml#L23).

You can add `pydantic~=2.0.0` which is short hand for anything between 2.0.0 and 3.0.0.

If `2.0.0` breaks for whatever reason try `2.1.0`, `2.2.0`, and so on.

solidiquis (changes requested):
Requested some changes but pretty straight-forward otherwise! Few things of note:
1. We should prefer `snake_case` for variables/functions in Python
2. Generally good practice to define `__init__` in classes and do default arguments there instead of in attribute declaration.
3. You'll need to add the `sift_grafana` module to the `pyproject.toml` [here](https://github.com/sift-stack/sift/blob/735ddca3ea26e22142c29518b5c32d819b968c5a/python/pyproject.toml#L17).

python/lib/sift_grafana/sift_query_model.py
bgfreeman7: These variables match the json in the schema, which are camelCase. I've tried to keep everything that was python only to the snake case

python/lib/sift_grafana/sift_query_model.py
bgfreeman7: Yeah that makes sense

python/lib/sift_grafana/sift_query_model.py
bgfreeman7: Unfortunately the main use case with pydantic validation doesn't seem to play nicely with the init style.

`sift_query = SiftQuery.model_validate(from_json(sift_query_json, allow_partial=True))` 

Definitely open to other ways to maintain the standard though

solidiquis (approved):

PR #82, opened 2024-08-15
feature(protos): new protos and various protos updates

solidiquis:
## Changes:
New protocol buffers and updates to existing protocol buffers, most notably:
- New saved searches protos to save search configurations on the home/overview page
- New remote files protos for programmatic file attachments
- Updates to rules proto for rule versioning


zachAtSiftStack (approved):
Approving

PR #87, opened 2024-08-19
python(enhancement): uri validation

solidiquis:
## Changes

Common issue where users provide a malformed URI to the Sift API. This PR adds functionality to accept various forms of the URI to prevent obscure DNS errors that result from malformed URIs.


zachAtSiftStack (approved):
Approving

PR #88, opened 2024-08-23
python(feature): timeout-based flushing for buffered ingestion

solidiquis:
## Changes

This PR adds support for timeout-based flushing for buffered ingestion. Users can now configure an interval in which a buffer will get flushed regardless of whether or not the buffer has been filled. If a buffer is flushed due to being filled, the timer gets restarted.

Additionally, this PR introduces the ability to customize `__exit__` behavior when using the buffered ingestion API in a context manager when there is an error.


zachAtSiftStack (commented):
Approving.

zachAtSiftStack (approved):
Approving

PR #89, opened 2024-09-03
python(feature): generate channel value from channel config

solidiquis:
## Changes

This PR adds the ability to generate a channel value directly from a channel config based on data-type.


zachAtSiftStack (approved):
Approving

PR #92, opened 2024-09-06
python(fix): add missing double channel computation

solidiquis:
## Changes

A check for the double channel data type was omitted from method that produced a channel value from a channel config.


niliayu (approved):
lgtm!

PR #96, opened 2024-09-16
python(chore): csv example ingestion

niliayu:
Adds an example of how to ingest data from a CSV file. This example is simple and makes a few assumptions (e.g. channels are all doubles), but hopefully is easy enough to understand an expand from for other uses.


solidiquis (changes requested):
Some small changes! There were some things I did in the original example that aren't suitable to show to users in official examples so we'll get those cleaned up :)

python/examples/ingestion_with_csv/main.py
```diff
@@ -0,0 +1,108 @@
...
+
+from dotenv import load_dotenv
+from sift_py.grpc.transport import SiftChannelConfig, use_sift_channel
+from sift_py.ingestion.channel import ChannelConfig, ChannelDataType, double_value
+from sift_py.ingestion.config.telemetry import TelemetryConfig
+from sift_py.ingestion.flow import FlowConfig, FlowOrderedChannelValues
+from sift_py.ingestion.service import IngestionService
+
+
+def _parse_csv(
```
solidiquis, line 16: nit: could you remove the `_` prefix?

python/examples/ingestion_with_csv/main.py
```diff
@@ -0,0 +1,108 @@
...
+                    "flow_name": flow_name,
+                    "timestamp": datetime.fromisoformat(timestamp_str),
+                    "channel_values": [double_value(float(raw_value)) for raw_value in values],
+                }
+            )
+
+    return flows
+
+
+def _load_telemetry_config(
```
solidiquis, line 45: Same here, could you remove the `_`?

solidiquis (changes requested):
Just a couple tiny nits and one final request: Would you mind truncating the CSV to just have 50 rows? The CSV as is is quite large and it'll blow up our repo size. 50 rows will get the same point across :)

python/examples/ingestion_with_csv/main.py
```diff
@@ -0,0 +1,108 @@
...
+
+    telemetry_config = _load_telemetry_config(sample_data_csv, asset_name, ingestion_client_key)
+    flows = _parse_csv(sample_data_csv, telemetry_config)
+
+    sift_channel_config = SiftChannelConfig(
+        uri=sift_uri,
+        apikey=apikey,
+    )
+
+    with use_sift_channel(sift_channel_config) as channel:
```
solidiquis, line 99: Oh actually it looks like we're also missing the actual ingestion step. You'll need to send the flows you created from the `parse_csv` function to `ingest_flows` method (I think that's what it's called). The docs should have that info :)

niliayu: [02e89ec](https://github.com/sift-stack/sift/pull/96/commits/02e89ec3b790cb788dcbdcfa51ecb9c6a40deb64) did I get that right? Since we're already in a context manager with the `with` that will clean things up for us when the ingest is done?

solidiquis: You will also have to use buffered ingestion within a `with` block as well, I think?

niliayu: Like so? [3e18495](https://github.com/sift-stack/sift/pull/96/commits/3e18495ea16e1fb840113ca1636600abf40aec4b)

solidiquis: Almost! I think you should be able to do `buffered_ingestion.try_ingest_flows(*flows)`. I would double check the function signature.

niliayu: Oh yeah, I completely missed that! Looks like that's the case 👍

niliayu: Thanks, updated! Will go ahead a re-request review

solidiquis (approved):
LGTM! Thanks for iterating with me!

niliayu:
Thanks for the quick feedback, super helpful in learning the patterns here!

PR #98, opened 2024-09-18
python(chore): Add multiple CSV ingestion example

niliayu:
Adds an example that processes channels from across multiple CSVs and ingests the data. Checks to make sure there is data present for across all channels for each timestamp and otherwise fills in an empty_value.


python/examples/ingestion_with_csv/multi_csv/main.py
```diff
@@ -0,0 +1,124 @@
...
+
+def parse_csv(
+    data: List[Dict], telemetry_config: TelemetryConfig
+) -> List[FlowOrderedChannelValues]:
+    flows: List[FlowOrderedChannelValues] = []
+
+    flow = telemetry_config.flows[0]  # Packed into a single flow for this example
+    flow_name = flow.name
+
+    all_timestamps: List = []
```
solidiquis, line 23: You could probably remove the type annotation here OR provide the type parameter to `List[T]` just to be more clear.

python/examples/ingestion_with_csv/multi_csv/main.py
```diff
@@ -0,0 +1,124 @@
...
+    for channel in data:  # Extract all timestamps from all channels
+        all_timestamps += channel.keys()
```
solidiquis, line 25: You could also just get away with `all_timestamps = [ts for ts in channel.keys()`]?

solidiquis (changes requested):
Couple small things :)

python/examples/ingestion_with_csv/multi_csv/main.py
niliayu: Ah yeah, I threw that in to get mypy to stop yelling at me... I'll clean this up

python/examples/ingestion_with_csv/multi_csv/main.py
niliayu: Yeah this would be nicer with a comprehension, I like it 👍

solidiquis (approved):
LGTM

PR #100, opened 2024-09-19
feat: Asset agnostic rule configs

niliayu:
This PR adds the ability to define rules in YAML separate from the main telemetry config, so that rules may be referenced and reused in the context of multiple assets.

A rule that's defined in a namespace can be referenced in the main telemetry config like so:
```
  - namespace: voltage
     name: undervoltage
     channel_references:
       - $1: *vehicle_state_channel
       - $2: *voltage_channel
```
With the corresponding rule defined in a separate YAML like so:
```
namespace: voltage
   - name: undervoltage
     description: Checks for undervoltage while accelerating
     expression: $1 == "Accelerating" && $2 < 40
     type: review
```
Or if the user wants to instantiate rules in python directly, the `load_rule_namespaces` function has been added that will allow them to provide a path to the directory containing their rule YAMLs and load them into memory:
```
   rule_namespaces = load_rule_namespaces(
         [
             RULE_NAMESPACES_DIR,
         ]
     )

    RuleConfig(
                 name="overvoltage",
                 namespace="voltage",
                 namespace_rules=rule_namespaces,
                 channel_references=[
                     # INFO: Can use either "channel_identifier" or "channel_config"
                     {
                         "channel_reference": "$1",
                         "channel_identifier": vehicle_state_channel.fqn(),
                     },
                     {
                         "channel_reference": "$2",
                         "channel_config": voltage_channel,
                     },
                 ],
             ),
```

You can see examples of this rules in dev in this run: https://dev.app.siftstack.com/share/3xBUAIQ

### DIFF
--- a/python/examples/ingestion_with_python_config/rule_modules/velocity.yml
+++ b/python/examples/ingestion_with_python_config/rule_modules/velocity.yml
@@ -1,0 +1,12 @@
+namespace: velocity
+
+rules:
+  - name: vehicle_stuck
+    description: Checks that vehicle velocity becomes nonzero 5s after entering accelerating state
+    expression: $1 == "Accelerating" && persistence($2 == 0, 5)
+    type: review
+
+  - name: vehicle_not_stopped
+    description: Makes sure vehicle velocity remains 0 while stopped
+    expression: $1 == "Stopped" && $2 > 0
+    type: review
--- a/python/examples/ingestion_with_python_config/rule_modules/velocity.yml
+++ b/python/examples/ingestion_with_python_config/rule_modules/velocity.yml
@@ -2,11 +2,11 @@ namespace: velocity
 
 rules:
   - name: vehicle_stuck
-    description: Checks that vehicle velocity becomes nonzero 5s after entering accelerating state
+    description: Triggers if the vehicle velocity is not 0 for 5s after entering accelerating state
     expression: $1 == "Accelerating" && persistence($2 == 0, 5)
     type: review
 
   - name: vehicle_not_stopped
-    description: Makes sure vehicle velocity remains 0 while stopped
+    description: Triggers if the vehicle velocity does not remain 0 while stopped
     expression: $1 == "Stopped" && $2 > 0
     type: review
--- a/python/examples/ingestion_with_yaml_config/telemetry_configs/nostromo_lv_426.yml
+++ b/python/examples/ingestion_with_yaml_config/telemetry_configs/nostromo_lv_426.yml
@@ -127,7 +127,7 @@ flows:
   - name: gpio_channel
     channels:
       - <<: *gpio_channel
-      
+
   - name: logs
     channels:
       - <<: *log_channel
--- a/python/examples/ingestion_with_yaml_config/telemetry_configs/nostromo_lv_426.yml
+++ b/python/examples/ingestion_with_yaml_config/telemetry_configs/nostromo_lv_426.yml
@@ -88,6 +88,11 @@ rules:
         - failure
         - nostromo
 
+  - name: voltage.overvoltage
+    channel_references:
+      - $1: *vehicle_state_channel
+      - $2: *voltage_channel
+ 
 flows:
   - name: readings
     channels:
--- a/python/examples/ingestion_with_yaml_config/telemetry_configs/nostromo_lv_426.yml
+++ b/python/examples/ingestion_with_yaml_config/telemetry_configs/nostromo_lv_426.yml
@@ -93,7 +93,25 @@ rules:
     channel_references:
       - $1: *vehicle_state_channel
       - $2: *voltage_channel
- 
+
+  - namespace: voltage
+    name: undervoltage
+    channel_references:
+      - $1: *vehicle_state_channel
+      - $2: *voltage_channel
+
+  - namespace: velocity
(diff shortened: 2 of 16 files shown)


python/lib/sift_py/ingestion/config/telemetry.py
```diff
@@ -179,6 +188,25 @@ def _from_yaml(
             )
 
         for rule in config_as_yaml.get("rules", []):
+            namespace = rule.get("namespace")
+
+            if namespace:
+                rule_namespace = rule_namespaces.get(str(namespace))
+                if not rule_namespace:
+                    # TODO: This is a placeholder.
```
niliayu, line 196: A placeholder because if this pattern makes sense we can add a more specific and helpful error.

python/examples/ingestion_with_yaml_config/telemetry_configs/nostromo_lv_426.yml
```diff
@@ -88,6 +88,30 @@ rules:
         - failure
         - nostromo
 
+  - namespace: voltage
```
niliayu, line 91: I ended up adding this `namespace` key because that made it obvious whether we should expect a fully formed rule or a reference to one in a separate namespace which just made validation a lot clearer, but definitely very open to feedback here

python/lib/sift_py/ingestion/config/yaml/load.py
```diff
@@ -275,11 +335,39 @@ def _validate_bit_field_element(val: Any):
...
+                "- channel_references",
+                f"List[Dict[str, {_type_fqn(ChannelConfigYamlSpec)}]]",
+                ["rules"],
+            )
+
+        for channel_reference in cast(List[Any], channel_references):
+            _validate_channel_reference(channel_reference)
+
+    if namespace:
+        # TODO: Probably should inform the user that the below keys
```
niliayu, line 367: Hopefully this makes sense -- I was thinking that if a user is trying to reference a rule in a namespace it would be nice to clearly tell them that the below keys should not be included in the top level yaml.

niliayu:
@marcsiftstack :
So one of the next things I was looking to do here was update the [python ingestion example](https://github.com/sift-stack/sift/blob/main/python/examples/ingestion_with_python_config/telemetry_config.py) to be able to read in the same rule yaml CSVs.  I made a [load_rule_namespaces](https://github.com/sift-stack/sift/blob/e8a094701c203bf9f0b46f2d24c736bb2c70b70a/python/lib/sift_py/ingestion/config/yaml/load.py#L56) function that's somewhat similar to the [load_named_expression_modules one used here](https://github.com/sift-stack/sift/blob/main/python/examples/ingestion_with_python_config/telemetry_config.py#L20), but I realized I put the logic to reassemble the rules from their namespaces in the yaml-specific [TelemetryConfig._from_yaml](https://github.com/sift-stack/sift/blob/e8a094701c203bf9f0b46f2d24c736bb2c70b70a/python/lib/sift_py/ingestion/config/telemetry.py#L211-L222). I basically set this up so that for the yaml example, the TelemetryConfig & RuleConfigs get reassembled from yaml to look exactly the same as if the rules had been defined in one flat yaml to begin with. So maybe for now that means we need to do some similar logic when creating the [RuleConfigs](https://github.com/sift-stack/sift/blob/main/python/examples/ingestion_with_python_config/telemetry_config.py#L108) for the example namespace rules, but I haven't had quite enough time to think about how best to make this look for now. I could definitely be missing something super simple here to make this work though!

python/lib/sift_py/ingestion/rule/config.py
```diff
@@ -26,19 +26,21 @@ class RuleConfig(AsJson):
...
         name: str,
-        description: str,
-        expression: str,
-        action: RuleAction,
         channel_references: List[
             Union[ExpressionChannelReference, ExpressionChannelReferenceChannelConfig]
         ],
+        description: str = "",
+        expression: str = "",
+        action: Optional[RuleAction] = None,
```
niliayu, line 40: Is there a better default value for this than `None`? Otherwise this is the only one that must be explicitly `Optional`.

solidiquis (approved):

python/examples/ingestion_with_python_config/rule_modules/velocity.yml
```diff
@@ -0,0 +1,12 @@
+namespace: velocity
+
+rules:
+  - name: vehicle_stuck
+    description: Checks that vehicle velocity becomes nonzero 5s after entering accelerating state
```
marcsiftstack, line 5: It seems like the wording doesn't match the check. Should it say: "Checks that vehicle velocity is zero 5s after entering accelerating state"?

Same for below:
"Checks if vehicle velocity greater than 0 while stopped"

niliayu: The rule logic is sort of "backwards", so the condition evaluating to true means that it failed. So my thinking was if the vehicle velocity doesn't become nonzero while in the accelerating state then it must be stuck, therefore the rule should be that a vehicle velocity of 0 for 5s while in the accelerating state means the vehicle is stuck. But you're technically right... it is checking exactly what you said, in order to make sure the opposite happens 🤷 Lemme clarify these a bit

marcsiftstack (approved):

PR #102, opened 2024-10-08
python(feat): Add CSV and TDMS upload service

marcsiftstack:
This PR adds a CsvUploadService and TdmsUploadService to the python client library to upload CSVs and TDMS files.

Output from running the example:
```
$ python main.py                       
Upload example complete!
```

Also modified main.py to test URL uploads by adding the following:
```
    url = "https://files.catbox.moe/hbrq2h.csv"
    status = csv_upload_service.upload_from_url(url, csv_config)
```
Output:
```
$ python main.py
Upload example complete!
```

Example of TDMS channels grouped in the UI and metadata in the description:
image


python/lib/sift_py/upload/config.py
```diff
@@ -0,0 +1,149 @@
...
+    "TIME_FORMAT_RELATIVE_NANOSECONDS",
+    "TIME_FORMAT_RELATIVE_MICROSECONDS",
+    "TIME_FORMAT_RELATIVE_MILLISECONDS",
+    "TIME_FORMAT_RELATIVE_SECONDS",
+    "TIME_FORMAT_RELATIVE_MINUTES",
+    "TIME_FORMAT_RELATIVE_HOURS",
+]
+
+
+VALID_DATA_TYPES = [
```
solidiquis, line 22: We have a `ChannelDataType` enum over in `lib/sift_py/ingestion/channel.py`. Any way we could use that instead of hardcoding these values?

python/lib/sift_py/upload/config.py
```diff
@@ -0,0 +1,149 @@
...
+    "CHANNEL_DATA_TYPE_ENUM",
+    "CHANNEL_DATA_TYPE_BIT_FIELD",
+]
+
+
+class _BaseModel(BaseModel):
+    model_config = ConfigDict(extra="forbid")
+
+
+class EnumType(_BaseModel):
```
solidiquis, line 40: We have `ChannelEnumType` over in `lib/sift_py/ingestion/channel.py`.

python/lib/sift_py/upload/config.py
```diff
@@ -0,0 +1,149 @@
...
+    key: int
+    name: str
+
+
+class BitFieldElement(_BaseModel):
```
solidiquis, line 45: We have `ChannelBitFieldElement` over in lib/sift_py/ingestion/channel.py`

python/lib/sift_py/upload/config.py
```diff
@@ -0,0 +1,149 @@
...
+                    "'relative_start_time' specified for non relative time format.",
+                )
+
+        return self
+
+
+class DataColumn(_BaseModel):
+    name: str
+    data_type: str
+    component: Optional[str] = ""
```
solidiquis, line 85: Having an optional default to a blank string (falsey) feels a little awkward and redundant. If we're going to default to non-`None` values could we get rid of the optional?

python/lib/sift_py/upload/config.py
```diff
@@ -0,0 +1,149 @@
...
+    time_column: TimeColumn
+    data_columns: Dict[int, DataColumn]
+
+    @model_validator(mode="after")
+    def validate_config(self):
+        if not self.data_columns:
+            raise PydanticCustomError("invalid_config_error", "Empty 'data_columns'")
+
+
+class CsvConfig:
```
solidiquis, line 140: Could we move this fella to the top of the file? It seems like the primary purpose of this module. If you `from __future__ import annotations` then python will allow it.

python/lib/sift_py/upload/status.py
```diff
@@ -0,0 +1,61 @@
...
+                if verbose:
+                    print("Upload failed")
+                return False
+            else:
+                raise Exception(f"Unknown status: {status}")
+            time.sleep(polling_interval)
+            polling_interval = min(polling_interval * 2, 60)
+
+    @staticmethod
+    def _compute_uri(restconf: SiftRestConfig) -> str:
```
solidiquis, line 51: Looks like we have this copy-pasta in a few places. Could we put this somewhere and have it be resusable?

python/lib/sift_py/upload/status.py
```diff
@@ -0,0 +1,61 @@
...
+            headers={"Authorization": f"Bearer {self._apikey}"},
+        )
+        response.raise_for_status()
+        return response.json().get("dataImport").get("status")
+
+    def wait(self, verbose: bool = False) -> bool:
+        polling_interval = 1
+        while True:
+            status = self.get_status()
+            if status == "DATA_IMPORT_STATUS_SUCCEEDED":
```
solidiquis, line 31: Would be nice to have these be python enums

python/lib/sift_py/upload/status.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def get_status(self):
+        response = requests.get(
+            url=f"{self._status_uri}/{self._data_import_id}",
+            headers={"Authorization": f"Bearer {self._apikey}"},
+        )
+        response.raise_for_status()
+        return response.json().get("dataImport").get("status")
+
+    def wait(self, verbose: bool = False) -> bool:
```
solidiquis, line 27: I may have been the one to suggest printing but I think we should change this a tad. Rather than printing I think the current implementation makes it quite easy for users to do this on their own. Rather than call it `wait`, I think something more descriptive like `wait_until_complete` or something like that would be good with no additional args other than `self`. Then inside the function we'll poll, and if the status is complete it would be nice if `wait` returned the upload_url and/or upload ID of the file that was uploaded.

python/lib/sift_py/upload/status.py
```diff
@@ -0,0 +1,61 @@
...
+        self._status_uri = urljoin(base_uri, self.STATUS_PATH)
+        self._apikey = restconf["apikey"]
+
+    def get_status(self):
+        response = requests.get(
+            url=f"{self._status_uri}/{self._data_import_id}",
+            headers={"Authorization": f"Bearer {self._apikey}"},
+        )
+        response.raise_for_status()
+        return response.json().get("dataImport").get("status")
```
solidiquis, line 25: As mentioned below, would be nice to convert this to a defined enum that has each member documented.

solidiquis (changes requested):
This is looking good! Left some comments to start. One concern I have about calling the new module `upload` is that it's a little to generic. The file attachments module technically does uploading. Ingestion is also technically uploading. Renaming the module to something like `data_import` is probably more inline with how this module will grow as we add in other file formats as well.

python/lib/sift_py/upload/config.py
niliayu: Is there any reason there isn't something like that for `TIME_FORMATS` too?

python/lib/sift_py/upload/csv.py
```diff
@@ -0,0 +1,149 @@
...
+        self,
+        path: Union[str, Path],
+        csv_config: CsvConfig,
+    ) -> DataImportStatus:
+        posix_path = Path(path) if isinstance(path, str) else path
+
+        if not posix_path.is_file():
+            raise Exception(f"Provided path, '{path}', does not point to a regular file.")
+
+        _, mimetype, content_encoding = self.__class__._mime_and_content_type_from_path(posix_path)
```
niliayu, line 40: Just a thought -- since we're getting the MIME, maybe it'd be nice to raise a warning if we detect a type that wouldn't play well with everything downstream of here? e.g. a pdf instead of a csv or something. Just a nit/thought though

python/examples/upload/csv/main.py
```diff
@@ -0,0 +1,64 @@
...
+    assert sift_uri, "expected 'SIFT_API_URI' environment variable to be set"
+
+    apikey = os.getenv("SIFT_API_KEY")
+    assert apikey, "expected 'SIFT_API_KEY' environment variable to be set"
+
+    asset_name = os.getenv("ASSET_NAME")
+    assert asset_name, "expected 'ASSET_NAME' environment variable to be set"
+
+    rest_config: SiftRestConfig = {
+        # Be sure to exclude the "https://" or "http://" scheme out of the uri
```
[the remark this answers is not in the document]
marcsiftstack: Copied this from the other examples: https://docs.siftstack.com/sift_py/sift_py/file_attachment.html

python/lib/sift_py/upload/config.py
marcsiftstack: I think the best path forward is to inherit from both BaseModel and ChannelEnumType, that way the ChannelEnumType class is the source of truth and we get the validation from BaseModel.

python/lib/sift_py/upload/config.py
marcsiftstack: See https://github.com/sift-stack/sift/pull/102#discussion_r1792820444

python/lib/sift_py/upload/config.py
marcsiftstack: Per our other discussion - these aren't generated. I will create the TimeFormat enum and we can always replace it with a protobuf generated one later if we want.

python/examples/upload/csv/main.py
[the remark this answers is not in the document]
niliayu: Ahh copy! Makes sense, keep consistency 👍

python/lib/sift_py/rest.py
```diff
@@ -14,3 +15,16 @@ class SiftRestConfig(TypedDict):
     uri: str
     apikey: str
     use_ssl: NotRequired[bool]
+
+
+def compute_uri(restconf: SiftRestConfig) -> str:
+    uri = restconf["uri"]
+
+    scheme_match = re.match(r"(.+://).+", uri)
```
marcsiftstack, line 23: I made this change so that this would work with localhost.

python/lib/sift_py/data_import/_config_test.py
```diff
@@ -0,0 +1,232 @@
+import pytest
+from sift_py.data_import.config import CsvConfig
+
+
+def test_empty_data_columns():
+    with pytest.raises(Exception, match="Empty 'data_columns'"):
+        CsvConfig(
+            {
+                "asset_name": "test_asset",
```
niliayu, line 9: Super small nit: You could consider making a fixture for this just to save a bit of copy-pasta, and then editing the `data_columns` or other properties as needed to trigger different raises and such. Could just make it a bit easier to update down the line if we change the shape of or add fields to `CsvConfig`

niliayu (commented):
Just one test nit I thought of, otherwise looks really nice to me! Will let Benji have the final say though since I'm still onboarding to a lot of the details here.

python/lib/sift_py/data_import/_config_test.py
marcsiftstack: Good call. Will update this since it makes it easier to see what changed.

solidiquis (changes requested):

python/lib/sift_py/rest.py
```diff
@@ -14,3 +15,16 @@ class SiftRestConfig(TypedDict):
...
     apikey: str
     use_ssl: NotRequired[bool]
+
+
+def compute_uri(restconf: SiftRestConfig) -> str:
+    uri = restconf["uri"]
+
+    scheme_match = re.match(r"^(\w+://).+", uri)
+    if scheme_match:
+        raise Exception(f"The URL scheme '{scheme_match.groups()[0]}' should not be included")
```
solidiquis, line 25: There is a `_clean_uri` util from `lib/sift_py/grpc/transport.py` that we extract and put into a top level internal util so folks can include `http://` or `https://` if they want. Folks always miss this detail and they get really funky error messages that isn't immediately obvious for anyone what went wrong.

python/lib/sift_py/data_import/csv.py
```diff
@@ -0,0 +1,198 @@
...
+        except KeyError as e:
+            raise Exception(f"Response missing required keys: {e}")
+
+        return DataImportStatus(self._rest_conf, data_import_id)
+
+    def simple_upload(
+        self,
+        asset_name: str,
+        path: Union[str, Path],
+        run_name: Optional[str] = None,
```
solidiquis, line 129: For args that default to `None` we should make them the last args in the function signature so they can omit them when calling, otherwise they'll have to explicitly pass in `None`.

python/lib/sift_py/data_import/status.py
```diff
@@ -0,0 +1,68 @@
...
+        response.raise_for_status()
+
+        status_text = response.json().get("dataImport").get("status")
+        status = DataImportStatusType.from_str(status_text)
+        if status is None:
+            raise Exception(f"Unknown data import status: {status_text}")
+
+        return status
+
+    def wait_until_complete(self) -> bool:
```
solidiquis, line 53: Instead returning a `bool` is it possible to return an `Optional[T]` where T is the response object from the endpoint that gives us the metadata about the uploaded file?

python/examples/data_import/csv/simple/main.py
```diff
@@ -0,0 +1,32 @@
...
+    assert sift_uri, "expected 'SIFT_API_URI' environment variable to be set"
+
+    apikey = os.getenv("SIFT_API_KEY")
+    assert apikey, "expected 'SIFT_API_KEY' environment variable to be set"
+
+    asset_name = os.getenv("ASSET_NAME")
+    assert asset_name, "expected 'ASSET_NAME' environment variable to be set"
+
+    rest_config: SiftRestConfig = {
+        # Be sure to exclude the "https://" or "http://" scheme out of the uri
```
solidiquis, line 24: we can remove this if we implement my other comment about `_clean_uri` from the `lib/sift_py/grpc/transport.py` module.

python/lib/sift_py/data_import/time_format.py
```diff
@@ -0,0 +1,83 @@
+from enum import Enum
+
+
+class TimeFormatType(Enum):
+    ABSOLUTE_RFC339 = "TIME_FORMAT_ABSOLUTE_RFC3339"
```
solidiquis, line 5: type: this should be RFC3339

python/lib/sift_py/data_import/time_format.py
```diff
@@ -0,0 +1,83 @@
...
+    ABSOLUTE_UNIX_NANOSECONDS = "TIME_FORMAT_ABSOLUTE_UNIX_NANOSECONDS"
+    RELATIVE_NANOSECONDS = "TIME_FORMAT_RELATIVE_NANOSECONDS"
+    RELATIVE_MICROSECONDS = "TIME_FORMAT_RELATIVE_MICROSECONDS"
+    RELATIVE_MILLISECONDS = "TIME_FORMAT_RELATIVE_MILLISECONDS"
+    RELATIVE_SECONDS = "TIME_FORMAT_RELATIVE_SECONDS"
+    RELATIVE_MINUTES = "TIME_FORMAT_RELATIVE_MINUTES"
+    RELATIVE_HOURS = "TIME_FORMAT_RELATIVE_HOURS"
+
+    @classmethod
+    def from_str(cls, val: str):
```
solidiquis, line 19: I don't know why `mypy` is complaining.. I think it's just confused. Simplifying the logic to the following fixed  `mypy` for me though:

```python
from enum import Enum
from typing import Optional
from typing_extensions import Self


class TimeFormatType(Enum):
    ABSOLUTE_RFC3339 = "TIME_FORMAT_ABSOLUTE_RFC3339"
    ABSOLUTE_DATETIME = "TIME_FORMAT_ABSOLUTE_DATETIME"
    ABSOLUTE_UNIX_SECONDS = "TIME_FORMAT_ABSOLUTE_UNIX_SECONDS"
    ABSOLUTE_UNIX_MILLISECONDS = "TIME_FORMAT_ABSOLUTE_UNIX_MILLISECONDS"
    ABSOLUTE_UNIX_MICROSECONDS = "TIME_FORMAT_ABSOLUTE_UNIX_MICROSECONDS"
    ABSOLUTE_UNIX_NANOSECONDS = "TIME_FORMAT_ABSOLUTE_UNIX_NANOSECONDS"
    RELATIVE_NANOSECONDS = "TIME_FORMAT_RELATIVE_NANOSECONDS"
    RELATIVE_MICROSECONDS = "TIME_FORMAT_RELATIVE_MICROSECONDS"
    RELATIVE_MILLISECONDS = "TIME_FORMAT_RELATIVE_MILLISECONDS"
    RELATIVE_SECONDS = "TIME_FORMAT_RELATIVE_SECONDS"
    RELATIVE_MINUTES = "TIME_FORMAT_RELATIVE_MINUTES"
    RELATIVE_HOURS = "TIME_FORMAT_RELATIVE_HOURS"

    @classmethod
    def from_str(cls, val: str) -> Optional[Self]:
        try:
            return cls(val)
        except ValueError:
            return None

    def as_human_str(self) -> str:
        return self.value

    def is_relative(self):
        return self in [
            self.RELATIVE_NANOSECONDS,
            self.RELATIVE_MICROSECONDS,
            self.RELATIVE_MILLISECONDS,
            self.RELATIVE_SECONDS,
            self.RELATIVE_MINUTES,
            self.RELATIVE_HOURS,
        ]
```

We'll need to do the same thing in the other places `mypy` is complaining about.

solidiquis (changes requested):
Getting there! Feel free to start dropping doc-comments

python/lib/sift_py/ingestion/channel.py
```diff
@@ -225,70 +256,73 @@ def from_pb(cls, val: channel_pb.ChannelDataType.ValueType) -> "ChannelDataType"
...
-        elif self == self.__class__.INT_64.value:
-            return ChannelDataTypeStrRep.INT_64.value
-        elif self == self.__class__.UINT_32.value:
-            return ChannelDataTypeStrRep.UINT_32.value
-        elif self == self.__class__.UINT_64.value:
-            return ChannelDataTypeStrRep.UINT_64.value
+        else:
+            raise Exception("Unreachable")
+
+    def as_human_str(self, api_format: bool = False) -> str:
```
marcsiftstack, line 293: This recommendation (https://github.com/sift-stack/sift/pull/102#discussion_r1801562128) worked for the other types but not `ChannelDataType` since it uses `ChannelDataTypeStrRep`.

python/examples/data_import/tdms/main.py
```diff
@@ -0,0 +1,31 @@
...
+    asset_name = os.getenv("ASSET_NAME")
+    assert asset_name, "expected 'ASSET_NAME' environment variable to be set"
+
+    rest_config: SiftRestConfig = {
+        "uri": sift_uri,
+        "apikey": apikey,
+    }
+
+    tdms_upload_service = TdmsUploadService(rest_config)
+    status = tdms_upload_service.upload("sample_data.tdms", asset_name, group_into_components=True)
```
solidiquis, line 30: Nice and simple

python/lib/sift_py/rest.py
```diff
@@ -2,6 +2,8 @@
 
 from typing_extensions import NotRequired
 
+from sift_py.grpc.transport import _clean_uri
```
solidiquis, line 5: non-blocking: we should move the _clean_uri into the top level utils module later

python/examples/data_import/csv/custom/main.py
```diff
@@ -0,0 +1,67 @@
...
+                "format": "TIME_FORMAT_ABSOLUTE_DATETIME",
+                # Can also use `TimeFormatType.ABSOLUTE_DATETIME`
+                "column_number": 1,
+            },
+            "data_columns": data_config,
+        }
+    )
+
+    import_service: DataImportService = csv_upload_service.upload(input_csv, csv_config)
+    import_service.wait_until_complete()
```
solidiquis, line 66: would be cool to set this to the returned data import and print it out

solidiquis (dismissed):

niliayu (approved):
Just two of the tiniest of nits

solidiquis (approved):

PR #105, opened 2024-10-22
feat: Add rule_client_keys, asset_names, and tag_names to rule definitions

niliayu:
This effort adds the parameters needed for `RuleConfig` to become independent of `TelemetryConfig`. This also adds the `rule_client_key` parameter to all rules. 

YAML defined rules can now look like the following:
```
namespace: velocity

rules:
  - name: vehicle_stuck
    description: Triggers if the vehicle velocity is not 0 for 5s after entering accelerating state
    expression: $1 == "Accelerating" && persistence($2 == 0, 5)
    type: review
    rule_client_key: vehicle_stuck_rule_key
    asset_names:
      - foo
      - bar
    tag_names:
      - baz
      - fab
  ```
Or inline, like the following:
```
            RuleConfig(
                name="overheating",
                description="Checks for vehicle overheating",
                expression='$1 == "Accelerating" && $2 > 80',
                channel_references=[
                    {
                        "channel_reference": "$1",
                        "channel_identifier": vehicle_state_channel.fqn(),
                    },
                    {
                        "channel_reference": "$2",
                        "channel_config": voltage_channel,
                    },
                ],
                rule_client_key="overheating_rule_client_key",
                asset_names=["foo", "bar"],
                tag_names=["baz", "fab"],
                action=RuleActionCreateDataReviewAnnotation(),
            ),
```
Although those parameters will not be used until a following PR that implements the second half of this work to connect the backend.

### CODE REVIEW

File: python/examples/ingestion_with_python_config/telemetry_config.py [hunk 1]
```diff
@@ -134,7 +134,7 @@ def nostromos_lv_426() -> TelemetryConfig:
             RuleConfig(
                 name="overvoltage",
                 namespace="voltage",
-                namespace_rules=rule_namespaces,
+                namespace_rules_dict=rule_namespaces,
```
niliayu, line 137: Sorry I wanted to rename this arg to be a little bit clearer and it blew up the diff a little bit

python/lib/sift_py/ingestion/config/telemetry.py
```diff
@@ -220,17 +221,26 @@ def _from_yaml(
                         }
                     )
 
+            if rule.get("asset_names") or rule.get("tag_names"):
```
niliayu, line 224: Ideally we would do this warning in `validate_rules()` but since that's looking at the list of instantiated `RuleConfig`s and `[asset|tag]_names` are not a field, we can't really check for it.

python/examples/ingestion_with_python_config/telemetry_config.py
solidiquis: Can we rename `namespace_rules_dict  to `namespaced_rules`?

python/lib/sift_py/ingestion/config/telemetry.py
```diff
@@ -220,17 +221,26 @@ def _from_yaml(
...
+                logger.warning(
```
solidiquis, line 225: Having now thought about this a bit more could you mind just adding a `TODO` note to emit a warning for now? I haven't had a chance yet to think about the logging/deprecation story for this lib yet. I personally hate it when I use library that emit a bunch of logs I have no control over because it adds to noise. The perfect solution will be to have a logger instance (and it's cihldren) scoped to this library that users can configure in terms of severity level and where the logs go (stdout, stderr, file, or `/dev/null`, etc.). This is something we can do in the future. Could you remove the dependency in the mean time as well?

solidiquis (changes requested):

python/lib/sift_py/ingestion/config/telemetry.py
niliayu: Sure, seems alright to me for now. I feel like it'd be good for us to continue those conversations pretty soon on this just to make sure we're giving clear and ample feedback to users about what's going on sooner than later

solidiquis (approved):

PR #106, opened 2024-10-23
python(feature): asset agnostic rules

solidiquis:
## Summary of Changes

Allow rules to be configured outside of the context of a telemetry config.

### DIFF
--- a/python/examples/asset_agnostic_rules/requirements.txt
+++ b/python/examples/asset_agnostic_rules/requirements.txt
@@ -1,2 +1,2 @@
 python-dotenv
-/Users/benjaminnguyen/Code/sift/python
+sift-stack-py
--- a/python/examples/asset_agnostic_rules/requirements.txt
+++ b/python/examples/asset_agnostic_rules/requirements.txt
@@ -1,0 +1,2 @@
+python-dotenv
+/Users/benjaminnguyen/Code/sift/python
--- a/python/lib/sift_py/ingestion/rule/config.py
+++ b/python/lib/sift_py/ingestion/rule/config.py
@@ -9,6 +9,7 @@ from sift.rules.v1.rules_pb2 import ActionKind
 
 from sift_py._internal.convert.json import AsJson
 from sift_py.ingestion.channel import ChannelConfig
+from sift_py.ingestion.config.yaml.spec import RuleYamlSpec
 
 
 class RuleConfig(AsJson):
@@ -34,6 +35,7 @@ class RuleConfig(AsJson):
     action: Optional[RuleAction]
     channel_references: List[ExpressionChannelReference]
     rule_client_key: Optional[str]
+    asset_names: List[str]
 
     def __init__(
         self,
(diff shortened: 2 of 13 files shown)


python/lib/sift_py/ingestion/rule/config.py
```diff
@@ -166,10 +176,17 @@ def interpolate_namespace_rule(
                     action = RuleActionCreateDataReviewAnnotation(
                         assignee=rule.get("assignee"), tags=tags
                     )
-                return description, expression, rule_client_key, action, asset_names, tag_names
+                return (
+                    description,
+                    cast(str, expression),
```
solidiquis, line 181: @niliayu this is one of the areas that we missed. I'm casting this to a string to get this to work but if you remove the cast `pyright` will tell you what's wrong.

python/lib/sift_py/ingestion/rule/config.py
```diff
@@ -166,10 +176,17 @@ def interpolate_namespace_rule(
...
+                    rule_client_key,
+                    action,
+                    asset_names,
+                    tag_names,
+                )
 
         raise ValueError(
-            f"Could not find rule '{rule}'. Does this rule exist in the namespace? {rule_list}"
+            f"Could not find rule '{candidate_name}'. Does this rule exist in the namespace? {rule_list}"
```
solidiquis, line 189: This was also flagged by pyright

zachAtSiftStack (dismissed):

python/lib/sift_py/rule/service.py
```diff
@@ -0,0 +1,229 @@
...
+
+                rule_name = rule_yaml["name"]
+                rule_fqn = f"{namespace}.{rule_name}"
+                rule_subexpr = interpolation_map.get(rule_fqn, {})
+
+                rule_configs.append(
+                    RuleConfig(
+                        name=rule_yaml["name"],
+                        namespace=namespace,
+                        namespace_rules=namespaced_rules,
```
niliayu, line 88: Just FYI -- when you pass in a namespace & namespace_rules dict, `RuleConfig` will interpolate in the values from the namespaced_rules dict already: https://github.com/sift-stack/sift/blob/155df518efe4d4a23a22964a879e4a0b257c68ae/python/lib/sift_py/ingestion/rule/config.py#L81
So it potentially will overwrite whatever you passed in for description, expression, etc., although it should be all the same since you're also pulling them out of the same dict.

solidiquis: oh that's actually so convenient! I'll include that in a cleanup PR. manually filling in all those fields was not fun.

python/examples/asset_agnostic_rules/requirements.txt
```diff
@@ -0,0 +1,2 @@
+python-dotenv
+/Users/benjaminnguyen/Code/sift/python
```
niliayu, line 2: This looks like it might be accidentally included?

solidiquis: nice catch i was just about ot merge

niliayu (approved):
:shipit:

PR #108, opened 2024-10-23
build: Audit typing with pyright, and add pyright to dev tools and ci

niliayu:
This PR adds `pyright` to the `dev` script and to github actions. It also fixes the typing & import issues not previously surfaced by pyright.

### DIFF
--- a/python/lib/sift_py/data_import/_config.py
+++ b/python/lib/sift_py/data_import/_config.py
@@ -4,7 +4,7 @@ from typing import Dict, List, Optional, Type, Union
 
 from pydantic import BaseModel, ConfigDict, field_validator, model_validator
 from pydantic_core import PydanticCustomError
-from typing_extensions import Self
+from typing_extensions import LiteralString, Self
 
 from sift_py.data_import.time_format import TimeFormatType
 from sift_py.ingestion.channel import ChannelBitFieldElement, ChannelDataType, ChannelEnumType
--- a/python/lib/sift_py/data_import/_config.py
+++ b/python/lib/sift_py/data_import/_config.py
@@ -4,7 +4,7 @@ from typing import Dict, List, Optional, Type, Union
 
 from pydantic import BaseModel, ConfigDict, field_validator, model_validator
 from pydantic_core import PydanticCustomError
-from typing_extensions import LiteralString, Self
+from typing_extensions import Self
 
 from sift_py.data_import.time_format import TimeFormatType
 from sift_py.ingestion.channel import ChannelBitFieldElement, ChannelDataType, ChannelEnumType
--- a/python/lib/sift_py/ingestion/config/telemetry.py
+++ b/python/lib/sift_py/ingestion/config/telemetry.py
@@ -5,7 +5,7 @@ from typing import Any, Dict, List, Optional, cast
 
 from typing_extensions import Self
 
-from sift_py._internal.channel import channel_fqn as _channel_fqn
+from sift_py._internal.channel import channel_fqn
 from sift_py.ingestion.channel import (
     ChannelBitFieldElement,
     ChannelConfig,
@@ -216,7 +216,7 @@ class TelemetryConfig:
                     channel_references.append(
                         {
                             "channel_reference": ref,
-                            "channel_identifier": _channel_fqn(name, component),
+                            "channel_identifier": channel_fqn(name, component),
                         }
                     )
 
--- a/python/lib/sift_py/ingestion/config/telemetry.py
+++ b/python/lib/sift_py/ingestion/config/telemetry.py
@@ -10,7 +10,7 @@ from sift_py.ingestion.channel import (
     ChannelConfig,
     ChannelDataType,
     ChannelEnumType,
-    _channel_fqn,
+    channel_fqn,
 )
 from sift_py.ingestion.config.yaml.load import (
     load_named_expression_modules,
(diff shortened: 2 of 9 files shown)


python/lib/sift_py/data_import/_config.py
```diff
@@ -71,7 +73,10 @@ def convert_format(cls, raw: Union[str, TimeFormatType]) -> str:
             if value is not None:
                 return value.as_human_str()
 
-        raise PydanticCustomError("invalid_config_error", f"Invalid time format: {raw}.")
+        raise PydanticCustomError(
+            cast(LiteralString, "invalid_config_error"),
```
solidiquis, line 77: Just curious why is pyright complaining about this? This seems kind of aggressive.

python/lib/sift_py/ingestion/config/telemetry.py
```diff
@@ -5,12 +5,12 @@
 
 from typing_extensions import Self
 
+from sift_py._internal.channel import channel_fqn as _channel_fqn
```
solidiquis, line 8: This doesn't need to be aliased. Could you update everything to just be `channel_fqn`?

python/lib/sift_py/ingestion/rule/config_test.py
```diff
@@ -158,9 +163,9 @@ def test_rule_namespace():
     assert valid_namespace_rule.name == "valid_rule"
     assert valid_namespace_rule.description == "A rule in a namespace"
     assert valid_namespace_rule.expression == "$1 > 10"
-    assert valid_namespace_rule.action.assignee == "bob@example.com"
-    assert valid_namespace_rule.action.tags == ["foo", "bar"]
-    assert valid_namespace_rule.action.kind() == RuleActionKind.ANNOTATION
+    assert valid_namespace_rule.action.assignee == "bob@example.com"  # type: ignore
+    assert valid_namespace_rule.action.tags == ["foo", "bar"]  # type: ignore
+    assert valid_namespace_rule.action.kind() == RuleActionKind.ANNOTATION  # type: ignore
```
solidiquis, line 168: What errors are we getting without the `type: ignore`?

solidiquis (changes requested):
Just some questions! Kind of funny that we need both `mypy` and `pyright` to catch everything.

python/lib/sift_py/ingestion/rule/config_test.py
niliayu: Pyright is upset about the optional type:
```
  /Users/ailinyu/repos/sift/python/lib/sift_py/ingestion/rule/config_test.py:166:40 - error: Cannot access attribute "assignee" for class "RuleAction"
    Attribute "assignee" is unknown (reportAttributeAccessIssue)
  /Users/ailinyu/repos/sift/python/lib/sift_py/ingestion/rule/config_test.py:166:40 - error: "assignee" is not a known attribute of "None" (reportOptionalMemberAccess)
  /Users/ailinyu/repos/sift/python/lib/sift_py/ingestion/rule/config_test.py:167:40 - error: Cannot access attribute "tags" for class "RuleAction"
    Attribute "tags" is unknown (reportAttributeAccessIssue)
  /Users/ailinyu/repos/sift/python/lib/sift_py/ingestion/rule/config_test.py:167:40 - error: "tags" is not a known attribute of "None" (reportOptionalMemberAccess)
  /Users/ailinyu/repos/sift/python/lib/sift_py/ingestion/rule/config_test.py:168:40 - error: "kind" is not a known attribute of "None" (reportOptionalMemberAccess)
```
I thought in this case since this is a test and these *should* be populated if the test is working, it would be better to ignore and let this just error out if something were to break in the test down the line. Open to other thoughts though, lmk what you think!

niliayu: (Or I guess it's actually that those are fields of child classes, not optional)

python/lib/sift_py/data_import/_config.py
niliayu: Oops sorry I missed this comment earlier -- I ended up reading/skimming a lot of discourse about this, and it seems like it's really to do with the fact that there's a lot of fundamental disagreement on type handling in python, and `LiteralString` seems to a fairly new construct that there isn't yet much agreement on how to handle, so pyright seems to have picked one direction. It seems strange to me as well. I'm a little meh on the cast, so we could silence pyright or just look into options for making it less strict in this case. Any strong opinions on the path here?

marcsiftstack: My vote would be to silence these.

niliayu: Ah I probably should have read this earlier: https://peps.python.org/pep-0675/
It looks like the intent is to make the inputs be `Literal["foo"]`, so I think I just did this in a super subpar way. Otherwise I don't disagree, I think for our use case an ignore is fine. I looks like we should be able to go the `Literal["foo"]` route as well if we want as that seems closer to the intention with this type. Let me see what pyright thinks of that.

niliayu: If it turns out it's pyright just being altogether too strict then I will also vote for just silencing.

niliayu: I added more details of what I've understood to our slack thread. For now I just pushed the version up that suppresses these, but let's chat more in slack if you want to get into the details more!

python/lib/sift_py/ingestion/rule/config_test.py
solidiquis: [Can we just have pyright ignore this rule completely](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#sample-pyprojecttoml-file)?

solidiquis (approved):

PR #110, opened 2024-10-28
python(feat): Add timestamp arg to VideoMetadata

marcsiftstack:
Tested locally with:
```
    remote_file = file_attachment_service.upload_attachment(
        path="/Users/mjulien/Downloads/Short3.mp4",
        entity=run,
        metadata=VideoMetadata(height=2160, width=3840, duration_seconds=28, timestamp=datetime(2024, 10, 19, 2, 22, 22, tzinfo=timezone.utc)),
        description="thrusters getting too hot" ,
    )
```

### DIFF
--- a/python/lib/sift_py/file_attachment/metadata.py
+++ b/python/lib/sift_py/file_attachment/metadata.py
@@ -5,8 +5,10 @@ Though optional, providing this information could help improve quality of render
 
 from __future__ import annotations
 
-from typing import Any, Type
+from datetime import datetime
+from typing import Any, Optional, Type
 
+from google.protobuf.timestamp_pb2 import Timestamp
 from sift.remote_files.v1.remote_files_pb2 import (
     ImageMetadata as ImageMetadataPb,
 )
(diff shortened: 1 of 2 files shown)


python/lib/sift_py/file_attachment/metadata.py
```diff
@@ -49,13 +62,16 @@ def from_pb(cls, message: VideoMetadataPb) -> Self:
             width=message.width,
             height=message.height,
             duration_seconds=message.duration_seconds,
+            timestamp=message.timestamp.ToDateTime(),  # type: ignore
```
niliayu, line 65: Just curious what the typing issue was here?

niliayu (commented):
Generally lgtm, just one little question!

python/lib/sift_py/file_attachment/metadata.py
marcsiftstack:
```
lib/sift_py/file_attachment/metadata.py:65: error: "Timestamp" has no attribute "ToDateTime"  [attr-defined]
```
I think it's failing to find the attribute because this is a protobuf.

niliayu: Hm gotcha. I would really like to understand better why there are so many issues with these tools understanding protobuf imports (like in my own local setup), but I think it's alright for now, especially since this isn't even an optional member so there shouldn't be any issue.

niliayu (approved):

PR #113, opened 2024-10-29
python(refactor): rules module

solidiquis:
## Changes

Refactored rules module to be in its own module outside of telemetry configs while preserving backwards compat.


python/lib/sift_py/rule/_config_test.py
niliayu, line 1: Do we need to get rid of the other `config_test` that lives in the [old location](https://github.com/sift-stack/sift/blob/b74f94cff4614b276051314367d655900926d49e/python/lib/sift_py/ingestion/rule/config_test.py) as well?

niliayu (changes requested):
Just one cleanup I think, otherwise lgtm!

niliayu (dismissed):
lgtm!

niliayu (approved):

PR #130, opened 2024-11-12
python(chore): ping example

solidiquis:
Simple ping example to use in public docs


niliayu (approved):
So cool! 
```
(venv) ailinyu@Ailins-MBP:~/repos/sift/python/examples/ping$ python main.py 
response: "Hello from Sift!"
```

PR #131, opened 2024-11-13
cpp: Add examples for C++

marcsiftstack:
https://dev.app.siftstack.com/share/6xgzIWX

### DIFF
--- a/cpp/README.md
+++ b/cpp/README.md
@@ -1,0 +1,80 @@
+# Sift C++
+
+This tutorial shows how to build a C++ application that uses the Sift gRPC API.
+
+## Installation
+
+##### Build gRPC and Protocol Buffers
+
+Follow the [official gRPC C++ tutorial](https://grpc.io/docs/languages/cpp/quickstart/#install-grpc) to build and locally install gRPC and Protocol Buffers. You should be able to build and run the [helloworld](https://grpc.io/docs/languages/cpp/quickstart/#build-the-example) example before continuing with the next section.
+
+Ensure that `$ which grpc_cpp_plugin` and `$ which protoc` generate a path to executables.
+
+You will also need to ensure that you have the [buf CLI](https://buf.build/docs/installation) installed.
+
+Verify that `$ which buf` generates a path to the executable, before proceeding to the compilation steps.
+
+##### C++ Protobuf Compilation
+
+1. Clone this repository onto your local machine and `cd` into it:
+
+```bash
+$ git clone https://github.com/sift-stack/sift
+$ cd sift
+```
+
+2. Assuming the path to the directory where we'll build the package is `$PACKAGE_DIR`, run the following command in the `sift` directory that you just cloned:
+
+```bash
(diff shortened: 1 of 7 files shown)


cpp/ingestion/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
+cmake_minimum_required(VERSION 3.14)
+project(IngestionExample)
+
+set(CMAKE_CXX_STANDARD 17)
```
jbhunter2, line 4: Not a big deal for a general example, but ULA is on C++11

jbhunter2: @niliayu when you chat through ingestion with them probably good to call out

cpp/README.md
```diff
@@ -0,0 +1,80 @@
+# Sift C++
+
+This tutorial shows how to build a C++ application that uses the Sift gRPC API.
+
+## Installation
+
+##### Build gRPC and Protocol Buffers
+
+Follow the [official gRPC C++ tutorial](https://grpc.io/docs/languages/cpp/quickstart/#install-grpc) to build and locally install gRPC and Protocol Buffers. You should be able to build and run the [helloworld](https://grpc.io/docs/languages/cpp/quickstart/#build-the-example) example before continuing with the next section.
```
niliayu, line 9: Nit, should we include some info here on version compatibility? Like I think we only support protobuf 4 or 5, and I'm assuming there's a corresponding set of grpc versions that are compatible with that as well?

cpp/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+Ensure that `$ which grpc_cpp_plugin` and `$ which protoc` generate a path to executables.
+
+You will also need to ensure that you have the [buf CLI](https://buf.build/docs/installation) installed.
+
+Verify that `$ which buf` generates a path to the executable, before proceeding to the compilation steps.
+
+##### C++ Protobuf Compilation
```
niliayu, line 17: super nitty nit, think it'd be nice to cross-link them to the docs for our protos in one of these sections? maybe as a like: "now that you have the protos compiled, refer to docs to find more info" sort of thing.

cpp/ping/ping.cc
```diff
@@ -0,0 +1,97 @@
...
+    {
+        std::cerr << "Run with SIFT_API_KEY environment variable" << std::endl;
+        exit(1);
+    }
+
+    std::shared_ptr<Channel> channel = CreateChannel(apiUrl, apiKey);
+    std::unique_ptr<sift::ping::v1::PingService::Stub> service = sift::ping::v1::PingService::NewStub(channel);
+
+    ClientContext context;
+    sift::ping::v1::PingRequest request;
```
niliayu, line 89: Love the inclusion of the ping service!! You have great code comments in the rest of this example, my super nitty nit here would be to add one here on why we're including this ping/why a user might want to, especially since this is only the second example that uses the ping service we're adding to the repo.

niliayu (approved):
Just going to stamp this -- we can handle the nits separately after getting this in

PR #136, opened 2024-11-16
python(feat): Add units_row and descriptions_row option to simpele cs…

marcsiftstack:
Some customers ad extra header rows with units. Also allow a row with descriptions.

This PR also adds logic to detect if an integer is a unit64 instead of assuming every integer is an int.

### DIFF
--- a/python/lib/sift_py/data_import/csv.py
+++ b/python/lib/sift_py/data_import/csv.py
@@ -181,12 +181,12 @@ class CsvUploadService:
         data_config = {}
         df = pd.read_csv(path, skiprows=skip_rows)
 
-        units: Optional[List[str]] = None
+        units: List[str] = []
         if units_row is not None:
             df_units = pd.read_csv(path, nrows=units_row)
             units = cast(List[str], df_units.iloc[units_row - 1].astype(str))
 
-        descriptions: Optional[List[str]] = None
+        descriptions: List[str] = []
         if descriptions_row is not None:
             df_descriptions = pd.read_csv(path, nrows=descriptions_row)
             descriptions = cast(List[str], df_descriptions.iloc[descriptions_row - 1].astype(str))
@@ -206,10 +206,10 @@ class CsvUploadService:
 
             data_config[i + 1] = {"name": header, "data_type": dtype}
 
-            if units is not None:
+            if units:
                 data_config[i + 1]["units"] = units[i] if units[i] != "nan" else ""
 
-            if descriptions is not None:
+            if descriptions:
                 data_config[i + 1]["description"] = (
                     descriptions[i] if descriptions[i] != "nan" else ""
                 )
--- a/python/lib/sift_py/data_import/csv.py
+++ b/python/lib/sift_py/data_import/csv.py
@@ -10,6 +10,7 @@ import requests
 from sift_py.data_import.config import CsvConfig
 from sift_py.data_import.status import DataImportService
 from sift_py.data_import.time_format import TimeFormatType
+from sift_py.ingestion.channel import ChannelDataType
 from sift_py.rest import SiftRestConfig, compute_uri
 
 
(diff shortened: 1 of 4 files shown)


python/lib/sift_py/data_import/csv.py
```diff
@@ -149,17 +152,45 @@ def simple_upload(
...
+
+        def is_uint64(n: pd.Series) -> bool:
+            int64_max = 2**63 - 1
+            return bool((n > int64_max).any())
+
         data_config = {}
-        df = pd.read_csv(path)
+        df = pd.read_csv(path, skiprows=skip_rows)
+
+        units: Optional[List[str]] = None
```
solidiquis, line 184: Can this just be initialized as an empty list? 
```
>>> bool([])
False
>>> bool(None)
False
```

python/lib/sift_py/data_import/csv.py
```diff
@@ -149,17 +152,45 @@ def simple_upload(
...
+        if units_row is not None:
+            df_units = pd.read_csv(path, nrows=units_row)
+            units = cast(List[str], df_units.iloc[units_row - 1].astype(str))
+
+        descriptions: Optional[List[str]] = None
```
solidiquis, line 189: same here

solidiquis (changes requested):
small nits

python/lib/sift_py/data_import/csv.py
```diff
@@ -181,12 +181,12 @@ def is_uint64(n: pd.Series) -> bool:
...
+        units: List[str] = []
         if units_row is not None:
             df_units = pd.read_csv(path, nrows=units_row)
             units = cast(List[str], df_units.iloc[units_row - 1].astype(str))
 
-        descriptions: Optional[List[str]] = None
+        descriptions: List[str] = []
         if descriptions_row is not None:
             df_descriptions = pd.read_csv(path, nrows=descriptions_row)
             descriptions = cast(List[str], df_descriptions.iloc[descriptions_row - 1].astype(str))
```
solidiquis, line 192: dont think we need these casts anymore either

python/lib/sift_py/data_import/csv.py
```diff
@@ -181,12 +181,12 @@ def is_uint64(n: pd.Series) -> bool:
         data_config = {}
         df = pd.read_csv(path, skiprows=skip_rows)
 
-        units: Optional[List[str]] = None
+        units: List[str] = []
         if units_row is not None:
             df_units = pd.read_csv(path, nrows=units_row)
             units = cast(List[str], df_units.iloc[units_row - 1].astype(str))
```
solidiquis, line 187: same here

solidiquis (changes requested):

python/lib/sift_py/data_import/csv.py
```diff
@@ -170,10 +203,20 @@ def simple_upload(
                 raise Exception(
                     f"Unable to upload data type in column {i+1} {header}. Inferred type: {inferred_dtype}"
                 )
+            if dtype == "int":
```
niliayu, line 206: What's your thinking for using `infer_dtype()` and doing this vs using `df.dtypes` for getting the datatype for each column?

marcsiftstack: No real reason, this was just the first way I tried.

niliayu: Gotcha! Would you be willing to try using `df.dtypes` to see if that would simplify things here? I wonder if we wouldn't need the `is_uint64` check and could save a few lines there.

python/lib/sift_py/data_import/csv.py
marcsiftstack: Mypy complains because astype is a `Series[bool]`.

python/lib/sift_py/data_import/csv.py
```diff
@@ -197,16 +187,20 @@ def is_uint64(n: pd.Series) -> bool:
             if i + 1 == time_column:
                 continue
 
-            inferred_dtype = pd.api.types.infer_dtype(df[df.columns[i]], skipna=False)
-            dtype = types.get(inferred_dtype)
-            if dtype is None:
+            raw_dtype = str(df[df.columns[i]].dtype)
+            if raw_dtype == "float64":
+                raw_dtype = "double"
```
niliayu, line 192: Why change this to `double`?

marcsiftstack: https://github.com/sift-stack/sift/blob/python/ch10_upload_service/python/lib/sift_py/ingestion/channel.py#L187

niliayu (approved):
lgtm!

solidiquis (approved):

PR #137, opened 2024-11-18
Python/ch10 upload service

marcsiftstack:


python/lib/sift_py/data_import/ch10.py
```diff
@@ -0,0 +1,111 @@
...
+        next(ch10_file)
+
+        config_info: Dict[str, Any] = {
+            "asset_name": asset_name,
+            "first_data_row": 1,
+            "time_column": {
+                "format": time_format,
+                "column_number": 1,
+            },
+            "data_columns": ch10_file.csv_config_data_columns,
```
niliayu, line 61: Is it possible for users to get to this point and have not provided `csv_config_data_columns`? (And can we raise an exception if it is possible? Otherwise no worries if it's not.)

python/lib/sift_py/data_import/ch10.py
```diff
@@ -0,0 +1,111 @@
...
+        try:
+            upload_info = response.json()
+        except (json.decoder.JSONDecodeError, KeyError):
+            raise Exception(f"Invalid response: {response.text}")
+
+        try:
+            upload_url: str = upload_info["uploadUrl"]
+            data_import_id: str = upload_info["dataImportId"]
+        except KeyError as e:
+            raise Exception(f"Response missing required keys: {e}")
```
niliayu, line 94: nit: Indicating to the users if we think this error (and potentially some of the similar errors in this function) is a them-problem that they need to fix or an us-problem that they should reach out to us about could be nice.

python/examples/data_import/ch10/main.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+if __name__ == "__main__":
+    """
+    Example of uploading a TDMS file into Sift.
+    """
+    load_dotenv()
+
+    sift_uri = os.getenv("SIFT_API_URI")
+    assert sift_uri, "expected 'SIFT_API_URI' environment variable to be set"
```
niliayu, line 27: Do we want to provide an `.env.example` with this example as well?

niliayu (commented):
Just a few nits!

python/lib/sift_py/data_import/ch10.py
```diff
@@ -0,0 +1,111 @@
...
+        config_info: Dict[str, Any] = {
+            "asset_name": asset_name,
+            "first_data_row": 1,
+            "time_column": {
+                "format": time_format,
+                "column_number": 1,
+            },
+            "data_columns": ch10_file.csv_config_data_columns,
+        }
+        if run_name is not None:
```
solidiquis, line 63: Can we just do `if run_name`

python/lib/sift_py/data_import/ch10.py
```diff
@@ -0,0 +1,111 @@
...
+            config_info["run_name"] = run_name
+
+        if run_id is not None:
```
solidiquis, line 66: `if run_id`

python/lib/sift_py/data_import/ch10.py
```diff
@@ -0,0 +1,111 @@
...
+        )
+
+        if response.status_code != 200:
+            raise Exception(
+                f"Config file upload request failed with status code {response.status_code}. {response.text}"
+            )
+
+        try:
+            upload_info = response.json()
+        except (json.decoder.JSONDecodeError, KeyError):
```
solidiquis, line 87: might be a good idea to put this into a variable and include in error output

python/lib/sift_py/data_import/ch10.py
```diff
@@ -0,0 +1,111 @@
...
+from sift_py.data_import.csv import CsvUploadService
+from sift_py.data_import.status import DataImportService
+from sift_py.data_import.time_format import TimeFormatType
+
+
+class BaseCh10File:
+    """
+    Base class for uploading IRIG Chapter 10/Chapter 11 files.
+
+    Implement a concrete version of this class that parses a ch10 stream and returns
```
solidiquis, line 16: I think it would be help to add type hints to the `__iter__` and `__next__` functions to make it very clear how this base class should be implemented.

solidiquis: Also i think an inline example showing how to implement this class would help remarkably. Checkout `lib/sift_py/ingestion/service.py` -> `buffered_ingestion`.

solidiquis (changes requested):

python/lib/sift_py/data_import/ch10.py
marcsiftstack: I'll update this one. The one above could still be a user error if they use the wrong api url.

solidiquis (changes requested):

solidiquis (approved):

PR #139, opened 2024-11-21
python(feat): Allow uploading compressed ch10

marcsiftstack:
Also make setting the csv data columns simpler.

### DIFF
--- a/python/lib/sift_py/data_import/ch10.py
+++ b/python/lib/sift_py/data_import/ch10.py
@@ -16,9 +16,7 @@ class BaseCh10File:
     Implement a concrete version of this class that parses a ch10 stream and returns
     a csv row of data on each iteration.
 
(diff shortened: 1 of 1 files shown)


python/lib/sift_py/data_import/ch10.py
```diff
@@ -79,17 +86,15 @@ def upload_ch10(
...
+        ch10_file.initialize_csv_data_columns()
 
         assert getattr(
             ch10_file, "csv_config_data_columns"
         ), "`csv_config_data_columns` was not set correctly on the first iteration"
 
         config_info: Dict[str, Any] = {
             "asset_name": asset_name,
-            "first_data_row": 1,
+            "first_data_row": 2,
```
marcsiftstack, line 97: Since we are not sending the entire file instead of skipping the header, this should be 1.

niliayu (approved):
lgtm!

PR #145, opened 2024-11-26
python(feat): Report templates service

niliayu:
This PR adds a `ReportTemplateService` that allows users to create, update, fetch, and archive report templates. There are also examples of how to define report templates in python, in yaml, and how to fetch, update, and archive those created templates. This also notably adds the `ReportTemplateConfig` as a pydantic model.


python/lib/sift_py/report_templates/service.py
```diff
@@ -0,0 +1,139 @@
...
+    def _update_report_template(
+        self, config: ReportTemplateConfig, report_template: ReportTemplate
+    ):
+        tags = []
+        if config.tags:
+            tags = [ReportTemplateTag(tag_name=tag) for tag in config.tags]
+
+        archived_date = None
+        if config.archived_date:
+            archived_date = to_timestamp_pb(config.archived_date)
```
niliayu, line 117: @solidiquis I noticed [this code comment ](https://github.com/sift-stack/sift/blob/main/python/lib/sift_py/_internal/time.py#L34-L35)on this function, do you have a better recommended approach?

python/lib/sift_py/report_templates/service.py
```diff
@@ -0,0 +1,139 @@
...
+            organization_id=report_template.organization_id,
+            client_key=report_template.client_key,
+            name=config.name,
+            description=config.description,
+            tags=tags,
+            rules=rules,
+            archived_date=archived_date,
+        )
+
+        field_mask = FieldMask(paths=["name", "description", "tags", "rules", "archived_date"])
```
niliayu, line 134: I was originally thinking we could be smarter about this and check each field to see which ones have actually changed, and just include those in the fieldmask, but I'm not sure if that buys us anything

python/lib/sift_py/report_templates/service.py
```diff
@@ -0,0 +1,139 @@
...
+            raise Exception(f"Report template with client key {client_key} or id {id} not found.")
+
+        return ReportTemplateConfig(
+            name=report_template.name,
+            template_client_key=report_template.client_key,
+            organization_id=report_template.organization_id,
+            tags=[tag.tag_name for tag in report_template.tags],
+            description=report_template.description,
+            rule_client_keys=[rule.client_key for rule in report_template.rules],
+            archived_date=to_timestamp_nanos(report_template.archived_date).to_pydatetime(),
```
solidiquis, line 63: `to_pydatetime` looks like its on its way to be deprecated based on docs. We can easily convert a pb timestamp to a python datetime using the [ToDatetime](https://googleapis.dev/python/protobuf/latest/google/protobuf/timestamp_pb2.html#google.protobuf.timestamp_pb2.Timestamp.ToDatetime) method.

python/lib/sift_py/report_templates/service.py
solidiquis: This should be fine

python/lib/sift_py/report_templates/service.py
solidiquis: Since this is "private" that should be fine? It's only accessible via the `create_or_update_*` method so I can't see any potential footguns but I'd just double check and consider this a bit more.

python/lib/sift_py/report_templates/service.py
```diff
@@ -0,0 +1,139 @@
...
+)
+from sift.report_templates.v1.report_templates_pb2_grpc import ReportTemplateServiceStub
+
+from sift_py._internal.time import to_timestamp_nanos, to_timestamp_pb
+from sift_py.grpc.transport import SiftChannel
+from sift_py.ingestion.config.yaml.load import load_report_templates
+from sift_py.report_templates.config import ReportTemplateConfig
+
+
+class ReportTemplateService:
```
solidiquis, line 25: Could we go pretty aggressive with the doc comments for this class? OPtional for the private method unless you think it'll help other devs

python/lib/sift_py/ingestion/config/yaml/spec.py
```diff
@@ -90,6 +91,29 @@ class FlowYamlSpec(TypedDict):
...
+    `tags`: Tags to associate with the report template.
+    `description`: Description of the report template.
+    `rule_client_keys`: List of rule client keys associated with the report template.
+    `archived_date`: Date when the report template was archived. Setting this field
+      will archive the report template, and unsetting it will unarchive the report template.
+    """
+
+    name: str
+    template_client_key: str
+    organization_id: str
```
solidiquis, line 110: Organization ID should be marked as `NotRequired`. We should reflect that in the doc comments as well.

python/lib/sift_py/ingestion/config/yaml/load.py
```diff
@@ -31,6 +32,25 @@ def read_and_validate(path: Path) -> TelemetryConfigYamlSpec:
     return _validate_yaml(raw_config)
 
 
+def load_report_templates(paths: List[Path]) -> List[ReportTemplateConfig]:
```
solidiquis, line 35: I think we should start breaking this yaml module apart. It is starting to feel awkward to have this be the dumping ground for parsing different yaml specs.

I think what make sense to me is having a top-level module along-side `report_template` maybe called `yaml_specs` and having the specs and parsing logic be separated by domain i.e. `yaml_specs/report_template`, `yaml_specs/rule`, `yaml_spec/telemetry_config` or something like that. We don't have to do this for everything right now, but can we move report templates logic out of this module? We can do the same for rule for your rules PR. We'll move telemetry config stuff out later.

niliayu: Sure, I think that sounds good! I'll go ahead and create this for report templates and then we can update this for the other domains the next time we touch them.

niliayu: Updated, let me know how that looks!

python/lib/sift_py/report_templates/service.py
niliayu: Updated! For the private ones I just added a docstring to the `_update` method to try to clarify how I'm doing things there.

solidiquis (approved):
LGTM!

python/examples/report_templates/report_template_with_python_config/.env-example
```diff
@@ -0,0 +1,8 @@
+# Retrieve the BASE_URI from the Sift team
+# Be sure to exclude "https://" from the BASE_URI
+#
+# BASE URIs and further details can be found here:
+#   https://docs.siftstack.com/ingestion/overview
```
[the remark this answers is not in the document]
niliayu: Ah really good shout! The next ticket I'm working on will update all of our examples, so I'll include updating these links to the new ones under the scope of that ticket, and just fix them all in one go that way. That ticket will be done before our next release as well, fwiw.

PR #146, opened 2024-12-04
python(refactor): Asset agnostic rules cleanup

niliayu:
This PR:
- Adds the capability to specify `channel_references` when loading rule definitions from YAML
- Moves rule & channel yaml parsing load functions and specs into yaml domain
    - Resolves a circular import associated with this refactor
- Adds code comments and more specific errors to RuleService


python/examples/asset_agnostic_rules/main.py
```diff
@@ -56,21 +57,33 @@
...
-                SubExpression(
-                    "velocity.vehicle_not_stopped",
-                    {
-                        "$1": "vehicle_state",
-                        "$2": "10",
-                    },
-                ),
+                SubExpression("velocity.vehicle_not_stopped", {"$2": 10}),
             ],
+            channel_references_map={
```
marcsiftstack, line 62: Is it possible to keep references for a rule colocated like the original? It makes it a lot easier to look back and forrth between the .yml and here to see the substitutions and understand the rule.

Maybe something like this:
```
                SubExpression(
                    "velocity.vehicle_not_stopped",
                    {
                        "$1": ExpressionChannelReference("vehicle_state"),
                        "$2": "10",
                    },
                ),
```

niliayu: Sure I think that makes sense!

niliayu: Addressed in [080e763](https://github.com/sift-stack/sift/pull/146/commits/080e763e87e7c6891bbd028672813277a02d59b5)!

python/examples/asset_agnostic_rules/main.py
```diff
@@ -56,20 +56,15 @@
...
-                SubExpression(
-                    "velocity.vehicle_not_stopped",
-                    {
-                        "$1": "vehicle_state",
-                        "$2": "10",
-                    },
+                SubExpression("velocity.vehicle_not_stopped", {"$2": 10}),
+            ],
+            channel_references=[
+                RuleChannelReference("overvoltage", {"$2": "vehicle_state"}),
```
marcsiftstack, line 62: Why does `SubExpression` include the namespace, but `RuleChannelReference` doesnt?

niliayu: That's a good shout, I forgot to mentioned this -- this issue going to be resolved in the next PR when I deprecate namespaces. So it'll be consistent when we release.

python/lib/sift_py/rule/service.py
```diff
@@ -48,35 +54,87 @@ def __init__(self, channel: SiftChannel):
...
+                interpolation_map[sub_expression.fully_qualified_rule_name] = (
+                    sub_expression.expressions
+                )
 
         rule_configs = []
         for namespace, rule_yamls in namespaced_rules.items():
             for rule_yaml in rule_yamls:
-                channel_references: List[
+                yaml_channel_references = rule_yaml.get("channel_references", [])
+                arg_channel_references = None
```
marcsiftstack, line 84: Add type hint.

python/lib/sift_py/rule/service.py
```diff
@@ -227,3 +337,17 @@ class SubExpression:
     def __init__(self, fully_qualified_rule_name: str, expressions: Dict[str, Any]):
         self.fully_qualified_rule_name = fully_qualified_rule_name
         self.expressions = expressions
+
+
+class RuleChannelReference:
```
marcsiftstack, line 342: Nit: This can just be a `dataclass` or `namedtuple`.

python/lib/sift_py/rule/service.py
```diff
@@ -48,35 +54,87 @@ def __init__(self, channel: SiftChannel):
...
 
         rule_configs = []
         for namespace, rule_yamls in namespaced_rules.items():
             for rule_yaml in rule_yamls:
-                channel_references: List[
+                yaml_channel_references = rule_yaml.get("channel_references", [])
+                arg_channel_references = None
+
+                if channel_references:
+                    for channel_map in channel_references:
```
marcsiftstack, line 87: Would `rule_channel_reference` make more sense than `channel_map` here?

niliayu: Yeah I like that!

python/lib/sift_py/rule/service.py
```diff
@@ -163,6 +234,45 @@ def _update_req_from_rule_config(
             ident = channel_reference_from_fqn(channel_reference["channel_identifier"])
             channel_references[ref] = ident
 
+        def search_channels(filter="", page_size=1_000, page_token="") -> Tuple[List[Channel], str]:
```
marcsiftstack, line 237: Any reason not to move this out into a common module?

niliayu: I was also thinking about this... there sort of already is one: https://github.com/sift-stack/sift/blob/main/python/lib/sift_py/ingestion/_internal/channel.py#L11-L40
In this case it doesn't work as-is since the filter string isn't generalizable. I could just refactor this to use though... looking at it now it looks like it is both private and not actually used anywhere else currently. I'll give that a shot!

python/lib/sift_py/yaml/channel_test.py
```diff
@@ -0,0 +1,169 @@
+import pytest
+
+from sift_py.ingestion.config.yaml.error import YamlConfigError
+from sift_py.yaml import channel
+
+
+def test__validate_channel_anchor():
```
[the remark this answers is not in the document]
niliayu: I'm assuming it's because it's a private function, so it kept the extra underscore at the front of the channel name. I've seen this in a few places in the repo. I don't have strong feelings about it personally, in a way it does indicate to me as I'm reading that the function being tested is private without having to go check separately.

python/lib/sift_py/rule/service.py
niliayu: Went with dataclass!

python/lib/sift_py/_internal/channel.py
```diff
@@ -1,5 +1,42 @@
...
+from sift.channels.v2.channels_pb2 import Channel as ChannelPb
+from sift.channels.v2.channels_pb2 import ListChannelsRequest, ListChannelsResponse
+from sift.channels.v2.channels_pb2_grpc import ChannelServiceStub
 
 
 def channel_fqn(name: str, component: Optional[str]) -> str:
     return name if component is None or len(component) == 0 else f"{component}.{name}"
+
+
+def get_asset_channels(
```
niliayu, line 12: Decided to move this to a more general location since this utility isn't specific to ingestion anymore.

marcsiftstack, line 12: I think this should just be called `get_channels` or something like that since it's not specific to `assets`.

python/lib/sift_py/_internal/user.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def get_active_users(
+    user_service: UserServiceStub,
+    filter: str,
+    page_size: int = 1_000,
+    page_token: str = "",
+) -> List[User]:
+    """
+    Queries all channels for the given `asset_id`.
```
marcsiftstack, line 15: Update docstring.

marcsiftstack (approved):

PR #148, opened 2024-12-06
python(chore): Standardize ch10 interface

marcsiftstack:


python/lib/sift_py/data_import/ch10.py
```diff
@@ -71,7 +71,7 @@ def __next__(self) -> str:
 class Ch10UploadService(CsvUploadService):
     """Service to upload ch10 files."""
 
-    def upload_ch10(
+    def upload(  # type: ignore
```
niliayu, line 74: What was the error we're suppressing here?

marcsiftstack: It was complaining about overriding the parent function with a different function signature.

niliayu (approved):

PR #150, opened 2024-12-09
python(feat): Add offline installation archives for various platforms

kevin-sift:
### Context
Some users wish to install `sift-stack-py` without access to pypi. 

### Description
This PR adds a new workflow `python_build.yaml` that builds offline installation archives (wheels for each dependency) that target all supported python versions and various platforms (linux, mac, windows) and architectures (e.g. x86_64/arm64).

Users will be able to install `sift-py-stack` by running `pip install sift-py-stack --no-index --find-links {path/to/offline-installation-archive}`. 

### Verification

Ran `python_release.yaml` workflow which calls the new `python_build.yaml` on a dev tag and successfully created a release

https://github.com/sift-stack/sift/actions/runs/12273019715

image

Manually downloaded and inspected archives as well as installing locally on my mac.

### DIFF
--- a/.github/workflows/python_build.yaml
+++ b/.github/workflows/python_build.yaml
@@ -49,6 +49,8 @@ jobs:
             arch: amd64
             runner: windows-latest
         python-version: ${{fromJson(needs.get-python-versions.outputs.python-versions)}}
+    outputs:
+      platforms: ${{ toJson(matrix.config) }}
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


.github/workflows/python_build.yaml
```diff
@@ -0,0 +1,214 @@
+name: Offline Installation Archive
+
+on:
+  workflow_dispatch:
+  pull_request:
```
kevin-sift, line 5: Remove this now that it works. We may want to trigger this on release (completion of python_release workflow)

solidiquis: +1 for triggering this on release.

solidiquis: If we trigger on release should we make this a `workflow_call` rather than a workflow dispatch? That way we use https://github.com/sift-stack/sift/blob/main/.github/workflows/python_ci.yaml upstream and do all the last minute sanity checks before committing to building

python/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "sift_stack_py"
-version = "0.3.0-rc.6"
+version = "0.3.0-rc.7"
```
solidiquis, line 3: Is there a reason we're bumping this?

solidiquis (changes requested):

.github/workflows/python_build.yaml
```diff
@@ -0,0 +1,211 @@
+name: Offline Installation Archive
+
+on:
+  workflow_dispatch:
```
kevin-sift, line 4: on workflow dispatch we should run ci first, if called no need

.github/workflows/python_build.yaml
```diff
@@ -0,0 +1,211 @@
...
+          python scripts/build_utils.py \
+            --dist-dir dist \
+            --package-name sift-stack-py \
+            ${{ matrix.platform.os == 'windows' && '--is-windows' || '' }}
+
+      - name: Create distribution archive
+        working-directory: python
+        shell: bash
+        run: |
+          if [ "${{ matrix.platform.os }}" = "windows" ]; then
```
kevin-sift, line 158: lets make tar balls instead of zips, we can publish both in the final merged artifacts

python/pyproject.toml
kevin-sift: I reverted

solidiquis (approved):
LGTM

PR #153, opened 2024-12-13
rust(feat): http2 keepalive

solidiquis:
## Changes

- In the same vein as this PR https://github.com/sift-stack/sift/pull/152, we want to enable keep-alive by default since this is a pretty esoteric setting that most users aren't aware about and thus face timeouts as a result of. A common area of timeout is downloading large amounts of data across a long time range from the data API. This PR enables keep-alive by default with documentation instructing folks on how to provide their own custom settings if they need it.
- This PR also adds a custom client-side user-agent for tracking library usage + helps with debugging.
- Re-exports protobuf-generated code in the top-level crate to prevent long import paths for users.

We will also take the Rust client library out of release candidate once this is merged.


rust/src/grpc/mod.rs
```diff
@@ -15,13 +16,27 @@ use interceptor::AuthInterceptor;
...
 pub fn use_sift_channel(channel_config: SiftChannelConfig) -> Result<SiftChannel> {
     let SiftChannelConfig { uri, apikey } = channel_config;
 
+    let crate_name = env!("CARGO_PKG_NAME");
+    let crate_version = env!("CARGO_PKG_VERSION");
+    let user_agent = format!("{crate_name}/{crate_version}");
+
     let channel = Endpoint::from_shared(uri)
+        .map(|conn| {
+            conn.keep_alive_while_idle(true)
```
niliayu, line 34: Do we eventually want to allow users to pass in their own keepalive config similar to how it's set up on the python side?

solidiquis: I will err on the side of no, for the time being because I don't think it's something the vast majority of users will want to think about on their own - they just want something that works well out of the box. Ultimately if they want to get really granular with their keepalive settings they are the ability to create their own custom grpc channel and use that instead

niliayu: Sounds good!

niliayu (approved):
lgtm!

PR #154, opened 2024-12-13
python(refactor): Update examples to use RuleService and deprecate namespaces

niliayu:
This PR:
- Deprecates `namespaces`, `SubExpression`, and `RuleChannelReferences`
- Updates ingestion examples to use `RuleService` to create rules (the new preferred workflow) instead of `TelemetryConfig`
- Removes the `asset_agnostic_rules` example since the asset agnostic approach is now the preferred/default (and is equivalent to the ingestion examples now)
- Updates `.env-example` doc links to point to the new documentation


python/lib/sift_py/rule/config.py
```diff
@@ -245,6 +187,26 @@ def from_str(cls, val: str) -> Optional["RuleActionKind"]:
         return None
 
 
+class RuleActionAnnotationKind(Enum):
```
niliayu, line 190: FYI on this one -- I originally had moved this due to it creating a circular dependency, then after re-arranging things on this PR it was circular again... so I moved it back, and things seem to work fine now.

python/examples/ingestion_with_python_config/telemetry_config.py
```diff
@@ -71,131 +66,166 @@ def nostromos_lv_426() -> TelemetryConfig:
...
+            name="overvoltage",
+            description="Checks for overvoltage while accelerating",
+            expression='$1 == "Accelerating" && $2 > 80',
+            rule_client_key="overvoltage-rule",
+            asset_names=["NostromoLV426"],
+            channel_references=[
+                # INFO: Can use either "channel_identifier" or "channel_config"
+                {
+                    "channel_reference": "$1",
+                    "channel_identifier": vehicle_state_channel.fqn(),
```
solidiquis, line 142: I'm thinking about how to distill the example to be something much more simple. When I originally wrote the example I tried to show a little bit of everything but that's probably not necessary and may lock us into feature sets that we eventually want to move away from. `components` for example is something we're planning to deprecate.

For this example in particular, rather than `channel_identifier` or `channel_config` what if we do a union type using `channel` as the key which takes in either a string or config?

Also we should probably create a method called `identifier` on `ChannelConfig` and make it an alias for `fqn`.

python/examples/ingestion_with_python_config/telemetry_config.py
```diff
@@ -17,20 +18,14 @@
 RULE_NAMESPACES_DIR = Path().joinpath("rule_modules")
 
 
-def nostromos_lv_426() -> TelemetryConfig:
+def nostromos_lv_426() -> Tuple[TelemetryConfig, List[RuleConfig]]:
     named_expressions = load_named_expression_modules(
```
solidiquis, line 22: Can we get rid of named expressions from the examples? With asset agnostic rules we can start moving away from this

python/examples/ingestion_with_yaml_config/main.py
```diff
@@ -36,10 +43,26 @@
         ingestion_service = IngestionService(
             channel,
             telemetry_config,
-            overwrite_rules=True,  # Overwrite any rules created in the Sift UI that isn't in the config
             end_stream_on_error=True,  # End stream if errors occur API-side.
         )
 
+        # Create/update rules
+        named_expressions = load_named_expression_modules(
```
solidiquis, line 50: As I said above, let's remove named expression modules from the examples in favor of pushing asset agnostic rules flow as much as we can.

python/examples/ingestion_with_python_config/main.py
```diff
@@ -26,7 +27,7 @@
         raise Exception("Missing 'BASE_URI' environment variable.")
 
     # Load your telemetry config
-    telemetry_config = nostromos_lv_426()
+    telemetry_config, rules = nostromos_lv_426()
```
solidiquis, line 30: This part feels a tad awkward to me, the fact that we return a config a rules separately and leave it up to the caller to then have to create rules manually.

Similar to how instantiating a telemetry config creates your asset, channels, and ingestion config for you, wondering if we could follow a similar model with the new `RuleService` where all you have to do is pass in a list of rule configs and all the create/update happens automatically for you. Let me know if you want to talk about this more. The UX seems a bit disjoint with the current direction that we're going.

solidiquis (changes requested):

python/lib/sift_py/rule/service.py
```diff
@@ -159,6 +159,60 @@ def load_rules_from_yaml(
...
 
+    def create_or_update_rules(self, rule_configs: List[RuleConfig]):
+        """
+        Create or update a list of rules via a list of RuleConfigs.
+        See `sift_py.rule.config.RuleConfig` for more information on configuation parameters for rules.
+        """
+        for config in rule_configs:
+            self.create_or_update_rule(config)
+
+    def attach_asset(self, rule: Union[str, RuleConfig], asset_names: List[str]):
```
solidiquis, line 170: Might be nice to return a new RuleConfig which contains the newly attached asset in the `asset_names`?

solidiquis (changes requested):
This is looking good! Could we also do a tags variant + unit testing for this service?

python/lib/sift_py/rule/service.py
niliayu: Addressed on [a5f8b29](https://github.com/sift-stack/sift/pull/154/commits/a5f8b2941221df3c0d57b372ad1bb0ca481bed91)! 
One question for you though that came up as I implemented this: https://github.com/sift-stack/sift/pull/154/commits/fe2366c4ca4d2bac1459ab3b7fe95b84e5a1047e#r1890750956

python/examples/ingestion_with_python_config/main.py
niliayu: Discussed offline -- we have a plan to align this UX that is in work now.

python/lib/sift_py/ingestion/_internal/ingest.py
```diff
@@ -58,13 +59,17 @@ def __init__(
     ):
         ingestion_config = self.__class__._get_or_create_ingestion_config(channel, config)
 
-        if not overwrite_rules:
-            self.__class__._validate_rules_synchronized(
-                channel, ingestion_config.asset_id, config.rules
+        if overwrite_rules:
+            warn(
```
niliayu, line 63: I wasn't sure if I was in the clear to just remove this arg, so I did this to be cautious. Happy to fully pull it out if we think we're in the clear for that though.

python/lib/sift_py/ingestion/_internal/ingest.py
```diff
@@ -58,13 +59,17 @@ def __init__(
...
+                "The 'overwrite_rules' argument is deprecated and will be removed in a future release."
             )
 
+        self.rule_service = RuleService(channel)
         if len(config.rules) > 0:
-            update_rules(channel, ingestion_config.asset_id, config.rules)
+            for rule in config.rules:
+                if config.asset_name not in rule.asset_names:
```
niliayu, line 70: I felt like part of the simplicity of just including rules in the telemetry config was so that you didn't have to necessarily separately configure the asset_names list on the rules, when the telemetry config they were a part of is already tied to a specific asset. So I thought it'd make sense to just ensure that the asset is configured on the rule by default. Let me know if you think that makes sense though.

python/lib/sift_py/rule/service.py
```diff
@@ -392,6 +392,57 @@ def _update_req_from_rule_config(
...
+                channel_references.append(
+                    {
+                        "channel_reference": ref,
+                        "channel_identifier": id.name,
+                    }
+                )
+            for action_config in condition.actions:
+                if action_config.action_type == ANNOTATION:
+                    assignee = action_config.configuration.annotation.assigned_to_user_id
+                    if assignee:  # TODO: Better way to disambiguate
```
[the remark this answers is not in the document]
niliayu: I think I may have answered my own question, here's what I tried: [fb0fca2](https://github.com/sift-stack/sift/pull/154/commits/fb0fca22b72a1bd16f7b3051a49e7471a2dd3827)
Let me know if that approach makes sense!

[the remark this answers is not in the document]
solidiquis: make sense to me!

python/lib/sift_py/ingestion/_internal/ingest.py
solidiquis: Could we specify the exact version we plan to remove it? Next the upcoming but the one after that and create it a ticket for it?

python/lib/sift_py/ingestion/_internal/ingest.py
solidiquis: makes sense to me!

solidiquis (changes requested):

python/examples/ingestion_with_python_config/telemetry_config.py
niliayu: Let me know if what I've pushed aligns with what you were thinking here or if you have an even simpler vision in mind!

niliayu:
In [6acc0ac](https://github.com/sift-stack/sift/pull/154/commits/6acc0ac6c3a36b92a06536810761305bb8530a0d) I decided to pull out some of the example rules I had added to demonstrate namespaces originally. Now that namespaces don't exist I don't think they add anything useful to the example any longer.

niliayu:
Following up from an offline discussion:

> This is looking good! Could we also do a tags variant + unit testing for this service?


This requires the TagsService that we have not yet implemented, but is ticketed as upcoming work, so the tags variant will be available in a future PR & release.

kevin-sift (approved):

PR #158, opened 2025-01-07
python(chore): v0.3.0 release prep

niliayu:
Updating CHANGELOG and bumping version for release.

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -1,28 +1,24 @@
 [project]
 name = "sift_stack_py"
-version = "0.3.0-rc.7"
+version = "0.3.0"
 description = "Python client library for the Sift API"
 requires-python = ">=3.8"
-readme = {file = "README.md", content-type = "text/markdown"}
-license = {file = "LICENSE"}
-classifiers  = [
-    "License :: OSI Approved :: MIT License",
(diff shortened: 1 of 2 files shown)


python/pyproject.toml
```diff
@@ -1,28 +1,24 @@
 [project]
 name = "sift_stack_py"
-version = "0.3.0-rc.7"
+version = "0.3.0"
```
niliayu, line 3: My formatters went pretty hard on this file, but all I really did was update this version on line 3 and move the npTDMS requirement to line 31.

zachAtSiftStack: confirmed only real change was the version update and npTDMS to optional

zachAtSiftStack:
Compared against https://github.com/sift-stack/sift/commit/0d5d247ccf7531933096f121830de2319e2e2bd0#diff-0dd30ea12dcdbf90c6eac76978a559ab02dcc5367f9d8af32630dfcc369534deR9 for a previous update of the version from release-candidate to a new release. 

The only thing really deferring is the squashing of release candidate change logs into one top level release change log. I don't believe that really matters so approving as is.

zachAtSiftStack (approved):
Approving, update matches previous releases.

PR #159, opened 2025-01-08
python(chore): Make npTDMS an optional dep

niliayu:
The `npTDMS` arg should not be required for the standard lib, but included as an optional dep for users who do use the lib to manage TDMS data. Users can include `sift-stack-py[tdms]` in their project dependencies if TDMS is required.

### DIFF
--- a/python/examples/data_import/tdms/requirements.txt
+++ b/python/examples/data_import/tdms/requirements.txt
@@ -1,2 +1,3 @@
 python-dotenv
 sift-stack-py
+sift-stack-py[tdms]
--- a/python/examples/data_import/tdms/requirements.txt
+++ b/python/examples/data_import/tdms/requirements.txt
@@ -1,3 +1,2 @@
 python-dotenv
-sift-stack-py
 sift-stack-py[tdms]
--- a/python/lib/sift_py/data_import/tdms.py
+++ b/python/lib/sift_py/data_import/tdms.py
@@ -13,7 +13,7 @@ try:
     )
 except ImportError as e:
     raise RuntimeError(
-        "The nptdms package is required to use the TDMS upload service. "
+        "The npTDMS package is required to use the TDMS upload service. "
         "Please include this dependency in your project by specifying `sift-stack-py[tdms]`."
     ) from e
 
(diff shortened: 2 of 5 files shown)


zachAtSiftStack:
Learning how pip optional dependencies work here on the fly. This looks correct to my new understanding. 

If I want to do something without nptdms I do
`pip install sift-stack-py`

If I want to do something with nptdms I do
`pip install sift-stack-py[tdms]`

zachAtSiftStack:
The only thing I suspect may happen in the future is a user try and use the tdms functionality and not know to do the optional dependency. I believe they will just get an ImportError.

We could wrap the nptdms imports in sift_py with our own custom exception handler to notify them? Would be a nice to have.

I think an update to https://docs.siftstack.com/docs/ingestion/data-import/tdms to call this out will make sense but that isn't blocking this PR.

kevin-sift:
Strongly agree with @zachAtSiftStack that we should wrap the `npTDMS` import with a `try/except` block.

niliayu:
Updated! Let me know what you think @kevin-sift @zachAtSiftStack [1e60cf2](https://github.com/sift-stack/sift/pull/159/commits/1e60cf22925eec37f0272b9d9e15f170c1e41091)

python/examples/data_import/tdms/requirements.txt
```diff
@@ -1,2 +1,3 @@
 python-dotenv
 sift-stack-py
+sift-stack-py[tdms]
```
kevin-sift, line 3:
```suggestion
sift-stack-py[tdms]
```

python/lib/sift_py/data_import/tdms.py
```diff
@@ -2,14 +2,20 @@
...
+        ChannelObject,
+        RootObject,
+        TdmsChannel,
+        TdmsFile,
+        TdmsWriter,
+        types,
+    )
+except ImportError as e:
+    raise RuntimeError(
+        "The nptdms package is required to use the TDMS upload service. "
```
kevin-sift, line 16:
```suggestion
        "The npTDMS package is required to use the TDMS upload service. "
```

zachAtSiftStack (approved):
All suggestions have been implemented from both reviewers.

PR #168, opened 2025-01-29
chore: update protos

derouinw:
Updated protos to the latest version of the sift APIs. Generated code for all languages.


solidiquis (approved):
noice

PR #174, opened 2025-02-13
python(feat): add support for calculated channels

alexluck-sift:


python/lib/sift_py/calculated_channels/service.py
```diff
@@ -0,0 +1,338 @@
...
+            ),
+        )
+        calculated_channel_configuration = CalculatedChannelConfiguration(
+            asset_configuration=asset_configuration, query_configuration=query_configuration
+        )
+        req = CreateCalculatedChannelRequest(
+            name=config.name,
+            description=config.description,
+            units=config.units,
+            client_key=config.client_key,
```
niliayu, line 193: Do you know what the behavior is on the back end if the client key provided already exists?

In some of the other services we check this before calling create, or otherwise call update instead. I'd either suggest following that sort of pattern (and updating the method name accordingly), or maybe just considering more simply checking & raising some nice message for users to understand if there's a client key name collision and what they should do.

python/lib/sift_py/calculated_channels/service.py
```diff
@@ -0,0 +1,338 @@
...
+
+        res = cast(
+            GetCalculatedChannelResponse,
+            self._calculated_channel_service_stub.GetCalculatedChannel(req),
+        )
+        return cast(CalculatedChannel, res.calculated_channel)
+
+    def list_calculated_channels(
+        self,
+        page_size: Optional[int] = None,
```
niliayu, line 80: It might be worth considering specifying a default page_size here and in the other list function. In some other places we've defaulted to `1_000` to make sure responses don't get too unwieldy.

python/lib/sift_py/calculated_channels/config.py
```diff
@@ -0,0 +1,104 @@
...
+
+    name: str
+    description: str = ""
+    expression: str
+    channel_references: List[
+        Union[ExpressionChannelReference, ExpressionChannelReferenceChannelConfig]
+    ]
+    units: Optional[str] = None
+    calculated_channel_id: Optional[str] = None
+    client_key: Optional[str] = None
```
niliayu, line 41: We generally want to encourage users to use client keys moving forward at this point, I would see if we could consider making the client key field non-optional. I don't think there's any reason a user wouldn't use a client key at this point. If I'm not mistaken, client key was kept optional in most of our other apis for backwards compatibility reasons, which I don't think we need to consider here since this service is new.

python/lib/sift_py/calculated_channels/service.py
```diff
@@ -0,0 +1,338 @@
...
+                for chan in resp.calculated_channel_versions
+            ],
+            resp.next_page_token,
+        )
+
+    def create_calculated_channel(
+        self, config: CalculatedChannelConfig
+    ) -> Tuple[CalculatedChannelConfig, CalculatedChannelValidationResult]:
+        """
+        Create a `CalculatedChannel` from a `CalculatedChannelConfig`. See
```
niliayu, line 165: It might be nice to add a little blurb describing what a client key is, in case it's a new concept to the user. We don't really describe what they are conceptually or how they're used in the main docs (which could be an oversight of the main docs as well)

niliayu (commented):
Got a few suggestions and nits! And per what we discussed offline, I think it's worth considering adding the YAML support here as well similar to what we've done for other services.

python/lib/sift_py/calculated_channels/config.py
```diff
@@ -0,0 +1,104 @@
...
+
+    name: NotRequired[str]
+    description: NotRequired[str]
+    units: NotRequired[str]
+    expression: NotRequired[str]
+    channel_references: NotRequired[
+        List[Union[ExpressionChannelReference, ExpressionChannelReferenceChannelConfig]]
+    ]
+    asset_ids: NotRequired[List[str]]
+    tag_ids: NotRequired[List[str]]
```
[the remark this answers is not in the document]
alexluck-sift: The reason I didn't do this is that we don't have a general tag or asset service, so we end up having to effectively re-implement this in each service. It would make more sense for us to create a service for each of those and use it from the other services rather than re-implement it each time we want to use it.

python/lib/sift_py/calculated_channels/service.py
alexluck-sift: The backend generates the following error:

```
grpc._channel._InactiveRpcError: <_InactiveRpcError of RPC that terminated with:
	status = StatusCode.ALREADY_EXISTS
	details = "unable to create calculated channel: already exists (87dd3b58-2cf3-477e-aac6-ff71720835f3)"
	debug_error_string = "UNKNOWN:Error received from peer  {grpc_message:"unable to create calculated channel: already exists (87dd3b58-2cf3-477e-aac6-ff71720835f3)", grpc_status:6, created_time:"2025-02-18T15:47:30.045883-08:00"}"
>
```

I think it makes sense to leave these types of errors up to the back-end (and if the errors aren't sufficiently descriptive, fix them there) as opposed to handling the validation logic on both the client and server.

Since I separated the create/update functions, in order to ensure result and user intent are explicit, I don't think we should infer that because it already exists, they want to update it. It may be that a user happened to accidentally choose a client key that is already in use and doesn't actually want to update the already existing calculated channel.

python/lib/sift_py/calculated_channels/service.py
alexluck-sift: The default values are defined by the back-end. My intent here in excluding a default value is to avoid having conflicting defaults and two "sources of truth".

python/lib/sift_py/calculated_channels/service.py
alexluck-sift: It is included in the CalculatedChannelConfig docs

python/lib/sift_py/calculated_channels/service.py
```diff
@@ -0,0 +1,338 @@
...
+        )
+        resp = cast(
+            CreateCalculatedChannelResponse,
+            self._calculated_channel_service_stub.CreateCalculatedChannel(req),
+        )
+        return self._calculated_channel_to_config(
+            cast(CalculatedChannel, resp.calculated_channel)
+        ), cast(CalculatedChannelValidationResult, resp.inapplicable_assets)
+
+    def revise_calculated_channel(
```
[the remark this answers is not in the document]
alexluck-sift: Updated

python/examples/calculated_channels/main.py
```diff
@@ -0,0 +1,35 @@
...
+            expression="$1 * 50",
+            channel_references=[
+                dict(channel_reference="$1", channel_identifier="mainmotor.velocity")
+            ],
+            asset_ids=["c8c724b9-6993-4423-9cc4-4c4ecdbc5751"],
+        )
+        chan, validation = service.create_calculated_channel(
+            config,
+        )
+        print(chan, validation)
```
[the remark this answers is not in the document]
alexluck-sift: updated

python/lib/sift_py/calculated_channels/service.py
niliayu: Ah, didn't realize the backend had one already. Works for me then!

python/lib/sift_py/calculated_channels/service.py
niliayu: Cool that makes sense to me -- I think the only reason I could think of for us catching and raising it ahead of time is if we wanted to give them some more explicit instruction on what the error means and what we'd recommend they do. In this case I think the error is decently clear if they understand how client keys work, but might be unclear if they are new to client keys and don't realize that client key exists == channel exists.

python/lib/sift_py/calculated_channels/config.py
niliayu: Chatted about this offline -- we're opening some tickets to improve the UI related to this.

niliayu (commented):
Looking great to me! Final couple of nits!

python/examples/calculated_channels/main.py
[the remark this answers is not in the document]
alexluck-sift: Deferring to possible refactor of examples with v1

python/lib/sift_py/calculated_channels/config.py
```diff
@@ -0,0 +1,111 @@
...
+    description: str = ""
+    expression: str
+    channel_references: List[
+        Union[ExpressionChannelReference, ExpressionChannelReferenceChannelConfig]
+    ]
+    units: Optional[str] = None
+    calculated_channel_id: Optional[str] = None
+    client_key: Optional[str] = None
+    asset_names: Optional[List[str]] = None
+    tags: Optional[List[str]] = None
```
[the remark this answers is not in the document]
alexluck-sift: updated

niliayu (approved):

PR #177, opened 2025-02-20
fix: ReportTemplate updates provide FieldMask keys based on what needs to be updated

niliayu:
This PR adds:
- A `ReportTemplateUpdate` class to allow users to easily specify update fields
- Updates the `ReportTemplateService` to specify only the fields actually being updated in the `FieldMask` passed back to the API
- Updates the ReportTemplate examples accordingly
- Updates the relevant unit tests

### DIFF
--- a/python/lib/sift_py/report_templates/config.py
+++ b/python/lib/sift_py/report_templates/config.py
@@ -27,6 +27,7 @@ class ReportTemplateConfig(BaseModel, AsJson):
 
     name: str
     template_client_key: str
+    template_id: Optional[str] = None  # read only
     organization_id: str = ""
     tags: Optional[List[str]] = None
     description: Optional[str] = None
--- a/python/lib/sift_py/report_templates/config.py
+++ b/python/lib/sift_py/report_templates/config.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-from datetime import datetime
 from typing import Any, List, Optional
 
 from pydantic import BaseModel, ConfigDict
@@ -26,13 +25,12 @@ class ReportTemplateConfig(BaseModel, AsJson):
     model_config = ConfigDict(arbitrary_types_allowed=True)
 
     name: str
-    template_client_key: str
-    template_id: Optional[str] = None  # read only
-    organization_id: str = ""
+    template_client_key: Optional[str] = None
+    template_id: Optional[str] = None
     tags: Optional[List[str]] = None
     description: Optional[str] = None
     rule_client_keys: List[str] = []
-    archived_date: Optional[datetime] = None
+    archived_date: Optional[bool] = False
 
     def as_json(self) -> Any:
         return self.model_dump_json()
@@ -54,8 +52,7 @@ class ReportTemplateUpdate(TypedDict):
 
     name: NotRequired[str]
     template_client_key: NotRequired[str]
-    organization_id: NotRequired[str]
     tags: NotRequired[List[str]]
     description: NotRequired[str]
     rule_client_keys: NotRequired[List[str]]
-    archived_date: NotRequired[datetime]
+    archived_date: NotRequired[bool]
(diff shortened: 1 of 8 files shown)


python/lib/sift_py/report_templates/service.py
```diff
@@ -33,20 +33,23 @@ class ReportTemplateService:
...
         """
         Create or update a report template via a ReportTemplateConfig. The config must contain a
         template_client_key, otherwise an exception will be raised. If a report template with the
         same client key exists, it will be updated. Otherwise, a new report template will be created.
         See `sift_py.report_templates.config.ReportTemplateConfig` for more information on available
         fields to configure.
         """
-        if not config.template_client_key:
-            raise Exception(f"Report template {config.name} requires a template_client_key")
+        if not config.template_client_key and not config.template_id:
```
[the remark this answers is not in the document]
alexluck-sift: This validation could be moved to after the `if updates:`...

python/lib/sift_py/report_templates/config.py
```diff
@@ -34,3 +36,26 @@ class ReportTemplateConfig(BaseModel, AsJson):
...
+    - `archived_date`: Updated date when the report template was archived.
+    """
+
+    name: NotRequired[str]
+    template_client_key: NotRequired[str]
+    organization_id: NotRequired[str]
+    tags: NotRequired[List[str]]
+    description: NotRequired[str]
+    rule_client_keys: NotRequired[List[str]]
+    archived_date: NotRequired[datetime]
```
alexluck-sift, line 61: In Calculated Channels i made this a boolean so that the user could simply archive/unarchive. Could be worth doing here too.

python/lib/sift_py/report_templates/config.py
```diff
@@ -34,3 +36,26 @@ class ReportTemplateConfig(BaseModel, AsJson):
...
+    - `tags`: Updated tags to associate with the report template.
+    - `description`: Updated description of the report template.
+    - `rule_client_keys`: Updated list of rule client keys associated with the report template.
+    - `archived_date`: Updated date when the report template was archived.
+    """
+
+    name: NotRequired[str]
+    template_client_key: NotRequired[str]
+    organization_id: NotRequired[str]
+    tags: NotRequired[List[str]]
```
alexluck-sift, line 58: you made a good point that this should probably be `tag_names` on Calculated Channels, but not a trivial update since the Config object uses `tags` and that would cause backwards compatibility issues.

python/lib/sift_py/report_templates/config.py
```diff
@@ -34,3 +36,26 @@ class ReportTemplateConfig(BaseModel, AsJson):
...
+    - `organization_id`: Updated organization ID that the report template belongs to.
+    - `tags`: Updated tags to associate with the report template.
+    - `description`: Updated description of the report template.
+    - `rule_client_keys`: Updated list of rule client keys associated with the report template.
+    - `archived_date`: Updated date when the report template was archived.
+    """
+
+    name: NotRequired[str]
+    template_client_key: NotRequired[str]
+    organization_id: NotRequired[str]
```
alexluck-sift, line 57: organization_id shouldn't be needed (and can't be changed as far as I know) since the API key is organization specific and will map the request to the correct org

python/lib/sift_py/report_templates/config.py
niliayu: Ugh yes that was an oversight on my original PR. Per offline discussion we'll address this on an upcoming refactor.

python/lib/sift_py/report_templates/config.py
niliayu: I like that, more simple/intuitive.

alexluck-sift (dismissed):

alexluck-sift (approved):

PR #187, opened 2025-03-06
python(bug): Add threaded and generator ingestion example

marcsiftstack:
Added an example that handles ingestion in a separate thread to prevent data generating from being blocked.
This example is a copy/paste of the `ingestion_with_python_config`. Changes were made to main.py and some small changes to simulator.py.

Also added an example that does true streaming using a generator. This example is a copy/paste of the `ingestion_with_python_config`. Changes were made to main.py and some small changes to simulator.py.
Example output:
```
$ python main.py
INFO:simulator:Beginning simulation for 'NostromoLV426' with run (7adb1d3a-79c6-43ab-95b1-bbc08c4f9d56)
INFO:__main__:Opening connection and streaming for 60s
INFO:__main__:Connection closed. Data streamed successfully
INFO:__main__:Opening connection and streaming for 60s
INFO:simulator:Completed simulation.
INFO:__main__:Simulation thread completed.
INFO:__main__:Connection closed. Data streamed successfully
```

### DIFF
--- a/python/lib/sift_py/ingestion/buffer.py
+++ b/python/lib/sift_py/ingestion/buffer.py
@@ -104,7 +104,7 @@ class BufferedIngestionService(Generic[T]):
                     self._buffer.append(req)
 
                 if len(self._buffer) >= self._buffer_size:
-                    self._flush()
+                    self._flush(self._buffer)
 
                 lhs_cursor = rhs_cursor
                 rhs_cursor = min(
@@ -139,7 +139,7 @@ class BufferedIngestionService(Generic[T]):
                     self._buffer.append(req)
 
                 if len(self._buffer) >= self._buffer_size:
-                    self._flush()
+                    self._flush(self._buffer)
 
                 lhs_cursor = rhs_cursor
                 rhs_cursor = min(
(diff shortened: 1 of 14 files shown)


python/lib/sift_py/ingestion/buffer.py
```diff
@@ -154,15 +154,18 @@ def flush(self):
 
         if self._flush_timer and self._lock:
             with self._lock:
-                self._flush()
+                # Save off the buffer so we don't need to hold onto the lock
+                buffer = self._buffer
```
solidiquis, line 158: I feel like we should explicitly copy/clone the list and assign to `buffer` and that buffer should be declared one scope up.. i.e. directly above `with self._lock`. I think what we're doing here is correct, but it hides some details that isn't immediately clear.

python/lib/sift_py/ingestion/buffer.py
```diff
@@ -154,15 +154,18 @@ def flush(self):
...
+            self._flush(self._buffer)
 
-    def _flush(self):
-        if len(self._buffer) > 0:
-            self._ingestion_service.ingest(*self._buffer)
-            self._buffer.clear()
+    def _flush(self, buffer: List[IngestWithConfigDataStreamRequest]):
+        if len(buffer) > 0:
+            self._ingestion_service.ingest(*buffer)
+            buffer.clear()
```
solidiquis, line 168: i don't think we need to clear the buffer as we no longer need it here

solidiquis (changes requested):

python/lib/sift_py/ingestion/buffer.py
marcsiftstack: Undid this change.

python/lib/sift_py/ingestion/buffer.py
marcsiftstack: Undid this change.

kevin-sift:
@marcsiftstack @solidiquis whats the state of this pr

solidiquis (approved):

PR #190, opened 2025-03-13
protos(feature): protobuf desc compatibility check

solidiquis:
# Summary

Add new endpoint to the `ProtobufDescriptorService` that allows users to check protobuf descriptor compatibility.

### DIFF
--- a/protos/sift/protobuf_descriptors/v2/protobuf_descriptors.proto
+++ b/protos/sift/protobuf_descriptors/v2/protobuf_descriptors.proto
@@ -6,7 +6,7 @@ import "google/api/annotations.proto";
 import "google/api/field_behavior.proto";
 import "protoc-gen-openapiv2/options/annotations.proto";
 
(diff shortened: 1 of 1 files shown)


protos/sift/protobuf_descriptors/v2/protobuf_descriptors.proto
```diff
@@ -6,7 +6,7 @@ import "google/api/annotations.proto";
 import "google/api/field_behavior.proto";
 import "protoc-gen-openapiv2/options/annotations.proto";
 
-
+option go_package = "azimuth/gen/protos/go/sift/protobuf_descriptors/v2;protobufdescriptorsv2pb";
```
bgfreeman7, line 9: Isn't this usually cleaned out with the `make sanitize` call?

solidiquis: you're correct. just fixed1

bgfreeman7 (approved):

PR #200, opened 2025-04-17
python(feat): Add support for TDMS files with time channel

marcsiftstack:
This change adds the ability to upload TDMS files where the timing information is included as a separate channel within each group. Example usage:
```
tdms_upload_service.upload("sample_data.tdms", asset_name, tdms_time_format=TdmsTimeFormat.TIME_CHANNEL)
```

Verification Evidence: SIFT-2882


python/lib/sift_py/data_import/tdms.py
```diff
@@ -87,31 +132,91 @@ def upload(
         if not posix_path.is_file():
             raise Exception(f"Provided path, '{path}', does not point to a regular file.")
 
-        with NamedTemporaryFile(mode="w", suffix=".csv") as temp_file:
-            valid_channels = self._convert_to_csv(path, temp_file.name, ignore_errors)
-            if not valid_channels:
-                raise Exception(f"No valid channels remaining in {path}")
-
-            csv_config = self._create_csv_config(
```
marcsiftstack, line 95: Moved this to _convert_to_csv since it will depend on the type of TDMS file being used.

python/lib/sift_py/data_import/tdms.py
```diff
@@ -42,6 +48,39 @@
...
+
+    See https://docs.siftstack.com/docs/data-model/assets-channels-runs#assets-and-channels
+
+    Args:
+        input_string: The string to sanitize.
+
+    Returns:
+        The sanitized string.
+    """
+    return input_string.translate(str.maketrans("", "", r"\"\`~|"))
```
marcsiftstack, line 81: Before this function would replace `/` with a space, not we just remove restricted characers. Will verify this change in behavior doesn't impact any existing users.

python/lib/sift_py/data_import/tdms.py
```diff
@@ -149,6 +250,9 @@ def normalize_name(channel_name: str) -> str:
                             "Set `ignore_errors` to True to skip channels without timing information."
                         )
 
+        if not valid_channels:
+            raise Exception(f"No valid channels remaining in {src_path}")
```
alexluck-sift, line 254: nit: "No valid channels found in..."

python/lib/sift_py/data_import/tdms.py
```diff
@@ -157,19 +261,187 @@ def normalize_name(channel_name: str) -> str:
 
             filtered_tdms_file = TdmsFile(f.name)
             df = filtered_tdms_file.as_dataframe(time_index=True, absolute_time=True)
-            df.to_csv(dst_path, encoding="utf-8")
+            df.to_csv(dst_file, encoding="utf-8")
+
+            # Close the file to make sure all contents are written.
+            # Required if using gzip compression to ensure all data
+            # is flushed: https://bugs.python.org/issue1110242
+            dst_file.close()
```
alexluck-sift, line 269: Could a `with` statement be used instead in the caller?

python/lib/sift_py/data_import/tdms.py
```diff
@@ -157,19 +261,187 @@ def normalize_name(channel_name: str) -> str:
...
+            updated_group_name = sanitize_string(group.name)
+            time_channels = get_time_channels(group)
+            if len(time_channels) != 1:
+                msg = (
+                    f"{group.name} contains more than one time channel"
+                    if len(time_channels) > 1
+                    else "no time channels"
+                )
+                if ignore_errors:
+                    print(f"{msg}. Skipping.")
```
alexluck-sift, line 329: Can we use logging instead of print statements?

python/lib/sift_py/data_import/tdms.py
```diff
@@ -157,19 +261,187 @@ def normalize_name(channel_name: str) -> str:
...
+                    )
+
+            time_channel = time_channels[0]
+            updated_channels = []
+            for channel in group.channels():
+                if channel == time_channel:
+                    updated_channel_name = TIME_CHANNEL_NAME
+                    data = to_datetime(channel.data).tz_localize("UTC")
+                    data = (
+                        data.strftime("%Y-%m-%dT%H:%M:%S.%f")
```
alexluck-sift, line 343: Make sure we document what timestamp format(s) we support clearly. There are a lot that I have seen in tdms over the years. For example, various types of epoch time.

python/lib/sift_py/data_import/tdms.py
```diff
@@ -157,19 +261,187 @@ def normalize_name(channel_name: str) -> str:
...
+                    continue
+                else:
+                    raise Exception(
+                        f"{msg}. Set `ignore_errors` to True to skip this group."
+                    )
+
+        if not valid_groups:
+            raise Exception(f"No valid groups remaining in {src_path}")
+
+        # Write the CSV manually instead of calling pandas.DataFrame.to_csv
```
alexluck-sift, line 388: Would `float_format` of `to_csv` allow this to be handled with Pandas?

python/lib/sift_py/data_import/tdms.py
```diff
@@ -157,19 +261,187 @@ def normalize_name(channel_name: str) -> str:
...
+                        channel.name: channel.data[i]
+                        for channel in updated_channels
+                    }
+                )
+        csv_writer.writerows(rows)
+
+        # Close the file to make sure all contents are written.
+        # Required if using gzip compression to ensure all data
+        # is flushed: https://bugs.python.org/issue1110242
+        dst_file.close()
```
alexluck-sift, line 411: same comment as above, use `with`?

python/lib/sift_py/data_import/tdms.py
marcsiftstack: The contextmanager for `NamedTemporaryFile` deletes the file when leaving the context, which we don't want to do.

python/lib/sift_py/data_import/tdms.py
marcsiftstack: Opened SIFT-3967 before to handle warnings consistently, would rather defer to that ticket to be consistent.

python/lib/sift_py/data_import/tdms.py
```diff
@@ -87,31 +132,91 @@ def upload(
...
     def _convert_to_csv(
-        self, src_path: Union[str, Path], dst_path: Union[str, Path], ignore_errors: bool
-    ) -> List[TdmsChannel]:
+        self,
+        src_path: Union[str, Path],
+        dst_file: TextIO,
+        asset_name: str,
+        prefix_channel_with_group: bool,
+        ignore_errors: bool,
+        run_name: Optional[str],
```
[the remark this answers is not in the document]
marcsiftstack: `Optional` just means that this can be `None`. See [here](https://docs.python.org/3/library/typing.html#typing.Optional):
> Note that this is not the same concept as an optional argument, which is one that has a default. An optional argument with a default does not require the Optional qualifier on its type annotation just because it is optional.

python/lib/sift_py/data_import/tdms.py
marcsiftstack: Sorry this comment is incorrect, the problematic call is `pandas.concat` when combining multiple`DataFrames` of different sizes. Will update.

python/lib/sift_py/data_import/tdms.py
marcsiftstack: 4132c726ec85c1ac2b3b5231fe9ef12e59f0ac05

python/lib/sift_py/data_import/tdms.py
marcsiftstack: 4132c726ec85c1ac2b3b5231fe9ef12e59f0ac05

python/lib/sift_py/data_import/tdms.py
marcsiftstack: 4132c726ec85c1ac2b3b5231fe9ef12e59f0ac05

python/lib/sift_py/data_import/tdms.py
marcsiftstack: 4132c726ec85c1ac2b3b5231fe9ef12e59f0ac05

alexluck-sift (approved):

PR #212, opened 2025-04-29
python(feat): Contextual channels support

niliayu:
Adds support to add contextual channels via the RuleService.


solidiquis (approved):
LGTM!

PR #219, opened 2025-05-15
python(feat): Add support for rule evaluation service and external rules

marcsiftstack:
This PR adds the `RuleEvaluationService` and support for "external" rules.

See examples for more detailed example, but here is the basic usage:
```
rule_configs = load_nostromos_lv_426_rule_configs()

with use_sift_channel(sift_channel_config) as channel:
    rule_eval_service = RuleEvaluationService(channel)
    report = rule_eval_service.evaluate_external_rules(run_id, rule_configs)
    finished = report.wait_until_done(timeout=60)
```

Verification: All examples ran


python/lib/sift_py/rule_evalutation/service.py
```diff
@@ -0,0 +1,261 @@
...
+            The EvaluateRulesPreviewResponse object.
+        """
+        rule_ids = self._rule_service.create_external_rules(rules)
+        return self.preview_against_run(run_id, rule_ids)
+
+    def preview_external_rules_from_yaml(
+        self,
+        run_id: str,
+        paths: List[Path],
+        named_expressions: Optional[Dict[str, str]] = None,
```
niliayu, line 196: We have a ticket to deprecate named expressions, so maybe we can just not support it here if it's not needed? Otherwise we'll just need to remove this scope when we get to that ticket which isn't too bad either.

python/lib/sift_py/rule_evalutation/report.py
```diff
@@ -0,0 +1,58 @@
...
+
+NOT_RUNNING_STATUS = [
+    ReportRuleStatus.REPORT_RULE_STATUS_CANCELED,
+    ReportRuleStatus.REPORT_RULE_STATUS_ERROR,
+    ReportRuleStatus.REPORT_RULE_STATUS_FAILED,
+    ReportRuleStatus.REPORT_RULE_STATUS_FINISHED,
+]
+
+
+class Report:
```
niliayu, line 17: This is sort of a minimal report service, maybe it'd make sense just to put it in its own sift_py/reports/service.py in case we want to built more off of it later on?

python/lib/sift_py/rule_evalutation/service.py
```diff
@@ -0,0 +1,261 @@
...
+        report_name: str = "",
+    ) -> Report:
+        """Evaluate a set of rules against assets.
+
+        Args:
+            asset_names: The list of assets to run against.
+            start_time: The start time to evaluate.
+            end_time: The end time to evaluate.
+            rules: Either a ReportTemplateConfig, a list of RuleConfigs, or a list of
+                RuleIdentifier's (typically from `RuleService.create_external_rules`).
```
niliayu, line 90: Nit, extra apostrophe I think (RuleIdentifiers)

python/lib/sift_py/rule_evalutation/service.py
```diff
@@ -0,0 +1,261 @@
...
+        """
+        asset_time_range = AssetsTimeRange(
+            assets=NamedResources(names=Names(names=asset_names)),
+            start_time=to_timestamp_pb(start_time),
+            end_time=to_timestamp_pb(end_time),
+        )
+        eval_kwargs = self._get_rules_kwargs(rules)
+
+        req = EvaluateRulesRequest(
+            report_name=report_name,
```
niliayu, line 104: What does the report look like if the report_name is empty?

python/lib/sift_py/rule_evalutation/service.py
```diff
@@ -0,0 +1,261 @@
...
+            **eval_kwargs,
+        )
+        res = cast(EvaluateRulesResponse, self._rule_evaluation_stub.EvaluateRules(req))
+
+        return Report(self._channel, res.report_id)
+
+    def evaluate_against_assets(
+        self,
+        asset_names: List[str],
+        start_time: datetime,
```
niliayu, line 78: Could be nice to match the arg of `to_timestamp_pb` and allow the flexibility to provide other timestamp formats here `Union[datetime, str, int]`

python/lib/sift_py/rule_evalutation/report.py
```diff
@@ -0,0 +1,58 @@
...
+            True if the report finished within the timeout.
+        """
+        start_time = time.time()
+        while True:
+            if timeout is not None:
+                if time.time() - start_time > timeout:
+                    return False
+
+            result = self.get_results()
+            all_done = all([summary.status in NOT_RUNNING_STATUS for summary in result.summaries])
```
niliayu, line 53: Just a thought, maybe it'd be nice to provide a separate helper to make it easy to check the status, I could just imagine that to be useful if somebody is wondering/debugging long running reports and wanting more visibility into what's going on from here.

niliayu (commented):
I think this looks great! Just a few thoughts/questions/small nits about a few things

python/lib/sift_py/rule_evalutation/service.py
marcsiftstack: I'd rather leave it. Unclear when we will implement that ticket so I'd rather not have this service and the other rule service in different states.

python/lib/sift_py/rule_evalutation/service.py
marcsiftstack: Sift will use a default report name which is the Run title.

python/lib/sift_py/rule_evalutation/report.py
marcsiftstack: What are you envisioning for this helper function? `get_results` is public, so they can still call that to get the current status.

python/examples/rule_evaluation/external_rules/python_config/.env-example
```diff
@@ -0,0 +1,3 @@
+SIFT_API_URI=""
+SIFT_API_KEY=""
+ASSET_NAME=""
```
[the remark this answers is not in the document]
marcsiftstack: 60a915ee5aaa4ede90ace5ee8b2dea53c8bf358f

python/examples/rule_evaluation/external_rules/yaml_config/.env-example
```diff
@@ -0,0 +1,3 @@
+SIFT_API_URI=""
+SIFT_API_KEY=""
+ASSET_NAME=""
```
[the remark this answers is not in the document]
marcsiftstack: 60a915ee5aaa4ede90ace5ee8b2dea53c8bf358f

python/lib/sift_py/rule_evalutation/report.py
marcsiftstack: 60a915ee5aaa4ede90ace5ee8b2dea53c8bf358f

python/lib/sift_py/rule_evalutation/service.py
marcsiftstack: 60a915ee5aaa4ede90ace5ee8b2dea53c8bf358f

python/lib/sift_py/report/_report_test.py
marcsiftstack, line 1: For some reason I had to rename this from `_service_test.py` to `_report_test.py`. I was getting this error:
```
================================================================================================================================== ERRORS ==================================================================================================================================
______________________________________________________________________________________________________ ERROR collecting lib/sift_py/rule_evaluation/_service_test.py _______________________________________________________________________________________________________
import file mismatch:
imported module '_service_test' has this __file__ attribute:
  /Users/mjulien/code/sift/python/lib/sift_py/report/_service_test.py
which is not the same as the test file we want to collect:
  /Users/mjulien/code/sift/python/lib/sift_py/rule_evaluation/_service_test.py
HINT: remove __pycache__ / .pyc files and/or use a unique basename for your test file modules
```

Not sure why the other `_service_test.py` files in the other directories don't have any issues.

python/lib/sift_py/rule_evalutation/report.py
niliayu: Oh yeah, using `get_results()` looks fine to me... don't remember exactly what I was thinking when I wrote this. lgtm!

niliayu:
Good stuff, LGTM! Last nit, mind changing the PR title to the conventional format before we merge?

PR #221, opened 2025-05-22
python(feat): Allow creating multiple new flows at the same time

marcsiftstack:
* Added the ability to create multiple new flows in the same call of `try_crew_flow` since the underlying API call already supports that.
* Add plural versions of the existing functions to match the naming of other functions in this class.

Verification:
Ran this modified example:
```
with use_sift_channel(sift_channel_config) as channel:
    tlm_config = TelemetryConfig.try_from_yaml("telemetry.yml")
    cmd_config = TelemetryConfig.try_from_yaml("command.yml")

    empty_config = TelemetryConfig(
        asset_name="NostromoLV426", ingestion_client_key=tlm_config.ingestion_client_key
    )
    ingestion_service = IngestionService(channel, empty_config)

    ingestion_service.try_create_flow(*(tlm_config.flows + cmd_config.flows))

    print(ingestion_service.flow_configs_by_name)
```
Output:
```
{ 
  'readings': <sift_py.ingestion.flow.FlowConfig object at 0x104059010>, 
  'voltage': <sift_py.ingestion.flow.FlowConfig object at 0x1040c4b90>, 
  'gpio_channel': <sift_py.ingestion.flow.FlowConfig object at 0x1040c4910>, 
  'logs': <sift_py.ingestion.flow.FlowConfig object at 0x1040b4510>, 
  'cmds': <sift_py.ingestion.flow.FlowConfig object at 0x1040b43e0>
}
```
flows from both telem configs appear as expected.

### DIFF
--- a/python/lib/sift_py/ingestion/_internal/ingest.py
+++ b/python/lib/sift_py/ingestion/_internal/ingest.py
@@ -337,26 +337,33 @@ class _IngestionServiceImpl:
             end_stream_on_validation_error=self.end_stream_on_error,
         )
 
-    def try_create_flow(self, flow_config: FlowConfig):
+    def try_create_flow(self, *flow_configs: FlowConfig):
         """
         Tries to create a new flow at runtime. Will raise an `IngestionValidationError` if there already exists
         a flow with the name of the `flow_config` argument.
         """
-
-        if flow_config.name in self.flow_configs_by_name:
-            raise IngestionValidationError(
-                f"There is already a flow with name '{flow_config.name}'."
-            )
+        for flow_config in flow_configs:
+            if flow_config.name in self.flow_configs_by_name:
+                raise IngestionValidationError(
+                    f"There is already a flow with name '{flow_config.name}'."
+                )
 
         create_flow_configs(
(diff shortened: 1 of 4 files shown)


python/lib/sift_py/ingestion/_internal/ingest.py
```diff
@@ -337,36 +337,50 @@ def create_ingestion_request(
             end_stream_on_validation_error=self.end_stream_on_error,
         )
 
-    def try_create_flow(self, flow_config: FlowConfig):
+    def try_create_flow(self, *flow_configs: FlowConfig):
```
niliayu, line 340: In order for this to be truly backwards compatible, wouldn't we want this arg to remain the same in case users have code that directly names it, like `ingestion_service.try_create_flow(flow_config=my_flow_config)` ?

marcsiftstack: Yeah that is true. I assumed no one would be doing that since this is the only argument for this function. I can change it back.

niliayu (approved):

PR #235, opened 2025-06-17
rust(feat): sift-stream-bindings via pyo3

niliayu:
This PR adds bindings for SiftStream via pyo3, and a github action to publish a pypi package with these bindings. This at its core exposes the `SiftStream::send` and `SiftStream::send_requests` functionality to python via the `SiftStreamBuilder` needed to produce a `SiftStream` object.


.github/workflows/maturin_release.yaml
```diff
@@ -0,0 +1,128 @@
+name: CI
```
niliayu, line 1: This file was initially produced using the [maturin generate-ci command,](https://www.maturin.rs/distribution.html#github-actions) but I made some manual edits to try and match the versions and architectures we use to build our python client lib: https://github.com/sift-stack/sift/blob/main/.github/workflows/python_build.yaml#L44-L48

.github/workflows/maturin_release.yaml
```diff
@@ -0,0 +1,128 @@
...
+        with:
+          command: sdist
+          args: --out dist
+      - name: Upload sdist
+        uses: actions/upload-artifact@v4
+        with:
+          name: wheels-sdist
+          path: dist
+
+  release:
```
niliayu, line 103: Maturin should be packaging all the binaries into the whl, so as I understand it we don't need to distribute the rust crate separately. 
See: 
- https://www.maturin.rs/tutorial.html#build-and-install-the-module-with-maturin-develop
- https://www.maturin.rs/distribution.html

rust/crates/sift_stream_bindings/src/stream/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+    fn from(flow: FlowPy) -> Self {
+        flow.inner
+    }
+}
+
+// PyO3 Method Implementations
+#[pymethods]
+impl SiftStreamPy {
+    pub fn send(&mut self, py: Python, flow: FlowPy) -> PyResult<Py<PyAny>> {
+        let mut inner = self.inner.lock().unwrap().take().unwrap();
```
danny-sift, line 52: Looks like there isn't a great way to avoid the `Arc::new(Mutex::new(Some(...)))` here since we likely do want to keep re-using the underlying `SiftStream` across calls. I think the implications are:
1. need to re-assign the sift_stream variable on the Python side to the result of this call, which may be a bit awkward in Python (ah looks like you're already doing the examples you shared)
  - I think one way we could possibly avoid the re-assignment requirement is to use an async-safe mutex (e.g. [tokio::sync::Mutex](https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html)).
    - there are some downside to using an async-safe mutex, usually around perf – i.e. the lock is held until completion of the async function – but we have this limitation already
  - this would allow you to get rid of the `Option` wrapper which is what necessitates the fact that this function needs to take ownership of the SiftStream and then return ownership back to the caller
  - seems like we don't currently have a direct dependency on tokio here, so it might not be worth it, just something to consider 🤷 
2. there can only be one active call to `send` at a time
  - looks like this is already a limitation of the `SiftStream` since it's `send` method requires a mutable reference, so nothing we can do about it here without changes to the underlying `sift_stream` crate

danny-sift (commented):
Overall looks pretty good from a Rust best practices perspective – had a few comments around concurrency stuff, but the main thing is that I think we should avoid calls to `unwrap()` wherever possible and aim to return errors instead. If we ever have a bug (or someone uses the lib incorrectly) it'd be great to get an error with context that can potentially be retried instead of crashing the program.

Also, it'd be great to have a few tests around this – though it'd probably best to write those tests in Python so that we know the bindings work as expected. Maybe you can re-purpose the examples you shared with me into tests (can be done in a follow-up PR though, imo).

rust/crates/sift_stream_bindings/src/stream/builder.rs
```diff
@@ -0,0 +1,88 @@
...
+    run_id: Option<String>,
+}
+
+// PyO3 Method Implementations
+#[pymethods]
+impl SiftStreamBuilderPy {
+    #[new]
+    pub fn new(uri: &str, apikey: &str) -> Self {
+        Self {
+            inner: Arc::new(Mutex::new(Some(SiftStreamBuilder::new(
```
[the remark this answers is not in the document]
niliayu: Hm I think that makes sense, I'll give that a try and report back!

rust/crates/sift_stream_bindings/src/stream/mod.rs
niliayu: So I could be missing something here, but the problem I ran into when looking into implementing this without the `Option` wrapper is that `finish()` still seems to need ownership since it's intentionally shutting down the stream and making it no longer usable, if I'm understanding right. Let me know if you see some way around that though.

danny-sift: Ah didn't notice that, that makes sense

rust/crates/sift_stream_bindings/src/stream/mod.rs
```diff
@@ -0,0 +1,164 @@
...
+        }
+    }
+}
+
+impl From<SiftStreamPy> for SiftStream<IngestionConfigMode> {
+    fn from(stream: SiftStreamPy) -> Self {
+        match stream.inner.lock() {
+            Ok(mut guard) => match guard.take() {
+                Some(stream) => stream,
+                None => panic!("SiftStreamPy inner was None"),
```
danny-sift, line 41: nit: these panics would also cause a crash if we ever hit them. I'm curious what the lint/clippy errors that you ran into were? If there's no way around it, then I think this is still better than calling `unwrap()` since we'd get a bit more context with the crash.

danny-sift (dismissed):
lgtm – just have an optional nit

.github/workflows/maturin_release.yaml
```diff
@@ -0,0 +1,128 @@
...
+    name: Release
+    runs-on: ubuntu-latest
+    if: ${{ startsWith(github.ref, 'refs/tags/') || github.event_name == 'workflow_dispatch' }}
```
alexluck-sift, line 106: For the release process tagging, what is the plan here? Shouldn't we filter on a different format to make sure it is a sift-stream-bindings tag?

rust/crates/sift_stream_bindings/src/stream/mod.rs
niliayu: Yeah this could have been the way I was trying to implement it, but I was trying to have this return an error type instead of the panic, which changed the `match` pattern to `Some(stream) => Ok(stream)` (with `None => Err(...)` on the next line), and that `Ok(stream)` return type seemed to be incompatible with the return type expected of the `From` trait, so it seemed like I needed to change the implementation in order to use that pattern

danny-sift: Ah I see, yeah that makes sense

danny-sift (approved):

PR #236, opened 2025-06-18
Python: high level client initial PR

alexluck-sift:
See:
[Sift Client Internal Architecture](https://github.com/sift-stack/sift/blob/python/high-level-client-init/python/lib/sift_client/_internal/README.md)
[Sift Client Contributing Guidelines](https://github.com/sift-stack/sift/blob/python/high-level-client-init/python/lib/sift_client/_internal/CONTRIBUTING.md)

[Sift Client Usage](https://github.com/sift-stack/sift/blob/python/high-level-client-init/python/lib/sift_client/README.md)

Code is still experimental and subject to continued change/iteration. If imported, a warning will be generated to warn users that it is experimental.


python/examples/sift_client/assets/main.py
```diff
@@ -0,0 +1,27 @@
...
+        api_key=apikey,
+        rest_url="https://api.development.siftstack.com",
+        grpc_url="https://grpc-api.development.siftstack.com",
+    )
+
+    # Find an asset
+    asset = sift.assets.find(name_contains="Nostromo")
+    print(asset.tags)
+    asset.update(dict(tags=["simulator", "my_new_tag"]))
+    print(asset)
```
ian-sift, line 18: could you add metadata example and evidence? I'm curious with the AssetUpdate preserves existing data or you need to construct the update w/ the existing data you want to keep or it will be deleted.

ian-sift (commented):
First pass. Not sure how stringent you want to be on testing here or if you've got tickets you're deferring that work to. I do think at least evidence of one of the examples would be nice.

python/lib/sift_client/types/base.py
```diff
@@ -0,0 +1,104 @@
...
+    def _update(self, other: BaseType[ProtoT, SelfT]) -> BaseType[ProtoT, SelfT]:
+        """Update this instance with the values from another instance"""
+        # This bypasses the frozen status of the model
+        for key in other.model_fields.keys():
+            if key in self.model_fields:
+                self.__dict__.update({key: getattr(other, key)})
+        return self
+
+
+# TODO: how to handle nulling fields (undefined or something?)
```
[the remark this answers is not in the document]
alexluck-sift: There are really 3 possible states we need to represent with these pydantic model fields:

* a value
* a null value
* not set

In the first two cases, these should be included in the field mask. In the latter, it would be omitted. The current implementation only accounts for not set and a value. This also may depend a bit on the gRPC autogenerated code which has some issues where Optional fields aren't respected and treated as required.

python/examples/sift_client/assets/main.py
alexluck-sift: That isn't fully figured out yet, I need to finish the Metadata types and low/high level clients.

niliayu (commented):
I've just taken a high level pass to get my head around the structure and would like to take a deeper look at some things to get my head around everything, but so far this is pretty sweet! Left some questions and comments for consideration in the meantime

python/lib/sift_client/errors.py
```diff
@@ -0,0 +1,29 @@
...
+
+_warned_already = False
+
+
+def _sift_client_experimental_warning():
+    global _warned_already
+    if _warned_already:  # Prevent sending this warning too many times
+        return
+    _warned_already = True
+    warnings.warn(
```
nathan-sift, line 26: Very minor comment, but by default warnings will only show once, so most of the surrounding code shouldn't be necessary.

python/lib/sift_client/resources/sync_stubs/__init__.pyi
```diff
@@ -0,0 +1,162 @@
...
+from sift_client.types.asset import Asset, AssetUpdate
+
+class AssetsAPI:
+    """
+    Sync counterpart to `AssetsAPIAsync`.
+
+
+    High-level API for interacting with assets.
+
+    This class provides a Pythonic, notebook-friendly interface for interacting with the AssetsAPI.
```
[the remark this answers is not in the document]
alexluck-sift: Yes, this is one of the many things that will be documented in the docs pass. However, this shouldn't be appreciably slower for most use cases except for streaming.

python/lib/sift_client/errors.py
alexluck-sift: ah excellent, good callout

python/lib/sift_client/types/base.py
```diff
@@ -0,0 +1,106 @@
...
+    def _from_proto(cls, proto: ProtoT, sift_client: SiftClient | None = None) -> SelfT: ...
+
+    def _apply_client_to_instance(self, client: SiftClient) -> None:
+        # This bypasses the frozen status of the model
+        self.__dict__["_client"] = client
+
+    def _update(self, other: BaseType[ProtoT, SelfT]) -> BaseType[ProtoT, SelfT]:
+        """Update this instance with the values from another instance"""
+        # This bypasses the frozen status of the model
+        for key in other.model_fields.keys():
```
ian-sift: Apparently this is deprecated and pydantic wants you to just access __pydantic_fields__ directly? https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_fields

alexluck-sift: updated to use the class

ian-sift (approved):

python/lib/sift_client/_internal/README.md
```diff
@@ -0,0 +1,193 @@
...
+
+The architecture consists of:
+
+1. **Low-level clients**: Pure async implementations that wrap gRPC stubs
+2. **High-level APIs**: Both sync and async versions that provide user-friendly interfaces
+3. **Dedicated event loop**: A separate event loop in its own thread for gRPC async operations
+
+### Design Decisions
+
+#### Why a Dedicated Event Loop?
```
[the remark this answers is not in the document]
niliayu: Cool, super helpful!

python/lib/sift_client/resources/sync_stubs/__init__.pyi
niliayu, line 1: I had this thought when writing the type stubs for sift-stream-bindings -- think there's any value in including a mypy `stubtest` step in ci or anything?

niliayu (commented):
I think things are looking really nice here, just leaving open a few comment threads to make sure they get addressed (or have a plan to be if needed) before moving on!

python/lib/sift_client/resources/sync_stubs/__init__.pyi
alexluck-sift: good idea, I will add

python/lib/sift_client/client.py
```diff
@@ -0,0 +1,59 @@
...
+            rest_client = RestClient(RestConfig(rest_url, api_key))
+        else:
+            raise ValueError(
+                "Invalid connection configuration. Please provide api_key, grpc_uri and rest_uri or a connection_config."
+            )
+
+        WithGrpcClient.__init__(self, grpc_client=grpc_client)
+        WithRestClient.__init__(self, rest_client=rest_client)
+
+        self.ping = PingAPI(self)
```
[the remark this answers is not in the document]
alexluck-sift: used a namedtuple

niliayu (approved):
🎉🎉🎉

PR #246, opened 2025-06-23
python(feat): Add public method for creating new run

nathan-sift:
Added an optional flag to attach_run() named `force_new`, which when true, forces creation of a new run even if an existing run of the same name exists. Defaults to False to match the existing behavior of the function.

Equivalent to running sift_py.ingestion._internal.run.create_run(), but consistent with the existing public interface.

**Verification:**
Ran the following simplified code snippet to verify a new run is attached despite a duplicate run name.

```
def test_attach_run(grpc_channel, config, run_name, force_new=False):
    logging.info(f"Attach run {run_name}")
    ingestion_service = IngestionService(grpc_channel, config)
    ingestion_service.attach_run(grpc_channel, run_name, force_new=force_new)
    write_data(ingestion_service)
    logging.info(f"Now using run_id {ingestion_service.run_id}")

with use_sift_channel(credentials) as grpc_channel:
        test_attach_run(grpc_channel, "TestRun")
        test_attach_run(grpc_channel, "TestRun")
        test_attach_run(grpc_channel, "TestRun", force_new=True)
        test_attach_run(grpc_channel, "TestRun")

```

Output:
```
Existing runs:
3f3aba34-737e-4e3d-996b-24bef14d7cb4: TestRun
e00eb986-4c25-4080-a17a-3693bd3efcef: TestRun

2025-06-23 16:44:54,746 - INFO - Attach run TestRun with force_new=False
2025-06-23 16:44:55,113 - INFO - Now using run_id "3f3aba34-737e-4e3d-996b-24bef14d7cb4"
2025-06-23 16:44:55,113 - INFO - Attach run TestRun with force_new=False
2025-06-23 16:44:55,501 - INFO - Now using run_id "3f3aba34-737e-4e3d-996b-24bef14d7cb4"
2025-06-23 16:44:55,502 - INFO - Attach run TestRun with force_new=True
2025-06-23 16:44:55,840 - INFO - Now using run_id "8af50950-497f-43d7-96a2-50ee73b266ff"
2025-06-23 16:44:55,840 - INFO - Attach run TestRun with force_new=False
2025-06-23 16:44:56,181 - INFO - Now using run_id "8af50950-497f-43d7-96a2-50ee73b266ff"

Current runs:
8af50950-497f-43d7-96a2-50ee73b266ff: TestRun
3f3aba34-737e-4e3d-996b-24bef14d7cb4: TestRun
e00eb986-4c25-4080-a17a-3693bd3efcef: TestRun
```

### DIFF
--- a/python/lib/sift_py/ingestion/_internal/ingest.py
+++ b/python/lib/sift_py/ingestion/_internal/ingest.py
@@ -136,39 +136,19 @@ class _IngestionServiceImpl:
         organization_id: Optional[str] = None,
         tags: Optional[List[str]] = None,
         metadata: Optional[Dict[str, Union[str, float, bool]]] = None,
+        force_new: bool = False,
     ):
         """
         Retrieve an existing run or create one to use during this period of ingestion.
-        """
-        run_id = get_run_id_by_name(channel, run_name)
 
-        if run_id is not None:
-            self.run_id = run_id
-            return
-
-        self.run_id = create_run(
-            channel=channel,
-            run_name=run_name,
-            description=description or "",
-            organization_id=organization_id or "",
-            tags=tags or [],
-            metadata=metadata,
-        )
-
-    def attach_new_run(
-        self,
-        channel: SiftChannel,
-        run_name: str,
-        description: Optional[str] = None,
-        organization_id: Optional[str] = None,
-        tags: Optional[List[str]] = None,
-        metadata: Optional[Dict[str, Union[str, float, bool]]] = None,
-    ):
+        Include `force_new=True` to force the creation of a new run, which will allow creation of a new run using an existing name.
         """
-        Create a new run to use during this period of ingestion.
-        Will generate a new `run_id` even if a run with the same name already exists, allowing multiple runs with the same name.
-        Prefer `attach_run` unless intent is to create a new run regardless of existing run names.
-        """
+        if not force_new:
+            run_id = get_run_id_by_name(channel, run_name)
+
+            if run_id is not None:
+                self.run_id = run_id
+                return
 
         self.run_id = create_run(
             channel=channel,
--- a/python/lib/sift_py/ingestion/_internal/ingest.py
+++ b/python/lib/sift_py/ingestion/_internal/ingest.py
@@ -155,6 +155,30 @@ class _IngestionServiceImpl:
             metadata=metadata,
         )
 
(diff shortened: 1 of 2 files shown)


python/lib/sift_py/ingestion/_internal/ingest.py
```diff
@@ -155,6 +155,30 @@ def attach_run(
             metadata=metadata,
         )
 
+    def attach_new_run(
```
niliayu, line 158: This is just my 2 cents, but I think if we were to create a `RunService` that would let people call `create_run` arbitrarily outside of an `IngestionService` instance, that would be a more generalizable approach to this. Then users would end up with something like
```
run_service = RunService(channel)
ingestion_service = IngestionService(channel, telem_config)

new_run_id = run_service.create(new_run_name)
ingestion_service.run_id = new_run_id
```
I'm not sure if there was a particular reason this wasn't done previously other than not wanting the overhead of having to write a public `RunService` at the time. The other reason I think having a `RunService` would be nice is so we can enable users to easily access the other parts of the runs api, to make updates to their run metadata and so on. Open to other thoughts here though, especially with the upcoming work on the client you have @alexluck-sift .

alexluck-sift: I'm of the opinion that we are going to be rewriting this anyway, so we should avoid any negative work. If the ask is focused around exposing the internal method, this accomplishes that. If we have strong need for exposing additional areas of the API via services (in the near-near term) then we can expand this to a run service, but I don't think there has been asks for that.

alexluck-sift (dismissed):

python/lib/sift_py/ingestion/_internal/ingest.py
niliayu: In that case, what is your opinion on the ergonomics of having a separate method here vs something like a `new=` flag on the existing attach_run method? Where `new=True` would initiate the same behavior encapsulated by `attach_new_run()`

nathan-sift: I just had the same thought, and agree that allowing a flag on the existing function makes more sense. Maybe `force_new=True`?

niliayu: Sgtm!

niliayu (approved):

PR #250, opened 2025-06-24
rust(enhancement): add unique ID to each instance of sift stream

solidiquis:
## Changes

Each instance of `SiftStream` now comes with a unique ID that gets emitted whenever `SiftStream` emits logs. This is useful for environments that stream with multiple instances of `SiftStream`.

### DIFF
--- a/rust/crates/sift_stream/src/stream/mode/ingestion_config.rs
+++ b/rust/crates/sift_stream/src/stream/mode/ingestion_config.rs
@@ -165,7 +165,7 @@ impl SiftStream<IngestionConfigMode> {
 
         #[cfg(feature = "tracing")]
         tracing::info!(
-            streaming_id = format!("{sift_stream_id}"),
+            sift_stream_id = sift_stream_id.to_string(),
             "Sift streaming successfully initialized"
         );
 
@@ -213,7 +213,7 @@ impl SiftStream<IngestionConfigMode> {
         ) else {
             #[cfg(feature = "tracing")]
             tracing::warn!(
-                sift_stream_id = format!("{}", self.mode.sift_stream_id),
+                sift_stream_id = self.mode.sift_stream_id.to_string(),
                 values = format!("{message:?}"),
                 "encountered a message whose channel values do not match any configured flows"
             );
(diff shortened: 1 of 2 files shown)


rust/crates/sift_stream/src/stream/mode/ingestion_config.rs
```diff
@@ -158,13 +164,17 @@ impl SiftStream<IngestionConfigMode> {
         begin_checkpoint_notifier.notify_one();
 
         #[cfg(feature = "tracing")]
-        tracing::info!("Sift streaming successfully initialized");
+        tracing::info!(
+            streaming_id = format!("{sift_stream_id}"),
```
danny-sift, line 168: nit: `UUID`s have a `to_string` method which we can use instead of `format!`

danny-sift, line 168: should we replace `streaming_id` -> `sift_stream_id` to match other logs?

rust/crates/sift_stream/src/stream/mode/ingestion_config.rs
```diff
@@ -238,7 +254,10 @@ impl SiftStream<IngestionConfigMode> {
             .is_err_and(|e| e.kind() == ErrorKind::BackupLimitReached)
         {
             #[cfg(feature = "tracing")]
-            tracing::info!("backup size limit reached - forcing checkpoint");
+            tracing::info!(
+                streaming_id = format!("{}", self.mode.sift_stream_id),
```
danny-sift, line 258: `streaming_id` -> `sift_stream_id` here too

danny-sift (commented):
Generally lgtm, the main thing is to be consistent wrt logging as either `streaming_id` or `sift_stream_id`.

I wonder if this is something we can use the `tracing` crate's concept of a span for? It looks like by default, they don't work in an async context, but there is some support:
https://docs.rs/tracing/latest/tracing/span/struct.Span.html#in-asynchronous-code

rust/crates/sift_stream/src/stream/mode/ingestion_config.rs
solidiquis: yes absolutely thanks for catching. Updated.

solidiquis: Was wondering if there was a lint rule we could enable for this but looks a bit contentious lol https://github.com/rust-lang/rust-clippy/issues/3753

solidiquis: Fixed

danny-sift (approved):

PR #261, opened 2025-07-01
python(feat): Add HDF5 upload service

nathan-sift:
Adds an `Hdf5UploadService` to allow ingestion of HDF5 files. Also adds an `Hdf5Config` used to define the incoming data.

```
hdf5_upload_service = Hdf5UploadService(rest_config)
hdf5_upload_service.upload("sample_data.h5", hdf5_config)
```

Uses the following config format:
```
{
  "asset_name": string,
  "run_name": string,
  "run_id": string,
  "time": {
    "format": string              // Required. See "Time Formats" below for options.
    "relative_start_time": string // Required if "format" is a relative time format.
                                  // Must be in RFC3339 format.
                                  // Will be added to all relative times to
                                  // generate an absolute timestamp.
  },
  "data": [
    {
      "name": string,
      "description": string,
      "units": string,
      
      "time_dataset": string,   // Example: "/sensors/A/timestamps",
      "time_column": number, // Optional: Column to extract from for 2D arrays. Default value = 1
      "value_dataset": string,
      "value_column": number, // Optional: Column to extract from for 2D arrays. Default value = 1
      
      "enum_types": [       // Optional. Only valid if data type is an enum
        {
          "key": number,     // The raw enum value
          "name": string     // The display value for the enum value
        },
      "bit_field_elements: [   // Optional. Only valid if data type is a bit field element
        {
          "index": number,    // Starting index of the bit_field
          "name": string,     // Name of the bit field element.
          "bit_count": number // Number of bits in the element
        },
    },
    ...
  ]
}
```

### Verification
See included unit testing. 

Example includes ingestion of a sample hdf5 file

**HDFView of file**
Screenshot 2025-07-01 at 2 02 15 PM

**Uploaded to Sift**
Screenshot 2025-07-01 at 2 08 30 PM


python/lib/sift_py/data_import/hdf5.py
```diff
@@ -0,0 +1,281 @@
...
+    with h5py.File(src_path, "r", libver="latest", swmr=True) as h5f:
+        for data_cfg in hdf5_config._hdf5_config.data:
+            df = _extract_hdf5_data_to_dataframe(h5f, data_cfg)
+            data_frames.append(df)
+
+    # Merge polars dataframes, sort by timestamp, then write to a temp file
+    # Could write csv without headers, but keeping for debugging purposes
+    # If header removed, need to update 'first_data_row' in _create_csv_config to 1
+    merged_df = data_frames[0]
+    if len(merged_df) > 1:
```
marcsiftstack, line 131: Could we use [concat](https://docs.pola.rs/api/python/stable/reference/api/polars.concat.html) to simplify this?

python/lib/sift_py/data_import/hdf5.py
```diff
@@ -0,0 +1,281 @@
...
+        dst_file: The output CSV file.
+        hdf5_config: The HDF5 config.
+
+    Returns:
+        The CSV config for the import.
+    """
+
+    csv_cfg = _create_csv_config(hdf5_config)
+    merged_df = _convert_hdf5_to_dataframes(src_path, hdf5_config)
+    merged_df.write_csv(dst_file.name)
```
marcsiftstack, line 103: You should be able to pass in the `dst_file` object. That combined with my other comment should end up saving it to a compressed file.

python/lib/sift_py/data_import/hdf5.py
```diff
@@ -0,0 +1,281 @@
...
+    csv_dict["description"] = data_cfg.description
+    csv_dict["enum_types"] = data_cfg.enum_types
+    csv_dict["bit_field_elements"] = data_cfg.bit_field_elements
+
+    return csv_dict
+
+
+def _split_hdf5_configs(hdf5_config: Hdf5Config) -> List[Hdf5Config]:
+    """
+    Split up hdf5_config into separate configs used to generate each CSV file
```
marcsiftstack, line 238: There's a problem with splitting these up. The way the API works is that if you define `run_name` in the CsvConfig it will create and ingest into a new run. If you define `run_id` it will ingest into the existing run. So if we split this into multiple configs and each one only defines `run_name` then the data will be uploaded into multiple different runs. We would need to create a run using the specified `run_name` then get the id for that run and use that in the `CsvConfig` to make sure everything get's uploaded to a single run.

marcsiftstack: It would be nice to not have to split up the configs tho. Have you tried playing around with the different [write_csv](https://docs.pola.rs/api/python/stable/reference/api/polars.DataFrame.write_csv.html) options to see if we can avoid the empty string issue?

python/lib/sift_py/data_import/_hdf5_test.py
```diff
@@ -0,0 +1,610 @@
...
+
+def test_parse_hdf5_data_cfg():
+    data_cfg = Hdf5DataCfg(
+        name="TestChannel",
+        time_dataset="/TestChannel",
+        value_dataset="/TestChannel",
+        data_type="CHANNEL_DATA_TYPE_DOUBLE",
+        units="m/s",
+        description="Test channel",
+        enum_types=[],
```
marcsiftstack, line 261: We should check enums and bit field types are parsed correctly as well.

python/lib/sift_py/data_import/_hdf5_test.py
```diff
@@ -0,0 +1,610 @@
...
+                    "time_dataset": "/Channel1_Time",
+                    "value_dataset": "/Channel1_Value",
+                    "data_type": "CHANNEL_DATA_TYPE_DOUBLE",
+                },
+            ],
+        }
+    )
+
+    data_dict = {
+        "/Channel1_Time": np.array([0, 1, 2], dtype=np.int64),
```
marcsiftstack, line 309: What happens if the time array and value array don't match in length?

marcsiftstack (changes requested):

python/lib/sift_py/data_import/_config.py
```diff
@@ -185,3 +209,136 @@ def validate_bit_fields(self) -> Self:
...
+    # Only valid if data_type is "CHANNEL_DATA_TYPE_BIT_FIELD"
+    bit_field_elements: List[BitFieldElement] = []
+
+    @field_validator("data_type", mode="before")
+    @classmethod
+    def convert_data_type(cls, raw: Union[str, ChannelDataType, Type]) -> str:
+        """
+        Converts the provided data_type value to a string.
+        """
+        if isinstance(raw, type):
```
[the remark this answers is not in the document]
nathan-sift: Great idea, I've created a `ConfigDataModel` that's inherited by `DataColumn` and `Hdf5DataCfg`.

python/lib/sift_py/data_import/hdf5.py
nathan-sift: pl.concat(data_frames, how="align") works in most cases, but since we can't define the column it choses to align to, it leaves us open to issues with corner cases, such as if we tried to concat two dataframes with the same data column name.

python/lib/sift_py/data_import/hdf5.py
```diff
@@ -0,0 +1,281 @@
...
+        # For each hdf5_split_config, convert to a csv file and upload
+        import_services = []
+        for config in split_configs:
+            with NamedTemporaryFile(mode="w", suffix=".csv") as temp_file:
+                csv_config = _convert_to_csv_file(
+                    path,
+                    temp_file,
+                    config,
+                )
+                import_services.append(
```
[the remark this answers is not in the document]
nathan-sift: Good idea, I've updated the code

python/lib/sift_py/data_import/hdf5.py
```diff
@@ -0,0 +1,281 @@
...
+            "format": hdf5_config._hdf5_config.time.format,
+            "column_number": 1,
+            "relative_start_time": hdf5_config._hdf5_config.time.relative_start_time,
+        },
+    }
+
+    data_columns = {}
+    for idx, data_cfg in enumerate(hdf5_config._hdf5_config.data):
+        col_num = idx + 2  # 1-indexed and col 1 is time col
+        data_columns[col_num] = _parse_hdf5_data_cfg(data_cfg)
```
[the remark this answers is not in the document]
nathan-sift: Agreed, changed. I removed the unit test on this function, since the behavior is covered in the config unit tests.

python/lib/sift_py/data_import/hdf5.py
nathan-sift: Done

python/lib/sift_py/data_import/hdf5.py
nathan-sift: Very good catch for the runs.

Agreed that splitting up the configs isn't a great approach. But as I mentioned in another comment, when CSV files are ingested internally, a CSV line like ",," is treated as an empty, not nil, string.

python/lib/sift_py/data_import/hdf5.py
nathan-sift: Done. But see other comment on compressed csv issues

python/lib/sift_py/data_import/hdf5.py
```diff
@@ -64,23 +68,73 @@ def upload(
...
-        # For each hdf5_split_config, convert to a csv file and upload
-        import_services = []
-        for config in split_configs:
-            with NamedTemporaryFile(mode="w", suffix=".csv") as temp_file:
+        # Ensures all temp files opened under stack.enter_context() will have __exit__ called as with a standard with statement
+        with ExitStack() as stack:
+            # First convert each csv file
+            csv_items: List[Tuple[str, CsvConfig]] = []
+            for config in split_configs:
+                temp_file = stack.enter_context(NamedTemporaryFile(mode="wt", suffix=".csv"))
```
marcsiftstack, line 76: `"t"` is the default so it's not needed here. We use `"wt"` in some other places when writing to a compressed file because text mode is not the default with `gzip.open`.

python/lib/sift_py/data_import/hdf5.py
```diff
@@ -64,23 +68,73 @@ def upload(
         # Split up hdf5_config into separate configs. String data is split into separate configs. All other data is a single config
         split_configs = _split_hdf5_configs(hdf5_config)
 
-        # For each hdf5_split_config, convert to a csv file and upload
-        import_services = []
-        for config in split_configs:
-            with NamedTemporaryFile(mode="w", suffix=".csv") as temp_file:
+        # Ensures all temp files opened under stack.enter_context() will have __exit__ called as with a standard with statement
```
marcsiftstack, line 71: It's unclear to me why we need to use`ExitStack`. Is there a reason why this doesn't work with the standard `with` syntax?
```
csv_items: List[Tuple[str, CsvConfig]] = []
for config in split_configs:
    with NamedTemporaryFile(mode="wt", suffix=".csv") as temp_file:
        csv_config = _convert_to_csv_file(
            path,
            temp_file,
            config,
        )
        csv_items.append((temp_file.name, csv_config))

if hdf5_config._hdf5_config.run_name != "":
    ...

import_services = []
for filename, csv_config in csv_items:
    ...

return import_services
```

python/lib/sift_py/data_import/hdf5.py
```diff
@@ -64,23 +68,73 @@ def upload(
...
                 import_services.append(
                     self._csv_upload_service.upload(
-                        temp_file.name, csv_config, show_progress=show_progress
+                        filename, csv_config, show_progress=show_progress
                     )
                 )
 
         return import_services
 
+    def _create_run(self, run_name: str) -> str:
```
marcsiftstack, line 104: We can use the internal gRPC utility:https://github.com/sift-stack/sift/blob/0be65e513663cdc0d03a2a2d78c96a394b7c194a/python/lib/sift_py/ingestion/_internal/run.py#L32

I think it's fine if the user has to initialize this service with both the REST and gRPC configs.

marcsiftstack (changes requested):
.

python/lib/sift_py/data_import/hdf5.py
nathan-sift: With `NamedTemporaryFile`, once you exit the with block, the temp file isn't just closed, but also deleted, so you wouldn't be able to upload the CSV files with the approach above.

marcsiftstack: Ah ok. Might be more clear if the comment says that. Something like "Use ExitStack so that temporary files are not removed until all files have been processed and uploaded".

python/lib/sift_py/data_import/hdf5.py
marcsiftstack: Per our offline discussion, shouldn't include the gRPC config.

python/lib/sift_py/data_import/hdf5.py
nathan-sift: https://github.com/sift-stack/sift/pull/261/commits/febb50eadc48517d5d5e0be44f0d48d9d053a5fe

python/lib/sift_py/data_import/hdf5.py
```diff
@@ -171,85 +185,149 @@ def _convert_hdf5_to_dataframes(
     Returns:
         A polars DataFrame containing the data.
     """
+    # Group data configs by matching time arrays to optimize downstream data processing
+    data_cfg_ts_map: Dict[Tuple[str, int], List[Hdf5DataCfg]] = {}
+    for data_cfg in hdf5_config._hdf5_config.data:
+        map_tuple = (data_cfg.time_dataset, data_cfg.time_column)
+        if map_tuple not in data_cfg_ts_map:
```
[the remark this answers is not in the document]
nathan-sift: https://github.com/sift-stack/sift/pull/261/commits/bf685f89f256a4b068ab3302a3e7ca64995e40bd

python/lib/sift_py/data_import/hdf5.py
```diff
@@ -171,85 +185,149 @@ def _convert_hdf5_to_dataframes(
...
-    if value_series.dtype == pl.Binary:
-        value_series = value_series.cast(pl.String)
 
-    return pl.DataFrame(data={"timestamp": time_series, hdf5_data_config.name: value_series})
+    data_frame = pl.DataFrame(data={"timestamp": time_series})
+
+    for hdf5_data_config in hdf5_data_configs:
+        if not hdf5_data_config.value_dataset in hdf5_file:
+            raise Exception(f"HDF5 file does not contain dataset {hdf5_data_config.value_dataset}")
+        if time_path != hdf5_data_config.time_dataset:
```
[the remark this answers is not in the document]
nathan-sift: Good call, changed to assert

https://github.com/sift-stack/sift/pull/261/commits/bf685f89f256a4b068ab3302a3e7ca64995e40bd

python/lib/sift_py/data_import/hdf5.py
```diff
@@ -267,8 +345,17 @@ def _create_csv_config(hdf5_config: Hdf5Config) -> CsvConfig:
         },
     }
 
+    # Map each data config to its channel name
+    config_map = {d_cfg.name: d_cfg for d_cfg in hdf5_config._hdf5_config.data}
+
+    if merged_df.columns[0] != "timestamp":
+        raise Exception(
```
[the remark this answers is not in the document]
nathan-sift: As above

https://github.com/sift-stack/sift/pull/261/commits/bf685f89f256a4b068ab3302a3e7ca64995e40bd

python/lib/sift_py/data_import/hdf5.py
```diff
@@ -171,85 +185,149 @@ def _convert_hdf5_to_dataframes(
...
+                merged = _merge_ts_dataframes(df1, df2)
+                next_round.append(merged)
+            else:
+                next_round.append(data_frames[i])
+        data_frames = next_round
+    merged_df = data_frames[0].sort("timestamp")
+    return merged_df
+
+
+def _merge_ts_dataframes(df1: pl.DataFrame, df2: pl.DataFrame) -> pl.DataFrame:
```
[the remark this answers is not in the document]
nathan-sift: Timeseries. I wanted to make the function less generic than just "merge_dataframes". Changed name to make it clearer

https://github.com/sift-stack/sift/pull/261/commits/bf685f89f256a4b068ab3302a3e7ca64995e40bd

python/lib/sift_py/data_import/status.py
```diff
@@ -65,34 +65,48 @@ class DataImportService(_RestService):
...
-    def wait_until_complete(self) -> DataImport:
+    def get_data_imports(self) -> Generator[DataImport, None, None]:
+        for idx in range(len(self._data_import_ids)):
+            yield self.get_data_import(idx=idx)
+
+    def wait_until_complete(self, idx: int = 0) -> DataImport:
         """
         Blocks until the data import is completed. Check the status to determine
         if the import was successful or not.
+        Waits on only the first data import if multiple provided through `add_data_import_id` and `idx` not passed
```
marc-sift, line 105: Is `add_data_import_id` a typo? Should that be `extend`?

python/lib/sift_py/data_import/hdf5.py
```diff
@@ -48,7 +48,7 @@ def upload(
         path: Union[str, Path],
         hdf5_config: Hdf5Config,
         show_progress: bool = True,
-    ) -> List[DataImportService]:
+    ) -> Optional[DataImportService]:
```
marc-sift, line 51: Why is this Optional?

python/lib/sift_py/data_import/status.py
marc-sift, line 1: Do we need to add any unit tests to _status_tests.py?

marc-sift (changes requested):

python/lib/sift_py/data_import/hdf5.py
nathan-sift: I went back and forth on this one. We could originally return a blank list, which isn't possible to do with DataImportService, so we would need to return None instead. That could happen if the user provides a config/file that ultimately has no data in it to be processed.

Alternatively, we could raise an Exception if we didn't actually upload any files. If we keep it as-is, I'll add better documentation as to what None means

python/lib/sift_py/data_import/status.py
nathan-sift: Good call, added tests
https://github.com/sift-stack/sift/pull/261/commits/abcfe198cd20cd17e531ae72a9831cd78cdbda36

python/lib/sift_py/data_import/hdf5.py
nathan-sift: Made documentation more clear
https://github.com/sift-stack/sift/pull/261/commits/abcfe198cd20cd17e531ae72a9831cd78cdbda36

nathan-sift: Done
https://github.com/sift-stack/sift/pull/261/commits/74d3189addfe481c544f43d9980d84a2212cb53f
Mypy corrections
https://github.com/sift-stack/sift/pull/261/commits/f1b90b1c88e8722d561894e2c26dc32a5cccb723

PR #268, opened 2025-07-04
python(feat): Add support for start and end times to rule evaluation

marc-sift:
This change adds support for specifying start and end times with evalutating rules against a run.


niliayu (dismissed):
lgtm!

PR #269, opened 2025-07-07
FD-45: Add calculated channels to sift client.

ian-sift:
Calculated channels work only from this branch: https://github.com/sift-stack/sift/tree/experimental/new-client-root


marcsiftstack (changes requested):

python/lib/sift_client/_internal/low_level_wrappers/calculated_channels.py
```diff
@@ -0,0 +1,383 @@
...
+            client_key: The client key of the calculated channel.
+            organization_id: The organization ID (required if using client_key and user belongs to multiple organizations).
+
+        Returns:
+            The CalculatedChannel.
+
+        Raises:
+            ValueError: If neither calculated_channel_id nor client_key is provided.
+        """
+        if calculated_channel_id:
```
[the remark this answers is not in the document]
ian-sift: if by the self, *, <kwargs> paradigm yes. we should follow that. will update that here. I think a lot of these i've done on my main branch and just need to rebase

python/lib/sift_client/resources/calculated_channels.py
```diff
@@ -0,0 +1,363 @@
...
+        if not calculated_channel_id and not client_key:
+            raise ValueError("Either calculated_channel_id or client_key must be provided")
+
+        calculated_channel = await self._low_level_client.get_calculated_channel(
+            calculated_channel_id=calculated_channel_id,
+            client_key=client_key,
+            organization_id=organization_id,
+        )
+
+        return self._apply_client_to_instance(calculated_channel)
```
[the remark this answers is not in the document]
ian-sift: ya i think this would probably be useful in most cases

python/lib/sift_client/types/_base.py
```diff
@@ -66,39 +89,95 @@ def to_proto_with_mask(self) -> tuple[ProtoT, field_mask_pb2.FieldMask]:
...
+                    value = temp
+                sub_paths = self._build_proto_and_paths(
+                    proto_msg, value, "", already_setting_path_override=True
+                )
+                if mapping_helper.update_field:
+                    paths.append(mapping_helper.update_field)
+            elif isinstance(value, dict):
+                if field_name in self._to_proto_helpers:
+                    if not self._to_proto_helpers[field_name].converter:
+                        raise ValueError(
```
marcsiftstack, line 129: This wouldn't happen from user error right? Should this be an assert instead?

python/lib/sift_client/types/_base.py
```diff
@@ -66,39 +89,95 @@ def to_proto_with_mask(self) -> tuple[ProtoT, field_mask_pb2.FieldMask]:
...
         for field_name, value in data.items():
             path = f"{prefix}.{field_name}" if prefix else field_name
 
-            if isinstance(value, dict):
-                # Get the submessage
-                sub_msg = getattr(proto_msg, field_name)
-                # Recursively process nested fields
-                sub_paths = self._build_proto_and_paths(sub_msg, value, path)
-                paths.extend(sub_paths)
+            if not already_setting_path_override and field_name in self._to_proto_helpers:
```
marcsiftstack, line 114: I'm not able to completely follow the logic in this function. I think it'll be helpful to talk through in person.

python/lib/sift_client/types/metadata.py
```diff
@@ -2,27 +2,81 @@
...
+)
 
 if TYPE_CHECKING:
-    from sift_client.client import SiftClient
+    pass
+
+
+def metadata_dict_to_proto(_metadata: dict[str, str | float | bool]) -> list[MetadataProto]:
+    """
+    Wraps metadata dictionary into a list of MetadataValue objects.
```
marcsiftstack, line 19: Would it be more accurate to say "Converts" instead of "Wraps"?

python/lib/sift_client/types/asset.py
```diff
@@ -17,21 +17,21 @@ class Asset(BaseType[AssetProto, "Asset"]):
     Model of the Sift Asset.
     """
 
-    asset_id: str
+    id: str
```
marcsiftstack, line 20: We should keep this as `asset_id` since that's how it's named in the API: https://docs.siftstack.com/docs/api/reference/protocol-buffers/assets#asset

python/lib/sift_client/types/asset.py
```diff
@@ -17,21 +17,21 @@ class Asset(BaseType[AssetProto, "Asset"]):
...
     name: str
     organization_id: str
     created_date: datetime
     created_by_user_id: str
     modified_date: datetime
     modified_by_user_id: str
     tags: list[str]
-    metadata: list[MetadataValue]
+    metadata: dict[str, str | float | bool]
```
marcsiftstack, line 28: Why are the Metadata changes needed?

python/lib/sift_client/types/metadata.py
marcsiftstack, line 1: Why are the Metadata changes needed? Seems like it's implemented in an inconsistent way by not using the BaseType.

python/lib/sift_py/_internal/time.py
```diff
@@ -26,7 +26,7 @@ def to_timestamp_nanos(arg: Union[TimestampPb, pd.Timestamp, datetime, str, int]
         return cast(pd.Timestamp, pd.Timestamp(dt))
 
     else:
-        return cast(pd.Timestamp, pd.Timestamp(arg))
+        return cast(pd.Timestamp, pd.Timestamp(arg, tz=timezone.utc))
```
marcsiftstack, line 29: Was this causing issues somewhere?

python/lib/sift_py/calculated_channels/service.py
```diff
@@ -316,6 +316,10 @@ def update_calculated_channel(
         req = UpdateCalculatedChannelRequest(
             calculated_channel=channel_updater, update_mask=update_mask, user_notes=update_notes
         )
+        print("req", req)
```
marcsiftstack, line 319: I think this was included on accident.

python/lib/sift_client/types/_base.py
```diff
@@ -50,7 +50,30 @@ class ModelUpdate(BaseModel, Generic[ProtoT], ABC):
 
     model_config = ConfigDict(frozen=False)
 
+    class MappingHelper(BaseModel):
```
marcsiftstack, line 53: Doesn't seem like we gain anything by making this a nested class. Should we pull it out?

python/lib/sift_client/types/calculated_channel.py
```diff
@@ -0,0 +1,174 @@
...
+if TYPE_CHECKING:
+    from sift_client.client import SiftClient
+
+
+class CalculatedChannel(BaseType[CalculatedChannelProto, "CalculatedChannel"]):
+    """
+    Model of the Sift Calculated Channel.
+    """
+
+    id: str
```
marcsiftstack, line 28: Can should use `calculated_channel_id` since that's what's used in the API: https://docs.siftstack.com/docs/api/reference/protocol-buffers/calculated_channels#calculatedchannel

python/lib/sift_client/types/calculated_channel.py
```diff
@@ -0,0 +1,174 @@
...
+    name: str
+    description: str
+    expression: str
+    channel_references: list[ChannelReference]
+
+    units: str | None
+    asset_ids: list[str] | None
+    tag_ids: list[str] | None
+    all_assets: bool | None
+    calculated_channel_id: str | None
```
marcsiftstack, line 38: Why is there both `calculated_channel_id` and `id`?

python/lib/sift_client/types/asset.py
```diff
@@ -17,21 +17,21 @@ class Asset(BaseType[AssetProto, "Asset"]):
...
     archived_date: datetime | None
 
     @property
     def is_archived(self):
         """Whether the asset is archived."""
-        # TODO: clean up this logic when gRPC returns a null
+        # TODO: clean up this logic when gRPC returns a null. Note this applies to calculated channels as well.
```
marcsiftstack, line 34: We should move this logic into a common function that all types can use.

python/lib/sift_client/types/calculated_channel.py
```diff
@@ -0,0 +1,174 @@
...
+
+    def update(
+        self,
+        update: CalculatedChannelUpdate | dict,
+        user_notes: str | None = None,
+    ) -> CalculatedChannel:
+        """
+        Update the Calculated Channel.
+
+        Args:
```
marcsiftstack, line 77: The args don't match the function.

python/lib/sift_client/types/_base.py
ian-sift: ya sounds good. This is ultimately to make this paradigm Luck set up work for field's whose nested path does not match the structure of the proto. For instance `expression_channel_references` is a to level attribute you can set for a `CalculatedChannelUpdate` so you don't have to mirror the nested structs to set `calculated_channel_configuration.query_configuration.sel.expression_channel_references`. 

I would not consider this funciton or really this PR in general production ready and would expect to ensure every thing works when we set up examples once all the core functionality is in. I've already had to make further updates to this function for `Run` and would be happy to explore different paradigms than the ModelUpdate one or different ways to make it work as a generalized solution

python/lib/sift_client/types/asset.py
ian-sift: If the purpose of this new client library is to abstract away from the clunkiness of the API, I would prefer to have all the types just use "id" so users can type `asset.id` and not `asset.asset_id` or even worse: `calculated_channel.calculated_channel_id`

python/lib/sift_client/types/_base.py
ian-sift: This is extremely tied (and only relevant) to ModelUpdate. I don't feel very strongly about it moving it one level higher to make includes easier though

python/lib/sift_client/types/asset.py
ian-sift: agree. I don't wanna do that on this PR though.

python/lib/sift_client/types/_base.py
marcsiftstack: Ah ok, that makes sense.

python/lib/sift_client/_internal/low_level_wrappers/calculated_channels.py
```diff
@@ -0,0 +1,383 @@
...
+        response = cast(ListCalculatedChannelVersionsResponse, response)
+
+        versions = [
+            CalculatedChannel._from_proto(cc) for cc in response.calculated_channel_versions
+        ]
+        return versions, response.next_page_token
+
+    async def resolve_calculated_channel(
+        self,
+        identifier: Any | None = None,  # ResourceIdentifier from protobuf
```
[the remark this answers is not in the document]
marcsiftstack: Looks like this was removed? Is that intentional?

python/lib/sift_client/types/_base.py
marcsiftstack: The only reason to use an inner class is if we want to hide that implementation detail or take advatange of the class scope, but we are using MappingHelper directly outside of the class. We also have code like this `MappingHelper = ModelUpdate.MappingHelper` so I don't think it makes sense to be an inner class.

python/lib/sift_client/types/asset.py
marcsiftstack: Ok that's fine.

python/lib/sift_client/types/asset.py
```diff
@@ -42,16 +42,6 @@ def created_by(self):
     def modified_by(self):
         raise NotImplementedError
 
-    @property
-    def runs(self):
-        # TODO: how to implement these efficiently without fetching all?
```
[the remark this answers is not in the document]
ian-sift: just removing changes i made on the runs pr

python/lib/sift_client/types/_base.py
ian-sift: moved up f9ada653018fb2e76940cbfd73738b1062df1a11

python/lib/sift_client/_internal/low_level_wrappers/calculated_channels.py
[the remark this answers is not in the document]
ian-sift: Ya I didn't actually understand th user facing use case here so removed. If we want to track adding it, i can spin off a ticket

python/lib/sift_client/resources/calculated_channels.py
```diff
@@ -0,0 +1,363 @@
...
+        order_by: str | None = None,
+        limit: int | None = None,
+    ) -> list[CalculatedChannel]:
+        """
+        List versions of a calculated channel.
+
+        Args:
+            calculated_channel_id: The ID of the calculated channel.
+            client_key: The client key of the calculated channel.
+            organization_id: The organization ID.
```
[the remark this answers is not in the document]
ian-sift: https://github.com/sift-stack/sift/pull/269/files#diff-5fa4ef24458d83409f482571f74d6019e1702b9677c5370294ed32ed7336403bR65

marcsiftstack (approved):

PR #270, opened 2025-07-08
Creating dev branch to base calc channels, rules, runs, and channels off

ian-sift:
of.


python/lib/sift_client/_internal/low_level_wrappers/ingestion.py
```diff
@@ -0,0 +1,405 @@
...
+            )
+        req = IngestWithConfigDataStreamRequestPy(
+            ingestion_config_id=flow.ingestion_config_id,
+            run_id=flow.run_id,
+            flow=flow.name,
+            timestamp=to_rust_py_timestamp(timestamp),
+            channel_values=rust_channel_values,
+            end_stream_on_validation_error=False,
+            organization_id=organization_id
+            or "",  # This will be filled in by the server # TODO: Confirm w/ Ailin
```
niliayu, line 359: Yes, we should be able to leave this unspecified -- the org ID can be resolved via the API key on the backend.

python/lib/sift_client/_internal/low_level_wrappers/ingestion.py
```diff
@@ -0,0 +1,405 @@
...
+    def __init__(self, grpc_client: GrpcClient):
+        """
+        Initialize the IngestionLowLevelClient.
+
+        Args:
+            grpc_client: The gRPC client to use for making API calls.
+        """
+        super().__init__(grpc_client=grpc_client)
+        # TODO: Better way?
+        # Rust GRPC client expects URI to have http(s):// prefix.
```
niliayu, line 138: For consideration: maybe urllib.parse?

PR #276, opened 2025-07-23
python(chore): Update channel_enum_type proto

nathan-sift:
Updates channel_enum_type proto to include optional is_signed flag


solidiquis:
Also tagging @adhi-goel for vis

adhi-goel (approved):

PR #277, opened 2025-07-23
fix(sift-stream-bindings): Pass bitfield values as vectors

niliayu:
Pass bitfield elements as vectors of bits, rather than the `ChannelBitFieldElementPy` which is meant for configuration info needed for the ingestion config. 
With this change, from python, bitfields may now be passed as e.g. `[0b000]`


rust/crates/sift_stream_bindings/src/stream/channel.rs
```diff
@@ -349,7 +349,7 @@ impl ChannelValueTypePy {
     #[staticmethod]
     pub fn bitfield(value: Vec<u8>) -> Self {
         Self {
-            inner: ChannelValueType::BitField(value),
+            inner: ChannelValueType::BitField(value.to_vec()),
```
ian-sift, line 352: is value not already a vec?

PR #296, opened 2025-08-18
python(feat): lazy flow creation

nathan-sift:
Implements lazy flow creation for the IngestionService.

Lazy flow creation registers the telemetry config at IngestionService initialization without any flow configs. Flow configs are registered only upon initial ingestion of that flow. This amortizes flow config registration over time, allowing very large telemetry configs which previously would have reached the maximum gRPC message size, and reducing the time needed for initilization of the IngestionService for larger configs.

By default, smaller telemetry configs will continue to be initialized up front. A `force_lazy_flow_ingestion` flag can be set to true at IngestionService initialization to force this behavior for smaller configs.

### Benchmark

**4 asset sizes**
- Small: 2 flows, 5 channels each
- Medium: 20 flows, 50 channels each
- Large: 200 flows, 1000 channels each
- Very large: 1000 flows, 2500 channels each

**Tests**
- 1 data point: Init IngestionService and ingest up to 50 flows worth of channels once
- 100 data points: Init IngestionService and ingest up to 50 flows worth of channels 10 times
- Performed on new assets, than repeated to provide existing asset stats
- Benchmarked both initialization time and ingestion time
- Tested with the following:
  - Baseline library 0.8.3
  - Lazy flow default behavior (`force_lazy_flow_ingestion` = False)
  - All lazy flow (`force_lazy_flow_ingestion` = True)

image

**Conclusions**
- No noticeable impact on smaller flows with default behavior (as expected)
- The slow large asset initialization issues are mitigated with lazy flow ingestion
  - 60s init reduces to <1s. First flow ingestion time increases, due to amortization of flow creation, but overall time is still faster. Also true for ingestion of existing assets
- Lack of significant ingest time increase with 10 data point test vs 1 point test indicates that ingestion remains fast after initial flows are sent.
- Lazy flow enables use of very large assets without need to split up flows beforehand. With the existing library, very large assets throw a gRPC error due to exceeding the gRPC message size

### DIFF
--- a/python/lib/sift_py/ingestion/_internal/ingest.py
+++ b/python/lib/sift_py/ingestion/_internal/ingest.py
@@ -65,38 +65,37 @@ class _IngestionServiceImpl:
         end_stream_on_error: bool = False,
         force_lazy_flow_creation: bool = False,
     ):
-        ingestion_config = None
-        if not force_lazy_flow_creation:
-            try:
-                ingestion_config = self.__class__._get_or_create_ingestion_config(
-                    channel, config, lazy_flows=False
-                )
-            except ProtobufMaxSizeExceededError:
-                pass
-            else:
-                self.use_lazy_flow_creation = False
-
-        if not ingestion_config:
+        try:
+            ingestion_config = self.__class__._get_or_create_ingestion_config(
+                channel, config, lazy_flows=force_lazy_flow_creation
+            )
+        except ProtobufMaxSizeExceededError:
             ingestion_config = self.__class__._get_or_create_ingestion_config(
                 channel, config, lazy_flows=True
             )
             self.use_lazy_flow_creation = True
+        else:
+            self.use_lazy_flow_creation = force_lazy_flow_creation
 
         self.ingestion_config = ingestion_config
 
+        # `flow_configs_by_name` will include all flows in the config, and anything already created
+        # `flow_configs_created` only includes those which are already registered
         if config._ingestion_client_key_is_generated and not self.use_lazy_flow_creation:
             # If this is a generated key, use the local telemetry config since it is static
-            # All flows have also already been created
+            # All flows have also already been created since we aren't lazy
             self.flow_configs_by_name = {flow.name: flow for flow in config.flows}
             self.flow_configs_created = {flow.name for flow in config.flows}
         else:
-            # If the user specified a client key, use the configuration from Sift since it
-            # may have been updated.
             # If using lazy flow creation, assume the list of flows is large enough that
             # `get_ingestion_config_flows` will be very slow. Instead lazily check for
             # flow existance during streaming
+            # Otherwise, since the user specified a client key, use the configuration from Sift since it
+            # may have been updated, and we've already registered any new flows
+
             if self.use_lazy_flow_creation:
-                flows = []
+                self.flow_configs_by_name = {flow.name: flow for flow in config.flows}
+                self.flow_configs_created: Set[str] = set()
             else:
                 flows = [
                     FlowConfig.from_pb(f)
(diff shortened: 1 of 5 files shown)


python/lib/sift_py/ingestion/_internal/ingest.py
```diff
@@ -58,21 +63,53 @@ def __init__(
         config: TelemetryConfig,
         run_id: Optional[str] = None,
         end_stream_on_error: bool = False,
+        force_lazy_flow_creation: bool = False,
     ):
-        ingestion_config = self.__class__._get_or_create_ingestion_config(channel, config)
+        ingestion_config = None
```
marc-sift, line 68: What do you think about this? Mostly trying to avoid the `not` conditions and also remove branches:
```
try:
    ingestion_config = self.__class__._get_or_create_ingestion_config(
        channel, config, lazy_flows=force_last_flow_creation
    )
except ProtobufMaxSizeExceededError:
    ingestion_config = self.__class__._get_or_create_ingestion_config(
        channel, config, lazy_flows=True
    )

self.use_lazy_flow_creation = ingestion_config.lazy_flow_creation
```

python/lib/sift_py/ingestion/_internal/ingest.py
```diff
@@ -58,21 +63,53 @@ def __init__(
...
+                    FlowConfig.from_pb(f)
+                    for f in get_ingestion_config_flows(
+                        channel,
+                        ingestion_config.ingestion_config_id,
+                    )
+                ]
+            # `flow_configs_by_name` will include all flows in the config, and anything already created
+            # `flow_configs_created` only includes those which are already registered
+            self.flow_configs_by_name = {flow.name: flow for flow in config.flows}
+            self.flow_configs_by_name.update({flow.name: flow for flow in flows})
```
marc-sift, line 111: `config.flows` should be the entire set of flows right? Do we still need this line?

python/lib/sift_py/ingestion/_internal/ingest.py
```diff
@@ -399,13 +449,61 @@ def create_flow(self, *flow_config: FlowConfig):
...
+            ):
+                missing_flow_config_names.add(request.flow)
+
+        flow_configs_to_create = [
+            self.flow_configs_by_name[flow_config_name]
+            for flow_config_name in missing_flow_config_names
+        ]
+
+        if flow_configs_to_create:
+            # print(f"Lazily create flows: {[item.name for item in flow_configs_to_create]}")
```
marc-sift, line 480: Forgot to remove this.

marc-sift (changes requested):
Looking good! Should also add unit tests if possible.

python/lib/sift_py/ingestion/_internal/ingest.py
nathan-sift: get_ingestion_config_flows() may also include flows that aren't supplied by the TelemetryConfig. To avoid changing that behavior for non-lazy ingestion, we still need to populate from get_ingestion_config_flows(). But I did rework this a bit to avoid overwriting flow_configs_by_name, so hopefully the behavior is more clear.

python/lib/sift_py/ingestion/_internal/ingest.py
nathan-sift: I like that
https://github.com/sift-stack/sift/pull/296/commits/b34dbe40ab19c70c71f1be238b978b1dc7792f4a

python/lib/sift_py/ingestion/_internal/ingest.py
nathan-sift: https://github.com/sift-stack/sift/pull/296/commits/b34dbe40ab19c70c71f1be238b978b1dc7792f4a

PR #305, opened 2025-08-31
Python (fix): explicit extras for build to exclude docs, dev

alexluck-sift:
Update offline build process to no longer include dev and docs extras.

### DIFF
--- a/.github/workflows/python_build.yaml
+++ b/.github/workflows/python_build.yaml
@@ -124,7 +124,13 @@ jobs:
         shell: bash
         run: |
           # Generate requirements file with all extras
-          pip-compile pyproject.toml --all-extras -o requirements-all.txt
+          pip-compile pyproject.toml \
+          --extra tdms \
+          --extra hdf5 \
+          --extra sift-stream \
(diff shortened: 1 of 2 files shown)


.github/workflows/python_build.yaml
```diff
@@ -124,7 +124,14 @@ jobs:
         shell: bash
         run: |
           # Generate requirements file with all extras
```
ian-sift, line 126: If not passing --all-extras should maybe change comment.

python/pyproject.toml
```diff
@@ -62,6 +62,8 @@ tdms = ["npTDMS~=1.9"]
 rosbags = ["rosbags~=0.0"]
 sift-stream = ["sift-stream-bindings>=0.1.2"]
 hdf5 = ["h5py~=3.11", "polars~=1.8"]
+# Ensure any new user build extras are added to .github/workflows/python_build.yaml
```
ian-sift, line 65: It would be nice to have a comment explaining what to put in development   vs make a new target for

PR #311, opened 2025-09-10
rust(feat): update protocol buffers

solidiquis:
## Changes

Updates protobuf generated code for `sift_rs`.


nathan-sift (approved):
Looks good, other than looking like it needs a pass from cargo fmt

PR #317, opened 2025-09-15
python(feat): Upload TDMS metadata to Sift Runs

marc-sift:
Add flag to ingest TDMS file properties as Sift Run metadata.


nathan-sift (approved):
Looks good

PR #333, opened 2025-10-01
rust(feat): Remove deprecated recovery strategies

tsift:
Removes previously deprecated recovery strategies `RecoveryStrategy::RetryWithInMemoryBackups` and `RecoveryStrategy::RetryWithDiskBackups`.

With the removal of these strategies, the `SiftStream:: backups_manager` member has also been simplified to the single backup manager type (though still optional based on recovery strategy).


nathan-sift (commented):
We should also update the docs in src/lib.rs

PR #334, opened 2025-10-01
rust(feat): SiftStream Metrics

nathan-sift:
Adds more comprehensive metrics to SiftStream and the ability for users to access those metrics.

- Metrics are all atomics, making updating them non-blocking, and are designed to have a negligible impact on performance
- Accessing metrics by the user is non-blocking for SiftStream
- Metrics are currently gated behind the `metrics-unstable` feature flag. This is primarily driven by the need to update metrics in the near future in likely API breaking ways
- Users can access metrics directly from a SiftStream instance through the get_metrics_snapshot function
- After enabling the metrics HTML server, users can also access metrics process-wide through '/' and '/metrics' endpoints. The server responds with a JSON representation of the `StreamStreamMetricsSnapshot` for each SiftStream instance, organized by the SiftStream Ids

**Metric Server Example Output**

`sift_stream::metrics::start_metrics_server(9100).await;`

```
> curl -s http://localhost:9100 | jq

{
  "27d777a3-56a9-4be8-823e-4412dddd3c06": {
    "elapsed_secs": 174.684,
    "loaded_flows": 2,
    "unique_flows_received": 2,
    "messages_received": 185,
    "messages_sent": 185,
    "message_rate": 1.0590552082617755,
    "bytes_sent": 23495,
    "byte_rate": 134.5000114492455,
    "messages_sent_to_backup": 187,
    "cur_retry_count": 0,
    "checkpoint": {
      "checkpoint_count": 3,
      "failed_checkpoint_count": 0,
      "checkpoint_timer_reached_cnt": 2,
      "checkpoint_manually_reached_cnt": 0,
      "cur_elapsed_secs": 54.669,
      "cur_messages_sent": 59,
      "cur_message_rate": 1.079222228319523,
      "cur_bytes_sent": 7585,
      "cur_byte_rate": 138.74407799667088
    },
    "backups": {
      "cur_checkpoint_file_count": 1,
      "cur_checkpoint_cur_file_size": 7757,
      "cur_checkpoint_bytes": 7757,
      "cur_checkpoint_messages": 59,
      "total_file_count": 3,
      "total_bytes": 24371,
      "total_messages": 187,
      "files_pending_ingestion": 0,
      "files_ingested": 0,
      "cur_ingest_retries": 0
    }
  }
}
```

### DIFF
--- a/rust/crates/sift_stream/src/metrics/mod.rs
+++ b/rust/crates/sift_stream/src/metrics/mod.rs
@@ -186,7 +186,7 @@ impl AtomicInstant {
     pub fn new(init: Instant) -> AtomicInstant {
         AtomicInstant {
             init_instant: init,
-            offset_ns: AtomicU64::new(0)
+            offset_ns: AtomicU64::new(0),
         }
     }
 
@@ -356,8 +356,7 @@ impl SiftStreamMetrics {
         let messages_sent_to_backup = self.messages_sent_to_backup.get();
         let cur_retry_count = self.cur_retry_count.get();
 
-        let stats =
-            StreamingStats::calculate(self.creation_time, messages_sent, bytes_sent);
+        let stats = StreamingStats::calculate(self.creation_time, messages_sent, bytes_sent);
 
         SiftStreamMetricsSnapshot {
             elapsed_secs: stats.elapsed_secs,
@@ -376,10 +375,7 @@ impl SiftStreamMetrics {
     }
 
     pub(crate) fn get_checkpoint_stats(&self) -> StreamingStats {
-        let start_time = self
-            .checkpoint
-            .checkpoint_start_time
-            .get();
+        let start_time = self.checkpoint.checkpoint_start_time.get();
         let messages_sent = self.checkpoint.cur_messages_sent.0.load(Ordering::Relaxed);
         let bytes_sent = self.checkpoint.cur_bytes_sent.0.load(Ordering::Relaxed);
 
--- a/rust/crates/sift_stream/src/metrics/mod.rs
+++ b/rust/crates/sift_stream/src/metrics/mod.rs
@@ -14,7 +14,7 @@ use serde::Serialize;
 
 #[cfg(feature = "metrics-unstable")]
 /// **Unstable Feature:** API may change at any time
-/// 
+///
 /// Snapshot of checkpoint-related metrics.
 ///
 /// Tracks checkpoint operations, success/failure rates, and current checkpoint performance.
@@ -35,7 +35,7 @@ pub struct CheckpointMetricsSnapshot {
 
 #[cfg(feature = "metrics-unstable")]
 /// **Unstable Feature:** API may change at any time
-/// 
+///
 /// Snapshot of backup-related metrics.
 ///
 /// Tracks file counts, byte totals, and ingestion status if backups are enabled.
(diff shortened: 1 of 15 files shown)


rust/crates/sift_stream/src/metrics/server.rs
```diff
@@ -0,0 +1,77 @@
...
+/// # Arguments
+/// * `port` - The port number to bind the HTTP server to
+pub async fn start_metrics_server(port: u16) {
+    use hyper::{Method, Request, Response};
+    use hyper::{server::conn::http1, service::service_fn};
+    use hyper_util::rt::TokioIo;
+    use std::{convert::Infallible, net::SocketAddr};
+    use tokio::net::TcpListener;
+
+    let addr = SocketAddr::from(([0, 0, 0, 0], port));
```
tsift, line 34: It might be nice to provide the address as an arg as well, for example if a user doesn't want to expose the http server to all network interfaces, or just have it be localhost.

rust/crates/sift_stream/src/stream/mode/ingestion_config.rs
```diff
@@ -162,7 +164,18 @@ impl SiftStream<IngestionConfigMode> {
         let begin_checkpoint_notifier = Arc::new(Notify::new());
         let sift_stream_id = Uuid::new_v4();
 
-        let data_stream = DataStream::new(sift_stream_id, data_rx, drain_tx);
+        // Spawn a task to register metrics without blocking
```
tsift, line 167: What is the rationale for not blocking here and using another task to register the metrics?

tsift (approved):

rust/crates/sift_stream/src/metrics/server.rs
nathan-sift: Added builder https://github.com/sift-stack/sift/pull/334/commits/76d4198c5814713ed676fc119f4292161cba0ecf

rust/crates/sift_stream/src/stream/mode/ingestion_config.rs
nathan-sift: SiftStream::new() isn't async, so we can't just await register_metrics, and making it async breaks the API. There are other ways we could work around this, but this makes it completely non-blocking to SiftStream. I did realize we can feature flag this, so I'll add that

nathan-sift: https://github.com/sift-stack/sift/pull/334/commits/4b6ce8fbb362ab0d8698264df9bd67a06ca9cb2e

rust/crates/sift_stream/src/metrics/mod.rs
```diff
@@ -0,0 +1,373 @@
...
+                    Duration::default()
+                })
+                .as_millis() as u64,
+            ..Default::default()
+        }
+    }
+
+    #[cfg(feature = "metrics-unstable")]
+    pub(crate) fn snapshot(&self) -> SiftStreamMetricsSnapshot {
+        let loaded_flows = self.loaded_flows.get();
```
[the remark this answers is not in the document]
nathan-sift: All of these are atomics as well. I've just added that wrapper over them so we don't need to call out the atomic ordering every time we get or set the value. This will all be abstracted down to just atomic reads/writes by the compiler.

As you mentioned, the alternative is a RWLock over a struct of normal ints, which would require us to lock the mutex every time we update a metric, which is undesirable for a number of reasons.

PR #346, opened 2025-10-17
Python(feat): disk caching

alexluck-sift:
Prototype of a gRPC disk cache


python/lib/sift_py/grpc/cache.py
```diff
@@ -0,0 +1,202 @@
...
+        metadata_dict: dict[str, str]
+        if not metadata:
+            metadata_dict = {}
+        else:
+            # Handle both tuple and grpc.aio.Metadata types
+            metadata_dict = {}
+            for key, value in metadata:
+                metadata_dict[key] = value
+
+        use_cache = metadata_dict.get(METADATA_USE_CACHE, "false").lower() == "true"
```
ian-sift: could optionally `bool(metdata_dict.get(METADATA_USE_CACHE))`

python/lib/sift_client/client.py
```diff
@@ -116,18 +117,22 @@ def __init__(
             api_key: The Sift API key for authentication.
             grpc_url: The Sift gRPC API URL.
             rest_url: The Sift REST API URL.
-            connection_config: A SiftConnectionConfig object to configure the connection behavior of the SiftClient.
+            connection_config: A SiftConnectionConfig object to configure the connection and cache behavior of the SiftClient.
```
ian-sift: would be helpful to add something here like the other examples you added in low_level_wrappers/base.py or caching.py since i'd imagine one of the first pages users would look at is https://sift-stack.github.io/sift/python/pr-346/reference/sift_client/client/

ian-sift (commented):
not finished w/ full pass, so just left some optional nits. 

Would be curious about instrumenting metrics on how often a cache was hit as part of evaluating this change

PR #450, opened 2025-12-30
rust(feat): Add test server to CLI

marc-sift:


rust/crates/sift_cli/build.rs
```diff
@@ -0,0 +1,21 @@
+/// Build descriptor's so that the Black Hole gRPC server can
+/// stand up the reflection service.
+fn main() -> Result<(), Box<dyn std::error::Error>> {
```
solidiquis: There shouldn't be a need for you to compile these protos since they're already compiled and available in the `sift_rs` crate which you can add as a dependency like here: https://github.com/sift-stack/sift/blob/737efa8536dd912a484039d93bf839981b50d9be/rust/crates/sift_cli/Cargo.toml#L28

rust/crates/sift_cli/src/cmd/test_server/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+use tonic::transport::Server;
+use tonic_reflection::server::Builder;
+
+pub const FILE_DESCRIPTOR_SET: &[u8] = include_bytes!("../../../descriptor.bin");
+pub mod metrics_streaming_client;
+pub mod server;
+use crate::cmd::test_server::metrics_streaming_client::Metrics;
+
+pub async fn run(ctx: Context, args: TestServerArgs) -> Result<ExitCode> {
+    let local_address = args.local_address.unwrap_or("127.0.0.1:50051".into());
```
solidiquis: I would prefer `0.0.0.0` since it's more user-friendly. Allows it to be reached inside of a docker network for example. Also as a note of good practice:

```rust
local_address = args.local_address.unwrap_or_else(|| "127.0.0.1:50051".to_string())
```

for lazy evaluation of the fallback rather than immediate

rust/crates/sift_cli/src/cmd/test_server/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+    let addr = local_address.parse()?;
```
solidiquis: Generally helpful to include contextual information with errors for better traceability:

```rust
use anyhow::{Result, Context};

let addr = local_address.parse().context("foobar")?;
```

solidiquis: Same for all other places.

rust/crates/sift_cli/src/cmd/test_server/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+
+    // Initialize streaming client.
+    let mut streaming_client =
+        MetricsStreamingClient::build(ctx, &args.stream_metrics, &args.metrics_asset_name)?;
+    if streaming_client.is_some() {
```
solidiquis: Prefer:

```rust
if let Some(client) = streaming_client.as_mut() {
    client.initialize.await.context("failed to initialize client").await?;
}
```

rust/crates/sift_cli/src/cmd/test_server/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+                &mut shutdown_rx,
+                metrics_tx,
+                args.stream_metrics.unwrap_or(false),
+            )
+            .await;
+    });
+
+    // Start task to ingest metrics to Sift.
+    let ingest_metrics_task = tokio::spawn(async move {
+        if streaming_client.is_none() {
```
solidiquis: Prefer:
```rust
let Some(mut client) = streaming_client else {
    return;
};
```

rust/crates/sift_cli/src/cmd/test_server/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+            return;
+        }
+
+        let mut client = streaming_client.unwrap();
```
solidiquis: see:
```rust
let Some(mut client) = streaming_client else {
    return;
};
```

rust/crates/sift_cli/src/cmd/test_server/server.rs
```diff
@@ -0,0 +1,306 @@
...
+use tokio::sync::mpsc::Sender;
+use tokio::sync::watch;
+use tokio_stream::StreamExt;
+use tonic::{Request, Response, Status, Streaming};
+use uuid::Uuid;
+
+#[derive(Default)]
+pub struct TestServer {
+    /// Total number of streams created.
+    total_num_streams: AtomicU32,
```
solidiquis: Do these need to be atomics?

solidiquis (changes requested):
Definitely run: `cargo fmt` and `cargo clippy`. The latter will give you a lot of good feedback.

rust/crates/sift_cli/src/cmd/test_server/server.rs
```diff
@@ -0,0 +1,316 @@
...
+                    let current_total_num_messages = self.total_num_messages.load(Relaxed);
+                    let current_total_num_streams = self.total_num_streams.load(Relaxed);
+                    let bytes_per_s = current_total_num_bytes_read - last_total_num_bytes_read;
+                    let messages_per_s = current_total_num_messages - last_total_num_messages;
+
+                    last_total_num_bytes_read = current_total_num_bytes_read;
+                    last_total_num_messages = current_total_num_messages;
+
+                    // Clear terminal and print metrics.
+                    stdout
```
tsift: I think the server will often be ran separately in a background thread, or as a daemon; I think that that might make recording the logs to a file look weird? I've heard similar feedback for the CLI generally when running it from a script and recording logs to a file.

rust/crates/sift_cli/src/cmd/test_server/server.rs
```diff
@@ -0,0 +1,316 @@
...
+    ) -> Result<Response<GetIngestionConfigResponse>, Status> {
+        let inner = request.into_inner();
+        let ingestion_configs = self.ingestion_configs_by_id.lock().unwrap();
+        let ingestion_config = ingestion_configs
+            .get(&inner.ingestion_config_id)
+            .ok_or(Status::not_found("ingestion config not found"))?;
+
+        Ok(Response::new(GetIngestionConfigResponse {
+            ingestion_config: Some(IngestionConfig {
+                ingestion_config_id: ingestion_config.ingestion_config_id.clone(),
```
tsift: nit: I think you can clone the entire ingestion_config directly? I might be wrong though.

rust/crates/sift_cli/src/cmd/test_server/mod.rs
marc-sift: 834ba0f1131e5a9f7c8e4545443a721dd820fc02

rust/crates/sift_cli/src/cmd/test_server/server.rs
```diff
@@ -0,0 +1,316 @@
...
+        Ok(Response::new(
+            sift_rs::assets::v1::DeleteAssetResponse::default(),
+        ))
+    }
+
+    /// No-op.
+    async fn list_assets(
+        &self,
+        _request: Request<sift_rs::assets::v1::ListAssetsRequest>,
+    ) -> Result<Response<sift_rs::assets::v1::ListAssetsResponse>, Status> {
```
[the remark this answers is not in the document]
marc-sift: The python client library depends on some of these. I will make the unused called return unimplemented, but some like `list_assets` are required.

rust/crates/sift_cli/build.rs
marc-sift: I updated the build to generate these, we're pulling it from sift_rs now: 086172455c3a37734c4561aeae0225718fefeec2

rust/crates/sift_cli/src/cmd/test_server/server.rs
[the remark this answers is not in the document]
marc-sift: 086172455c3a37734c4561aeae0225718fefeec2

rust/crates/sift_cli/src/cmd/test_server/server.rs
```diff
@@ -0,0 +1,316 @@
...
+
+        Ok(Response::new(new_ingestion_config))
+    }
+
+    /// No-op.
+    async fn create_ingestion_config_flows(
+        &self,
+        _request: Request<CreateIngestionConfigFlowsRequest>,
+    ) -> Result<Response<CreateIngestionConfigFlowsResponse>, Status> {
+        Ok(Response::new(CreateIngestionConfigFlowsResponse::default()))
```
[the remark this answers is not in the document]
marc-sift: I created a ticket to track this as a future improvement.

tsift (approved):

PR #533, opened 2026-04-10
Create a subprocess for uploading the report as it goes

ian-sift:


python/lib/sift_client/util/test_results/context_manager.py
```diff
@@ -100,7 +101,28 @@ def __init__(
...
+                    ["replay-test-result-log", "--incremental", str(self.log_file)],
+                    stdin=subprocess.PIPE,
+                    stdout=subprocess.DEVNULL,
+                    stderr=subprocess.DEVNULL,
+                )
+            finally:
+                os.dup2(saved_stderr, 2)
+                os.close(saved_stderr)
+        else:
+            self._replay_proc = None
```
alexluck-sift, line 122: define this at the class level to make it a bit easier to follow (right now it is just defined/set in this function)

python/lib/sift_client/_internal/low_level_wrappers/test_results.py
```diff
@@ -864,124 +954,167 @@ async def replay_log_file(
         if not log_path.exists():
             raise FileNotFoundError(f"Log file not found: {log_file}")
 
-        simulated_report: TestReport | None = None
-        simulated_steps_by_id: dict[str, TestStep] = {}
-        simulated_steps_order: list[str] = []
-        simulated_measurements_by_id: dict[str, TestMeasurement] = {}
-        simulated_measurements_order: list[str] = []
+        if incremental:
```
alexluck-sift, line 957: Perhaps a bit more error handling?

Any thoughts on how to handle partial writes/killed processes and replaying from those?

jsonl is a good call since at least each line is independent (vs. I've seen json logs/dbs get corrupted very easily) so should be fairly isolated

ian-sift: Ya i think this only happens if the process is killed. But lines are small and have to be parseable into the request. 

replay halts loudly on any unparseable line — this is asserted by test_malformed_log_line_skipped in python/lib/sift_client/_tests/resources/test_test_results.py:713. The ReportContext.__exit__ path catches that, logs replay instructions, and surfaces the failure.

The user gets pointed at import-test-result-log <file> to retry manually, but the bad line stops automated replay until someone trims it.

PR #534, opened 2026-04-10
rust(feat): Add tracing scoped dispatch to capture sift-stream logs 

tsift:
Adds tracing event capturing via "scoped dispatching" so that `SiftStream` internally can collect it's own tracing events and include them in metrics forwarded to Sift for debugging.


nathan-sift (approved):
lgtm!

PR #535, opened 2026-04-13
update protos

wei-qlu:
Protos commit SHA: 0d9e66dbe02e93069d2f1e9a185e699a04d47e07


python/lib/sift/py.typed
alexluck-sift, line 1: this shouldn't be removed, might need to tweak our process to avoid deletion

wei-qlu: it's being deleted by `make gen`, I'll look into it

wei-qlu: `make gen` deletes and recreates `python/lib/sift`, which removes `py.typed`. Since it's not generated by buf, the script now adds it afterwards.

alexluck-sift (approved):

PR #540, opened 2026-04-16
python(chore): sift-stack-py v0.14.0 prep

wei-qlu:
Bumps sift-stack-py to v0.14.0. Data import API now fully available in sift_client (CSV/Parquet/TDMS/HDF5), Parquet added as an export output format, plus test result logging and progress indicators. See the CHANGELOG for the full list of changes.

### DIFF
--- a/python/CHANGELOG.md
+++ b/python/CHANGELOG.md
@@ -3,7 +3,7 @@ All notable changes to this project will be documented in this file.
 
 This project adheres to [Semantic Versioning](http://semver.org/).
 
-## [v0.14.0] - April 17, 2026
+## [v0.14.0] - April 20, 2026
 
 ### What's New
 
--- a/python/CHANGELOG.md
+++ b/python/CHANGELOG.md
@@ -3,69 +3,20 @@ All notable changes to this project will be documented in this file.
 
 This project adheres to [Semantic Versioning](http://semver.org/).
 
-## [v1.0.0] - April 16, 2026
-
-This is the first major release of `sift-stack-py`. `sift_client` has reached feature parity with `sift_py` and is now the recommended interface for all new development. `sift_py` (deprecated since [v0.10.0](#v0100---january-30-2026)) continues to work and ship in this release; all features previously available only in `sift_py`, including data imports and exports, are now available in `sift_client`.
-
-### Migrating from `sift_py` to `sift_client`
-
-`sift_client` introduces a unified, Pythonic interface and is not a drop-in replacement for `sift_py`. The major differences are summarized below. For complete examples, see the [sift_client documentation](https://sift-stack.github.io/sift/python/latest/#sift-client-api-new).
-
-#### Unified client vs. per-service classes
-
-`sift_py` exposed a separate service class per domain (`IngestionService`, `RuleService`, `AssetService`, `ReportTemplateService`, etc.). `sift_client` exposes a single `SiftClient` with resource accessors:
-
-```python
-# sift_py
-from sift_py.grpc.transport import SiftChannelConfig, use_sift_channel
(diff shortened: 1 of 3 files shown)


python/CHANGELOG.md
```diff
@@ -3,8 +3,92 @@ All notable changes to this project will be documented in this file.
...
 
+## [v0.14.0] - April 22, 2026
 
-https://github.com/sift-stack/sift/commit/e1305363e95cafb4c980fdd493f69e72660a52fb
+### What's New
+
+#### Data Import API in SiftClient
+The `sift_client` module now exposes a data import API supporting CSV, Parquet, TDMS, and HDF5. With this addition, all features previously available only in `sift_py` are now available in `sift_client`, which remains the recommended interface for new development. `sift_py` (deprecated since [v0.10.0](#v0100---january-30-2026)) continues to work and ship in this release.
+
+Migrating from `sift_py`: the per-format upload services (`CsvUploadService`, `ParquetUploadService`, `Hdf5UploadService`, `TdmsUploadService`) are replaced by a single `client.data_import.import_from_path` method. Where `sift_py` required hand-building a config object for every import, `sift_client` auto-detects it from the file extension. Use the new public `client.data_import.detect_config(...)` first if you need to inspect or patch the result before importing. `import_from_path` returns a `Job` you can optionally wait on.
```
marc-sift: sift_py contained `simple_upload` variants that did not require hand building a config. We should include some information telling users they don't need to use a dedicated function anything and can instead pass in an empty config to rely on auto detection.

python/CHANGELOG.md
```diff
@@ -3,8 +3,92 @@ All notable changes to this project will be documented in this file.
...
+
+client = SiftClient(api_key=apikey, grpc_url=grpc_url, rest_url=rest_url)
+job = client.data_import.import_from_path("data.csv", asset="my_asset")
+job.wait_until_complete()
+```
+
+Format-by-format support:
+- **CSV**: auto-detected from `.csv`. Supports an optional JSON metadata row (row 1 or row 2) for specifying channel names, units, data types, and the time column format.
+- **Parquet**: requires an explicit `data_type` (`PARQUET_FLATDATASET` or `PARQUET_SINGLE_CHANNEL_PER_ROW`) since `.parquet` alone doesn't disambiguate the layout. Detection only reads the file footer, so it stays fast on large files.
+- **HDF5**: new in this release. Auto-detected from `.h5` / `.hdf5`. Detection runs fully client-side (no server round-trip) and walks each dataset.
```
marc-sift: I think it's less important to mention that it all happens client side, and more important to mention that the auto detection works if your HDF5 format is in a specific layout. Otherwise, you will need to create your own config.

python/CHANGELOG.md
```diff
@@ -3,8 +3,92 @@ All notable changes to this project will be documented in this file.
...
+- **TDMS**: new in this release. Auto-detected from `.tdms`. Detection is also fully client-side and maps TDMS groups and channels onto Sift channels.
```
marc-sift: I think it's less important to mention that it all happens client side, and more important to mention the different import options for TDMS.

wei-qlu: 9a2806b79aff23b86f4ad91c00ed541b5fc41938

python/CHANGELOG.md
wei-qlu: 9a2806b79aff23b86f4ad91c00ed541b5fc41938

python/CHANGELOG.md
wei-qlu: 9a2806b79aff23b86f4ad91c00ed541b5fc41938

solidiquis (approved):

PR #555, opened 2026-05-07
rust(feat): Rename ingestion_config for python linting

tsift:
Renames `ingestion_config` to `ingestion_config_form` to avoid a name conflict in python linting causing a false finding.


nathan-sift (approved):
lgtm

PR #556, opened 2026-05-11
rust(fix): allow partial sift-cli config updates

wei-qlu:
## Summary
`sift-cli config update` with a subset of `--grpc-uri` / `--rest-uri` / `--api-key` individually printed "Nothing to update." and exited. `is_update_empty` joined the three "missing flag" checks with `||` instead of `&&`, firing the early return whenever any flag was absent. Flipped the operator and added unit tests for the empty / partial / full cases.

## Validation 

Unit and manual tests

image

image


solidiquis (approved):

solidiquis (changes requested):
move test to separate file

solidiquis (dismissed):

solidiquis (approved):

PR #560, opened 2026-05-13
Python(fix): fix log not being passed with update calls

alexluck-sift:
Previously, log file generated TestResult objects were not updating with the log file references. This meant calls to methods like `update` were passing through to the API instead of the log file resulting in errors (since the entity did not yet exist on the server).


python/pyproject.toml
```diff
@@ -171,7 +171,7 @@ file-imports = [
 ]
 hdf5 = [
     'h5py~=3.11',
-    'polars~=1.8',  # only used by sift_py; remove once sift_py is fully deprecated
```
ian-sift, line 174: is this no longer valid?

python/lib/sift_client/_tests/resources/test_test_results_log_file.py
```diff
@@ -0,0 +1,357 @@
...
+        api._low_level_client.import_log_file = AsyncMock(return_value=replay_result)
+
+        result = await api.import_log_file(log_path)
+
+        assert result.report._log_file is None
+        assert all(s._log_file is None for s in result.steps)
+        assert all(m._log_file is None for m in result.measurements)
+
+
+class TestEndToEndLogFileRouting:
```
ian-sift: could just this end to end test be added to the existing test resutls test?

ian-sift (dismissed):

python/pyproject.toml
alexluck-sift: it is, but this comment is removed by the auto generation of the extras block. The fact that this comment exists cause the pre-push hook to fail.

ian-sift (approved):

PR #565, opened 2026-05-15
rust(feat): adding hdf5 imports

Brandon-Shippy:
## Description

:ticket: ENG-10457 :ticket: 

Adding hdf5 import support for the following formats:
- 1D
- 2D
- Compound

Require the user to specify the format of the file and the timestamp. 

Accounts for if user potentially put in the improper schema, then auto detects what format they should put.

Note: In the current build the Jobs page will show HDF5 imports as "Unknown Import", so in https://github.com/sift-stack/azimuth/pull/11523 I fixed this as well for the next release.

## Verification

Manual testing of importing hdf5 files and unit tests

1d preview
Screenshot 2026-05-15 at 1 19 58 PM
2d preview
Screenshot 2026-05-15 at 1 20 34 PM
compound preview
Screenshot 2026-05-15 at 1 21 23 PM
hdf5 help preview
Screenshot 2026-05-15 at 1 21 51 PM
If user improperly puts in the wrong schema into the cli args
Screenshot 2026-05-15 at 1 32 49 PM


rust/crates/sift_cli/src/cli/mod.rs
```diff
@@ -420,3 +425,29 @@ pub struct ImportTdmsArgs {
...
+    /// Start time (RFC3339) if the time format is relative
+    #[arg(short = 's', long)]
+    pub relative_start_time: Option<String>,
+
+    /// (two-d / compound) Index of the time column or field. Defaults to 0.
+    #[arg(long, default_value_t = 0)]
+    pub time_index: u64,
+
+    /// (compound) Name of the time field. Overrides --time-index when set.
+    #[arg(long)]
```
tsift, line 455: Does `clap` have an option to require this be mutually exclusive with time-index? Not sure thats the behavior we would want, but would make the "override" behavior more explicit.

rust/crates/sift_cli/src/cmd/import/hdf5/detect_hdf5_schema.rs
```diff
@@ -0,0 +1,301 @@
...
+        TypeDescriptor::Unsigned(IntSize::U8) => Some(ChannelDataType::Uint64),
+        TypeDescriptor::Float(FloatSize::U4) => Some(ChannelDataType::Float),
+        TypeDescriptor::Float(FloatSize::U8) => Some(ChannelDataType::Double),
+        TypeDescriptor::FixedAscii(_)
+        | TypeDescriptor::FixedUnicode(_)
+        | TypeDescriptor::VarLenAscii
+        | TypeDescriptor::VarLenUnicode => Some(ChannelDataType::String),
+        TypeDescriptor::Enum(_) => Some(ChannelDataType::Int32),
+        TypeDescriptor::Compound(_)
+        | TypeDescriptor::FixedArray(_, _)
```
tsift: Do we need to support these ones?

rust/crates/sift_cli/src/cmd/import/hdf5/detect_hdf5_schema.rs
```diff
@@ -0,0 +1,301 @@
...
+    } else {
+        anyhow!(
+            "no datasets matched --schema {selected_label}. \
+             Did you mean --schema {}?",
+            suggestions.join(" or --schema ")
+        )
+    }
+}
+
+fn same_schema(a: &Hdf5Schema, b: &Hdf5Schema) -> bool {
```
tsift: I think this could be accomplished with `#[derive(PartialEq, Eq)` on `Hdf5Schema`, then just a `a == b`.

rust/crates/sift_cli/src/cmd/import/hdf5/detect_hdf5_schema.rs
```diff
@@ -0,0 +1,301 @@
...
+                continue;
+            }
+            let Some(channel_type) = hdf5_to_sift_data_type(&field.ty) else {
+                continue;
+            };
+            let channel_name = format!("{}.{}", name.trim_start_matches('/'), field.name);
+            let channel_config = ChannelConfig {
+                name: channel_name,
+                data_type: channel_type as i32,
+                ..Default::default()
```
tsift, line 302: Does hdf5 have any other metadata we could include? Like units or descriptions?

rust/crates/sift_cli/src/cmd/import/hdf5/detect_hdf5_schema.rs
```diff
@@ -0,0 +1,301 @@
...
+    match result {
+        Ok((data, _)) if data.is_empty() => {
+            Err(no_match_error(&datasets, schema, time_index, time_field))
+        }
+        Ok(other) => Ok(other),
+        Err(e) => Err(e),
+    }
+}
+
+fn no_match_error(
```
tsift, line 79: nit: Might be beneficial to include more code comments, particularly for functions/structs/types, for additional context on the intent of the code so it can help future contributors get up to speed, and with all the usage of AI these days, can also help agents review and catch issues/find improvements.

tsift (changes requested):

rust/crates/sift_cli/src/cmd/import/hdf5/detect_hdf5_schema.rs
Brandon-Shippy: Yes it is there and supported, but only for the 1D schema from how it is currently setup. In a 1D schema each dataset has its own attributes. In a 2D schema it has a set of attributes for the whole thing, its not separated. In a Compound schema its a little weird, but it seems like it also doesnt store metadata per field; it follows a similar style as 2D with having one set of attributes. 

So parsing it out may not be useful beyond just 1D

Brandon-Shippy: Adding additional 1D metadata, because of the reasons above

rust/crates/sift_cli/src/cli/mod.rs
Brandon-Shippy: https://github.com/sift-stack/sift/pull/565/commits/6e6e8c9c5ae1c1430b0f5654bbc10c4b6599a24e

rust/crates/sift_cli/src/cmd/import/hdf5/detect_hdf5_schema.rs
Brandon-Shippy: https://github.com/sift-stack/sift/pull/565/commits/81ce51fecb9990d7259da852441e65fda37d8bea

rust/crates/sift_cli/src/cmd/import/hdf5/detect_hdf5_schema.rs
Brandon-Shippy: https://github.com/sift-stack/sift/pull/565/commits/b10044398c29b465d3a58fa5faaa641c4b890900

as we discussed, removed the ones that we cannot guarantee 100% would translate into a sift type. Prefer to throw error to user and make the supported types clear via -help flag

tsift (approved):

PR #567, opened 2026-05-15
Python(feat): pytest plugin improvements

alexluck-sift:
## Summary
- Promote the plugin to `sift_client.pytest_plugin` (was buried at `util/test_results/pytest_util`) so users wire it in with a single `pytest_plugins = ["sift_client.pytest_plugin"]`.
- Add ini keys mirroring every CLI flag and a default `sift_client` fixture that reads `SIFT_*` env vars with `sift_grpc_uri` / `sift_rest_uri` ini fallback, so per-project config can live in `pyproject.toml`.
- Reorganize plugin tests into `_tests/pytest_plugin/` so each follow-up PR drops a file rather than appending to a monolith.
- Add `sift_include` / `sift_exclude` markers and a `sift_test_results_autouse` ini key so projects scope which tests produce Sift reports. Closest marker wins, defaults to the ini value. Bulk-apply per directory with `pytest_collection_modifyitems` in a conftest.
- Gate the autouse fixtures per test, with `report_context` created lazily. Mixed unit/integration suites can share one `pytest` invocation without unit tests paying the backend round-trip.


python/lib/sift_client/pytest_plugin.py
```diff
@@ -0,0 +1,310 @@
...
+def pytest_addoption(parser: pytest.Parser) -> None:
+    """Register Sift-specific command-line options and ini keys.
+
+    Each option can be set on the command line or under ``[tool.pytest.ini_options]``
+    in ``pyproject.toml`` (or ``[pytest]`` in ``pytest.ini``). CLI values take
+    precedence over ini values, which take precedence over the built-in default.
+    """
+    group = parser.getgroup("sift", description="Sift test results")
+    group.addoption(
+        "--sift-test-results-log-file",
```
ian-sift: can we set these w/ the variables so there's less surface are to potentially create regressions/number of places to update? f"--{_INI_LOG_FILE.replace('_','-')}"?

python/lib/sift_client/pytest_plugin.py
```diff
@@ -0,0 +1,310 @@
...
+    return ini_value if isinstance(ini_value, str) and ini_value else None
+
+
+@pytest.fixture(scope="session")
+def sift_client(pytestconfig: pytest.Config) -> SiftClient:
+    """Default ``SiftClient`` resolved from environment variables and ini keys.
+
+    Each credential is read from its environment variable first. The URIs
+    (``SIFT_GRPC_URI``, ``SIFT_REST_URI``) additionally fall back to the
+    ``sift_grpc_uri`` / ``sift_rest_uri`` ini keys, since they are stable
```
ian-sift, line 302: is this the correct order? Should we prefer env over ini?

ian-sift (changes requested):

python/lib/sift_client/pytest_plugin.py
alexluck-sift: I was thinking that the ini config would be more explicit and less confusing (since env is not very transparent in a lot of cases). It may be more confusing if you see a config with it set one way and not sure why it wasn't used. No strong opinion though.

python/lib/sift_client/pytest_plugin.py
alexluck-sift: I took another approach so everything is still greppable by adding a new _Option pattern

ian-sift (approved):

ian-sift (approved):

PR #569, opened 2026-05-15
Python(feat): pytest graceful handle missing connection

alexluck-sift:
## Why

Target handled missing connection via `--sift-test-results-check-connection`, which yielded `None` from `step` / `report_context` / `module_substep` on ping failure. Any test calling `step.measure(...)` then crashed with `AttributeError`, forcing `if step is None: pytest.skip(...)` guards throughout user code. The flag name also didn't convey what it actually did.

## Changes

- Adds `--sift-offline`: skips the ping, routes create/update calls to a JSONL log file for replay via `import-test-result-log`, tolerates missing `SIFT_*` env vars.
- Adds `--sift-disabled` (also `SIFT_DISABLED=1`): nothing contacts Sift, no log file is written. `step.measure(...)` still evaluates bounds and returns a real pass/fail boolean. No credentials required.
- Exposes `is_simulated` on `TestReport` / `TestStep` / `TestMeasurement` / `ReportContext` so callers can branch on provenance (true for disabled and offline modes).
- Replaces the opaque check-connection toggle with two named modes that say what they do — `offline` (write to log) vs `disabled` (skip entirely).
- Online mode now fail-fasts: a failed ping aborts the session with `pytest.UsageError` naming both flags as escape hatches.
- Renames CLI/ini args for brevity (`sift_test_results_*` → `sift_*`).

### DIFF
--- a/python/lib/sift_client/_tests/conftest.py
+++ b/python/lib/sift_client/_tests/conftest.py
@@ -79,5 +79,17 @@ def ci_pytest_tag(sift_client):
 
 
 def pytest_configure(config: pytest.Config) -> None:
-    """Enable the Sift connection-check mode for the fixtures used in this test suite since we run w/ mock client in non-integration tests."""
-    config.option.sift_test_results_check_connection = True
+    """Pick a Sift plugin mode based on whether integration tests are running.
+
+    Integration runs targeting a real backend stay online with the log file
+    disabled (writes go inline). Other runs default to ``--sift-disabled`` so
+    unit tests don't need credentials or a log file.
+    """
+    is_integration_run = "integration" in (config.option.markexpr or "")
+    have_real_backend = all(
+        os.getenv(name) for name in ("SIFT_API_KEY", "SIFT_GRPC_URI", "SIFT_REST_URI")
+    )
+    if is_integration_run and have_real_backend:
+        config.option.sift_test_results_log_file = False
+    else:
+        config.option.sift_disabled = True
--- a/python/lib/sift_client/_tests/conftest.py
+++ b/python/lib/sift_client/_tests/conftest.py
@@ -89,6 +89,6 @@ def pytest_configure(config: pytest.Config) -> None:
     """
     is_integration_run = "integration" in (config.option.markexpr or "")
     if is_integration_run:
-        config.option.sift_test_results_log_file = False
+        config.option.sift_log_file = False
     else:
         config.option.sift_disabled = True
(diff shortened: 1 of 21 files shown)


python/lib/sift_client/_tests/conftest.py
```diff
@@ -79,5 +79,16 @@ def ci_pytest_tag(sift_client):
...
+
+    Integration runs (``-m integration``) stay online with the log file
+    disabled so writes go inline against the real backend; missing
+    ``SIFT_*`` env vars surface as an actionable usage error from the
+    plugin's ``sift_client`` fixture. Every other run defaults to
+    ``--sift-disabled`` so unit tests don't need credentials.
+    """
+    is_integration_run = "integration" in (config.option.markexpr or "")
+    if is_integration_run:
+        config.option.sift_log_file = False
```
alexluck-sift: use log file instead

python/lib/sift_client/pytest_plugin.py
```diff
@@ -248,37 +283,60 @@ def _report_context_impl(
...
         test_case = base_name
-    log_file = _resolve_log_file(pytestconfig)
+    # When the client is in simulate mode (disabled), bypass the log-file
+    # pipeline — pinning a path would create a file the simulate path never
+    # writes to — and skip the import subprocess (no real RPCs to replay).
+    simulate = sift_client._simulate
+    log_file: str | Path | bool | None = False if simulate else _resolve_log_file(pytestconfig)
     git_metadata = _option_or_ini(pytestconfig, _GIT_METADATA)
     include_git_metadata = True if git_metadata is None else bool(git_metadata)
+    offline = False if simulate else _is_offline(pytestconfig)
```
ian-sift: Took me a sec to confirm that offline still simulates since we are now deriving a `simulate` from this plugin where we then define `offline` as `!simulate` but then ultimately both paths still call `simulate_<request>`. Idk if keeping the terminology as `disabled` and `offline` would be clearer. 

Fine to keep as is though because I could still follow after re-reading. Can resolve this comment  or address it and ping me

ian-sift (approved):

python/lib/sift_client/pytest_plugin.py
alexluck-sift: agreed, renamed some of the args to make it explicit what is being done. Now we are explicitly setting log replay.

ian-sift (approved):

PR #580, opened 2026-05-20
python(fix): support nested calculated channel references

wei-qlu:
## Summary
  - Surfaces a `calculated_channel: CalculatedChannel | str` field on `ChannelReference` and routes it into the proto's `calculated_channel_version_id` oneof when serializing calculated channels (create + update), rules (create + update), and exports. A `CalculatedChannel` input is normalized to its `version_id` via a validator; a string is interpreted as a version_id directly.
  - Skips the name-based channel lookup in `resolve_calculated_channels` when a version reference is already provided, so nested references don't get treated as native channel lookups and fail.
  - Updates `Rule._from_proto` and `ChannelReference._from_proto` to read the version ID back off the proto oneof on the response path.

Use case:
```python
cc = client.calculated_channels.find(name="TEST_CC")

# Reference the latest version (cc IS the latest)
ChannelReference(channel_reference="$1", calculated_channel=cc)

# Pin to a specific version_id (e.g., copied from the UI)
ChannelReference(channel_reference="$1", calculated_channel="8feb6ccd-...")
```

## Validation
  - Unit tests covering:
    - `ChannelReference` validation and proto round-tripping (oneof read + write), including object input, string input, and rejection cases
    - Calculated channel serialization on both Create and Update paths
    - Rules serialization helper in both directions, plus `Rule._from_proto` round trip
    - `resolve_calculated_channels` skipping the name lookup when a version reference is provided
  - Manually created and updated a nested CC using the new `calculated_channel` field on `ChannelReference` and confirmed there were no missing dependents.
  - Manually created a rule whose expression references a CC by version_id and confirmed it evaluated correctly.

### DIFF
--- a/python/lib/sift_client/_internal/low_level_wrappers/rules.py
+++ b/python/lib/sift_client/_internal/low_level_wrappers/rules.py
@@ -80,6 +80,7 @@ def _channel_reference_to_proto(ref: ChannelReference) -> ChannelReferenceProto:
         return ChannelReferenceProto(
             calculated_channel_version_id=ref.calculated_channel_version_id
         )
+    assert ref.channel_identifier is not None
     return ChannelReferenceProto(name=ref.channel_identifier)
 
 
--- a/python/lib/sift_client/_internal/low_level_wrappers/rules.py
+++ b/python/lib/sift_client/_internal/low_level_wrappers/rules.py
@@ -76,10 +76,10 @@ logger = logging.getLogger(__name__)
 
 
 def _channel_reference_to_proto(ref: ChannelReference) -> ChannelReferenceProto:
-    if ref.calculated_channel_version_id:
-        return ChannelReferenceProto(
-            calculated_channel_version_id=ref.calculated_channel_version_id
-        )
+    # After ChannelReference validation, calculated_channel is always a version_id string.
+    if ref.calculated_channel:
+        assert isinstance(ref.calculated_channel, str)
+        return ChannelReferenceProto(calculated_channel_version_id=ref.calculated_channel)
     assert ref.channel_identifier is not None
     return ChannelReferenceProto(name=ref.channel_identifier)
 
--- a/python/lib/sift_client/_internal/low_level_wrappers/rules.py
+++ b/python/lib/sift_client/_internal/low_level_wrappers/rules.py
@@ -76,12 +76,13 @@ logger = logging.getLogger(__name__)
 
 
 def _channel_reference_to_proto(ref: ChannelReference) -> ChannelReferenceProto:
-    # After ChannelReference validation, calculated_channel is always a version_id string.
+    # ChannelReference's validator normalizes calculated_channel to a version_id str
+    # and guarantees exactly one of calculated_channel / channel_identifier is set.
     if ref.calculated_channel:
-        assert isinstance(ref.calculated_channel, str)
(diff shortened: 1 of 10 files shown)


python/lib/sift_client/sift_types/channel.py
```diff
@@ -344,13 +344,35 @@ def runs(self) -> list[Run]:
...
+
+    Exactly one of `channel_identifier` or `calculated_channel_version_id` must be set.
+    Use `calculated_channel_version_id` to reference another calculated channel by its
+    version ID (required for nested calculated channels, since names are not unique).
+    """
 
     channel_reference: str  # The key of the channel in the expression i.e. $1, $2, etc.
-    channel_identifier: str  # The name of the channel
+    channel_identifier: str | None = None  # The name (or ID) of an existing channel.
+    calculated_channel_version_id: str | None = None  # The version ID of a calculated channel.
```
alexluck-sift: This feels clunky and is inconsistent with our other surfaces. We should expose a first class CalculatedChannelVersion type to pass in here (and of course let a user pass the ID directly if they have it.

Think through how a user is going to realistically build this up:
1) Find the calculated channel they want
2) Get the version of it they want (usually would just be latest)
3) pass that in

wei-qlu: Updated to accept a CalculatedChannel (which already contains version_id, config, etc.) or a version_id string directly

python/lib/sift_client/_internal/low_level_wrappers/rules.py
```diff
@@ -75,6 +75,15 @@
 logger = logging.getLogger(__name__)
 
 
+def _channel_reference_to_proto(ref: ChannelReference) -> ChannelReferenceProto:
+    # After ChannelReference validation, calculated_channel is always a version_id string.
+    if ref.calculated_channel:
+        assert isinstance(ref.calculated_channel, str)
```
alexluck-sift: Avoid asserts outside of tests and instead raise specific errors. 
Since this appears to be just type narrowing, use type checking/casting instead

python/lib/sift_client/sift_types/channel.py
```diff
@@ -344,13 +345,48 @@ def runs(self) -> list[Run]:
...
+    Exactly one of `channel_identifier` or `calculated_channel` must be set.
+    To reference another calculated channel (nested CC), pass either a fetched
+    `CalculatedChannel` (its `version_id` is used) or a `version_id` string
+    directly. Both normalize to the `version_id` string after validation.
+    """
 
     channel_reference: str  # The key of the channel in the expression i.e. $1, $2, etc.
-    channel_identifier: str  # The name of the channel
+    channel_identifier: str | None = None  # The name (or ID) of an existing channel.
+    # A CalculatedChannel or its version_id. Normalized to the version_id string after validation.
```
alexluck-sift: this style of docstring doesn't show up in the docs

python/lib/sift_client/_internal/low_level_wrappers/rules.py
wei-qlu: updated to use cast

python/lib/sift_client/sift_types/channel.py
wei-qlu: resolved

alexluck-sift (approved):

PR #582, opened 2026-05-21
python(fix): infer parquet time column for int64/uint64 timestamps

wei-qlu:
## Summary
- This adds a client-side fallback that picks the first INT64 or UINT64 column whose name (case-insensitive) is `ts`, `timestamp`, or `time` and removes it from `data_columns`.
- Applies to both `PARQUET_FLATDATASET` and `PARQUET_SINGLE_CHANNEL_PER_ROW` layouts.

## Test plan
- Unit tests: `pytest python/lib/sift_client/_tests/resources/test_data_imports.py`
- Import a Parquet file with a bare `int64` `timestamp` column via `client.data_imports.import_from_path(...)` and confirm the returned config promotes it to `time_column` and excludes it from `data_columns`.
- Import a Parquet file whose timestamp uses Arrow `timestamp(ns)` and confirm server-side detection still wins (no regression).

### DIFF
--- a/python/lib/sift_client/resources/data_imports.py
+++ b/python/lib/sift_client/resources/data_imports.py
@@ -335,11 +335,12 @@ def _infer_time_column(
 ) -> str | None:
     """Pick a likely time column when the server couldn't identify one.
 
-    Returns the path of the first INT64 or UINT64 column whose name
+    Returns the path of an INT64 or UINT64 column whose name
     (case-insensitive) matches one of ``ts``, ``timestamp``, or ``time``.
     Returns None otherwise.
     """
-    for name, data_type, path in columns:
+    data_columns = sorted(columns, key=lambda c: c[0].lower())
+    for name, data_type, path in data_columns:
         if data_type in _TIME_COLUMN_TYPES and name.lower() in _TIME_COLUMN_NAMES:
             return path
     return None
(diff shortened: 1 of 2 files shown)


python/lib/sift_client/resources/data_imports.py
```diff
@@ -328,16 +353,30 @@ def _parse_parquet_detect_response(
...
                 dc for dc in parquet_config.data_columns if dc.path != time_path
             ]
         return parquet_config
     elif proto.HasField("single_channel_per_row"):
-        return ParquetSingleChannelPerRowImportConfig._from_proto(
+        scpr_config = ParquetSingleChannelPerRowImportConfig._from_proto(
             proto, footer_offset=footer_offset, footer_length=footer_length
         )
+        if not scpr_config.time_column.path:
+            inferred = _infer_time_column(
```
marc-sift: Nit: You use name `time_path` above, we should use the same var name here for consistency.

python/lib/sift_client/resources/data_imports.py
```diff
@@ -320,6 +324,27 @@ def _parse_csv_detect_response(proto) -> CsvImportConfig:
...
+def _infer_time_column(
+    columns: Iterable[tuple[str, ChannelDataType, str]],
+) -> str | None:
+    """Pick a likely time column when the server couldn't identify one.
+
+    Returns the path of the first INT64 or UINT64 column whose name
+    (case-insensitive) matches one of ``ts``, ``timestamp``, or ``time``.
+    Returns None otherwise.
+    """
+    for name, data_type, path in columns:
```
marc-sift: We should sort the columns alphabetically to get deterministic behavior when there are multiple channels that meet the criteria.

marc-sift (changes requested):

python/lib/sift_client/resources/data_imports.py
wei-qlu: 78ab9b14a09857450b37327de10f8d0a79af53d0 updated

python/lib/sift_client/resources/data_imports.py
wei-qlu: 78ab9b14a09857450b37327de10f8d0a79af53d0 updated

marc-sift (approved):

PR #585, opened 2026-05-25
python(refactor): update CSV import example

wei-qlu:
## Summary
Replace deprecated `sift_py` CSV import examples (`csv/simple` + `csv/custom`) with a single `sift_client` example using `import_from_path`.

## Validation
In a new venv, run `python main.py` from `python/examples/data_import/csv/ and confirm import completes in Sift.


Brandon-Shippy (approved):
lgtm

PR #591, opened 2026-05-26
python(refactor): update parquet import example

wei-qlu:
## Summary
Replace deprecated `sift_py` parquet import example with three `sift_client` examples: `flat_dataset`, `single_channel_per_row/single_channel`, `single_channel_per_row/multi_channel`. Each folder ships its own `main.py`, `sample_data.parquet`, `requirements.txt`, and `.env.example`, sharing the same underlying telemetry

Update `uv.lock` to match current `sift-stack-py` version 0.16.2

## Validation
- Ran `python main.py` in each of the three folders and verified the uploads completed and channels were visible in the Sift


Brandon-Shippy (approved):
lgtm, comments make sense-- good examples

PR #593, opened 2026-05-27
python(feat): add HDF5 schema detection

wei-qlu:
## Summary
- Adds `Hdf5Schema` enum (`ONE_D`, `TWO_D`, `COMPOUND`) and a required `hdf5_schema` parameter on `import_from_path` and `detect_config`.
- Implements per-schema detection in `_internal/util/hdf5.py`, datasets that don't fit the chosen schema are not included.

## Test plan
- Imported a test HDF5 file under each schema (`ONE_D`, `TWO_D`, `COMPOUND`); imports completed and channels matched expectations
- Confirmed `import_from_path("file.h5", ...)` without `hdf5_schema` raises a clear `ValueError`
- Unit tests cover each schema's happy path, filter behavior, and the shared primitives


python/lib/sift_client/resources/data_imports.py
```diff
@@ -189,6 +198,7 @@ async def detect_config(
         self,
         file_path: str | Path,
         data_type: DataTypeKey | None = None,
+        hdf5_schema: Hdf5Schema | None = None,
```
alexluck-sift: if we add support for other file types that require schema, this is going to be messy to expand. Perhaps simply a `schema` that takes Hdf5Schema now and possibly other types in the future?

python/lib/sift_client/resources/data_imports.py
```diff
@@ -48,6 +49,7 @@ async def import_from_path(
         asset: Asset | str | None = None,
         config: ImportConfig | None = None,
         data_type: DataTypeKey | None = None,
+        hdf5_schema: Hdf5Schema | None = None,
```
alexluck-sift: same feedback here - this seems overly catered towards hdf5 and not extensible to future situations where we need to use the same pattern

python/lib/sift_client/_internal/util/hdf5.py
```diff
@@ -1,96 +1,284 @@
...
+# Per-group time dataset names, case-insensitive, in priority order.
+_TIME_DATASET_NAMES = ("time", "timestamp", "timestamps", "ts")
 
-def _detect_attr(dataset: h5py.Dataset, candidates: list[str], default: str = "") -> str:
-    """Return the first matching HDF5 attribute value, or *default*."""
-    possible = [dataset.attrs.get(attr) for attr in candidates if dataset.attrs.get(attr)]
-    return str(possible[0]) if possible else default
 
+@dataclass(frozen=True)
+class _DetectedDataset:
```
marc-sift: This information is already contained in the [h5py.Dataset](https://docs.h5py.org/en/stable/high/dataset.html#h5py.Dataset.name). Do we need this wrapper?

python/lib/sift_client/_internal/util/hdf5.py
```diff
@@ -1,96 +1,284 @@
...
+def detect_hdf5_config(file_path: str | Path, schema: Hdf5Schema) -> Hdf5ImportConfig:
+    """Detect an HDF5 import config under the given ``schema``. Datasets that
+    don't fit the chosen schema are not included."""
+    path = Path(file_path)
+    with h5py.File(path, "r") as h5file:
+        columns = _BUILDERS[schema](_collect_datasets(h5file))
+
+    return Hdf5ImportConfig(
+        asset_name="",
+        time_format=TimeFormat.ABSOLUTE_UNIX_NANOSECONDS,
```
marc-sift: I think we should actually leave this as unspecified. If the user forgets to set this correctly, failing loudly is better than ingesting a potentially wrong timestamp.

python/lib/sift_client/sift_types/data_import.py
```diff
@@ -87,6 +87,29 @@ class DataTypeKey(Enum):
 }
 
 
+class Hdf5Schema(Enum):
+    """How an HDF5 file's datasets map to Sift channels.
+
+    Required when importing HDF5: the same file extension covers multiple
```
marc-sift: This is not required right? It can optionally be used to import HDF5 files if the users file fits into one of these formats.

marc-sift (changes requested):

python/lib/sift_client/_internal/util/hdf5.py
wei-qlu: it was redundant, removed 4c4a8c69287558ee5ed8dc17401888c5ff136ae5

python/lib/sift_client/_internal/util/hdf5.py
wei-qlu: single entry for time_format across different file tpyes 6199c3ec23da2be11b58635262a29c4eadd358bc

python/lib/sift_client/sift_types/data_import.py
wei-qlu: updated docs, it's only required for auto-detection, skipped if they provide their own config d8d8834cc0a884eb6cba7083e6116913654f879e

python/lib/sift_client/resources/data_imports.py
wei-qlu: updated to `schema`, new formats can be easily swapped in 1e42e6a6f8cc2e105594ab0203e375f3551cfb15

python/lib/sift_client/resources/data_imports.py
wei-qlu: 1e42e6a6f8cc2e105594ab0203e375f3551cfb15

python/lib/sift_client/resources/data_imports.py
```diff
@@ -48,6 +52,8 @@ async def import_from_path(
         asset: Asset | str | None = None,
         config: ImportConfig | None = None,
         data_type: DataTypeKey | None = None,
+        schema: Hdf5Schema | None = None,
```
marc-sift: Discussed offline - we can get rid of this additional `schema` argument and rely on updates to `DataTypeKey` alone. This will keep the similar interface as different Parquet schemas.

```
class DataTypeKey(Enum):
    """Supported file types for data import detection."""

    CSV = DATA_TYPE_KEY_CSV
    PARQUET_FLATDATASET = DATA_TYPE_KEY_PARQUET_FLATDATASET
    PARQUET_SINGLE_CHANNEL_PER_ROW = DATA_TYPE_KEY_PARQUET_SINGLE_CHANNEL_PER_ROW
    TDMS = DATA_TYPE_KEY_TDMS
    HDF5 = DATA_TYPE_KEY_HDF5
    HDF5_ONE_D = DATA_TYPE_KEY_HDF5_ONE_D
    HDF5_TWO_D = DATA_TYPE_KEY_HDF5_TWO_D
    HDF5_COMPOUND = DATA_TYPE_KEY_HDF5_COMPOUND
```

marc-sift (approved):

PR #594, opened 2026-05-28
Python(feat): pytest summary output

alexluck-sift:
- New end-of-run **Sift report** panel: report name in the section title, then `Status`, `Steps`, `Measurements` (color-coded: green pass, red fail/error, yellow skip), `Test case`, `System`, `Log file`, and a clickable `Report` link (online) or upload command (offline). Suppressed under `-q`.
- New session-start header showing the SDK version and active mode (e.g. `Sift: sift-stack-py 0.17.0, online mode`).
- New `--sift-open-report` / `sift_open_report` flag opens the report in a browser at session end (gated to interactive terminals so it does nothing in CI or captured shells).
- New `--sift-report-url-base` / `sift_report_url_base` / `SIFT_APP_URL` env var for on-prem and custom deployments. Online auto-derives the web-app URL for known Sift hosts; the override fills the gap for others. `SiftClient.app_url` exposes the resolved origin so non-plugin consumers can use it too.
- Richer `step.fail_if_measurements_failed()` output: lists each out-of-bounds measurement by name, value, and bounds (e.g. `[FAILED] b = 99.0 (min 0.0, max 2.0)`). Backed by `TestMeasurement.__str__`, so the same format is available in logs and the REPL.
- `ReportContext` now retains created steps and measurements with `step_status_counts` / `measurement_counts` properties. Docs and unit/integration tests updated for the new panel, options, and failure output.

## Verification

Sift info added to pytest header:
image

`fail_if_measurements_failed()` output in pytest:
image

Offline summary:
image

Online summary:
image

### DIFF
--- a/python/lib/sift_client/_internal/urls.py
+++ b/python/lib/sift_client/_internal/urls.py
@@ -19,7 +19,7 @@ _API_HOST_TO_FRONTEND_ORIGIN: dict[str, str] = {
 }
 
 
-def _origin(url: str) -> str:
+def parse_origin(url: str) -> str:
     """Normalize a URL or bare host into a ``scheme://host[:port]`` origin.
 
     Bare hosts (no scheme) are assumed to be ``https``.
@@ -29,7 +29,7 @@ def _origin(url: str) -> str:
     return f"{parsed.scheme}://{parsed.netloc}".rstrip("/")
 
 
-def _host(url: str) -> str:
+def parse_host(url: str) -> str:
     """Extract ``host[:port]`` from a URL or bare host string."""
     candidate = url if "://" in url else f"https://{url}"
     return urlparse(candidate).netloc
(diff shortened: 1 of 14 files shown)


python/lib/sift_client/_internal/urls.py
```diff
@@ -0,0 +1,55 @@
...
+    """Normalize a URL or bare host into a ``scheme://host[:port]`` origin.
+
+    Bare hosts (no scheme) are assumed to be ``https``.
+    """
+    candidate = url if "://" in url else f"https://{url}"
+    parsed = urlparse(candidate)
+    return f"{parsed.scheme}://{parsed.netloc}".rstrip("/")
+
+
+def _host(url: str) -> str:
```
ian-sift: i think transport.py in grpc+grpc_transport could reuse this (fine to do in future pr)

ian-sift (approved):

PR #599, opened 2026-05-29
python(feat): add update, archive, and unarchive to Channel

wei-qlu:
## Summary

You can now update channels from the Python Sift client. A channel's description, unit, metadata, and archived state are all editable, either through `ChannelsAPI.update(channel, update)` with a `ChannelUpdate` model or directly on a `Channel` with `update()`, `archive()`, and `unarchive()`. They all return the updated channel.

Units now work by name: an update takes something like "volts", and reads give back the name instead of an id. Reads also include `metadata` and `is_archived`, so updates round-trip.

## Validation

Unit tests cover the `ChannelUpdate` mapping, unit-name resolution, and the update/archive/unarchive methods. 

Manual QA: get a channel and confirm its unit comes back as a name, run `channel.update(ChannelUpdate(description="updated", unit="ms"))` and confirm both round-trip on the next read, then archive and unarchive it and confirm the state in the Sift app.


python/lib/sift_client/sift_types/channel.py
```diff
@@ -253,6 +254,10 @@ class Channel(BaseType[ChannelProto, "Channel"]):
     bit_field_elements: list[ChannelBitFieldElement] = Field(default_factory=list)
     enum_types: dict[str, int] = Field(default_factory=dict)
     asset_id: str
+    display_description: str
```
alexluck-sift: these should both just be "description", "unit"

python/lib/sift_client/sift_types/channel.py
```diff
@@ -253,6 +254,10 @@ class Channel(BaseType[ChannelProto, "Channel"]):
...
+    display_unit: str
+    metadata: dict[str, str | float | bool] = Field(default_factory=dict)
+    active: bool
```
alexluck-sift: for consistency, this should be "is_archived"

python/lib/sift_client/sift_types/channel.py
```diff
@@ -344,6 +382,35 @@ def runs(self) -> list[Run]:
         return self.asset.runs
 
 
+class ChannelUpdate(ModelUpdate[ChannelProto]):
+    """Model of the Channel fields that can be updated."""
+
+    display_description: str | None = None
+    display_unit: str | None = None
```
alexluck-sift: same feedback as the read model

python/lib/sift_client/sift_types/channel.py
```diff
@@ -332,6 +341,35 @@ def data(
...
+        self._update(updated_channel)
+        return self
+
+    def archive(self) -> Channel:
+        """Archive the channel by setting it inactive.
+
+        Returns:
+            The archived Channel.
+        """
+        return self.update({"active": False})
```
alexluck-sift: we have an archive on the resource that should be used instead

python/lib/sift_client/_internal/low_level_wrappers/channels.py
```diff
@@ -120,6 +122,21 @@ async def list_all_channels(
             max_results=max_results,
         )
 
+    async def update_channel(self, update: ChannelUpdate) -> Channel:
```
alexluck-sift, line 159: if we have a batch version, we should also implement that

python/lib/sift_client/sift_types/channel.py
alexluck-sift: oh we already have these. I don't think these fields should be here at all

python/lib/sift_client/_internal/low_level_wrappers/channels.py
wei-qlu: no batch rpc yet, I plan to make a ticket to implement that server side. current fix would be N individual updates

python/lib/sift_client/sift_types/channel.py
wei-qlu: updated

python/lib/sift_client/sift_types/channel.py
wei-qlu: updated

python/lib/sift_client/sift_types/channel.py
wei-qlu: updated

python/lib/sift_client/sift_types/channel.py
wei-qlu: updated

python/lib/sift_client/resources/channels.py
```diff
@@ -172,6 +175,7 @@ async def list_(
             max_results=limit,
             **({"page_size": page_size} if page_size is not None else {}),
         )
+        await self._resolve_unit_names(channels)
```
alexluck-sift: can we do this entirely in the low level client? The swapping from UUID to string is a bit clunky. We should never write the Channel object with anything but the string unit.

python/lib/sift_client/sift_types/channel.py
```diff
@@ -344,6 +379,58 @@ def runs(self) -> list[Run]:
...
+    def _add_resource_id_to_proto(self, proto_msg: ChannelProto):
+        if self._resource_id is None:
+            raise ValueError("Resource ID must be set before adding to proto")
+        proto_msg.channel_id = self._resource_id
+
+    def to_proto_with_mask(self) -> tuple[ChannelProto, field_mask_pb2.FieldMask]:
+        """Convert to proto with field mask."""
+        proto, mask = super().to_proto_with_mask()
+        # `is_archived` is the inverse of the proto's `active` field.
+        if self.is_archived is not None:
```
alexluck-sift: This is what the `converter` is for on the MappingHelper, shouldn't need to override `to_proto_with_mask`

python/lib/sift_client/resources/channels.py
wei-qlu: resolved

python/lib/sift_client/sift_types/channel.py
wei-qlu: resolved

alexluck-sift (approved):

PR #602, opened 2026-05-29
Python(feat): flexible pytest naming and cleaned up options

alexluck-sift:
## Summary

Make the pytest plugin's report fields configurable, and consolidate every plugin setting into one declarative registry. Report content can now be set from `pyproject.toml` (static project defaults) or `SIFT_REPORT_*` environment variables (per-run values). Adding or changing a setting is now a single registry entry instead of edits scattered across option registration, resolution, and docs.

## Configurable report content

- Static defaults live under `[tool.sift.pytest.report]` in `pyproject.toml`: `name`, `test_case`, `test_system_name`, `system_operator`, `serial_number`, `part_number`, and a `metadata` table.
- `name` and `test_case` accept placeholders: `{target}`, `{command}`, `{args}`, `{rootdir}`, `{timestamp}`, `{count}`, `{git_repo}`, `{git_branch}`, `{git_commit}`.
- Per-run values (CI, hardware-bench unit cycling) come from `SIFT_REPORT_TEST_SYSTEM_NAME`, `_SYSTEM_OPERATOR`, `_SERIAL_NUMBER`, and `_PART_NUMBER`. Env values win over TOML; `pytest-dotenv` loads `.env` for local dev.
- Report metadata is set from the `[tool.sift.pytest.report.metadata]` TOML table (typed scalar values). For dynamic per-run keys, attach them in conftest via the `report_context` fixture.
- The full pytest command is always recorded on the report under the `pytest_command` metadata key.
- When neither is configured, defaults are now predictable: `name` falls back to `{target} {timestamp}` and `test_case` to `{target}`. `target` is derived from the collected tests rather than the raw command line, so it stays consistent regardless of flag order, `-k`/`-m` filters, or the path form typed. It is anchored to the project name and narrows with the selection (single test, single file, common directory, or whole project), with the verbatim invocation still available via `{command}` and `pytest_command`.

## Settings resolution cleanup

- Every setting now lives in one `_Option` registry. Each entry declares only the surfaces it uses (CLI flag, ini key, TOML path, env var) and drives option registration, value resolution, the docs settings reference, and the unknown-key typo detector.
- Each setting has one clear home: pytest behavior in `[tool.pytest.ini_options]`, connection endpoints from env with ini fallback, report content in `[tool.sift.pytest.report]`. A CLI flag exists only where a real per-run override workflow justifies it.
- Unknown `SIFT_*` env vars and `[tool.sift.pytest.*]` keys produce a warning with a closest-match suggestion, so typos surface immediately.
- The web-app origin setting is now `app_url` (env `SIFT_APP_URL`, ini `sift_app_url`), aligned with the other connection URIs and `SiftConnectionConfig.app_url`.

## Migration

- The example project moves from `pytest.ini` to `pyproject.toml` to demonstrate the `[tool.sift.pytest.report]` configuration.

## Docs

- The configuration guide's settings reference is generated from the registry and kept in sync by `test_settings_reference_docs_in_sync`.

## Follow-up

- `_mixins/metadata.py` documents a planned `MetadataMixin`: metadata updates currently replace the whole map, so callers merge at the call site until the read-merge-write helper lands.

## Testing

- New tests cover report fields, report-name templating, the settings-reference sync, and the typo detector. The full plugin suite (127 tests) and `mypy lib` pass.


python/lib/sift_client/_internal/pyproject_config.py
```diff
@@ -0,0 +1,84 @@
...
+    1. ``config.inipath`` when it is itself a ``pyproject.toml`` (the common
+       case: project uses ``[tool.pytest.ini_options]`` so pytest loaded the
+       ini settings directly from pyproject).
+    2. ``<config.rootpath>/pyproject.toml``.
+    3. A bounded walk upward from ``rootpath`` for monorepo layouts where
+       pytest's rootdir is a subdirectory and the project pyproject lives
+       higher up.
+    """
+    inipath = config.inipath
+    if inipath is not None and inipath.name == "pyproject.toml" and inipath.is_file():
```
ian-sift, line 45: does this force all users to use pyproject.toml instead of a .ini file? can't pytest resolve ini options from both?

python/lib/sift_client/_tests/pytest_plugin/test_disabled.py
```diff
@@ -67,20 +67,6 @@ def test_disabled_does_not_require_credentials(
...
-        self,
-        pytester: pytest.Pytester,
-        clear_sift_env: None,
-        write_plugin_conftest: Callable[[], None],
-        monkeypatch: pytest.MonkeyPatch,
-    ) -> None:
-        """``SIFT_DISABLED=1`` triggers disabled mode without the CLI flag."""
-        write_plugin_conftest()
-        pytester.makepyfile("def test_runs(step): step.measure(name='v', value=1.0)")
-        monkeypatch.setenv("SIFT_DISABLED", "1")
```
ian-sift, line 80: is this no longer a settable env var?

python/lib/sift_client/_tests/pytest_plugin/test_report_fields.py
```diff
@@ -0,0 +1,300 @@
...
+
+def _create_report_dict(log_text: str) -> dict:
+    """Parse the JSON payload from the ``[CreateTestReport:...]`` log line."""
+    for line in log_text.splitlines():
+        if line.startswith("[CreateTestReport:"):
+            return json.loads(line[line.index("{") :])
+    raise AssertionError(f"no CreateTestReport line in log:\n{log_text}")
+
+
+def _metadata_pairs(report: dict) -> dict[str, object]:
```
ian-sift: this and _create_report_dict should probably be in a reusable place or we should match this tests approach to what we do elsewhere since I wouldn't expect that part to be unique.     Should metadata_proto_to_dict be used here? or should we add a new metadata helper to util/metadata.py

python/lib/sift_client/pytest_plugin.py
```diff
@@ -220,207 +222,429 @@ def _build_hierarchy_chain(
...
-    set either way lands on the same config slot. ``cli_flag=None`` makes the
-    option ini-only (e.g. the URI fallbacks).
+    """One setting, declared once across every surface it exposes.
+
+    Declare only the surfaces the setting uses. ``pytest_addoption``, the
+    resolvers (:meth:`resolve` / :meth:`resolve_merged`), the docs renderer
+    (:func:`_render_settings_reference`), and the typo detector all read from
+    the same registry, so adding or changing a setting is one edit.
+
+    Two shapes:
```
ian-sift: just pointing out this section was hard to read (assuming Claude did it?) Not sure what extra info this docstring should convey, but it doesn't seem like it would need to be much more than this class exists to contain the logic for resolving from the various potential places for specifying an option.

python/lib/sift_client/pytest_plugin.py
```diff
@@ -220,207 +222,429 @@ def _build_hierarchy_chain(
...
+        if self.ini_type and not self.ini:
+            raise ValueError(f"_Option({self.name!r}): ini_type requires ini")
+        if self.merge and not self.toml:
+            raise ValueError(f"_Option({self.name!r}): merge=True needs toml")
+        if not any([self.cli, self.ini, self.toml, self.env]):
+            raise ValueError(f"_Option({self.name!r}): declares no surfaces")
+        if self.category not in _CATEGORIES:
+            raise ValueError(f"_Option({self.name!r}): category must be one of {_CATEGORIES}")
+
+    def resolve(self, config: pytest.Config | None) -> Any:
```
ian-sift, line 318: this could maybe be a classmethod? Don't really care too much about this util class but it does feel weird how much the option class contains vs having all this logic exist in a parser class that returns options

python/docs/guides/pytest_plugin/configuration.md
```diff
@@ -120,64 +118,208 @@ def sift_client() -> SiftClient:
...
+**Precedence within a setting:** env > CLI flag > ini key > TOML > built-in
+default. No setting exposes both env and CLI, so the chain isn't ambiguous in
+practice.
+
+The plugin scans `SIFT_*` env vars and `[tool.sift.pytest.*]` keys at session
+start; anything outside these tables fires a warning with a closest-match
+suggestion, so typos like `SIFT_REPORT_SERIALNUM` surface immediately.
 
-## Ini options
+<!-- BEGIN settings-reference (auto-generated from _OPTIONS in pytest_plugin.py; regenerate via test_settings_reference_docs_in_sync) -->
```
ian-sift, line 144: is test_settings_reference_docs_in_sync called manually?

python/lib/sift_client/_tests/pytest_plugin/test_report_fields.py
```diff
@@ -0,0 +1,300 @@
...
+        result.assert_outcomes(passed=1)
+        pairs = _metadata_pairs(_create_report_dict(log_path.read_text()))
+        assert pairs.get("build_id") == "v1.2.3"
+        # Ints and floats share the proto's numeric slot.
+        assert pairs.get("lane") == 2
+        assert pairs.get("verbose") is True
+        # Auto-recorded keys still present alongside the typed entries.
+        assert "pytest_command" in pairs
+
+    def test_metadata_from_toml_table(
```
ian-sift: think all this is redundant to test_metadata_table_typed_values?

ian-sift (changes requested):

python/lib/sift_client/_internal/pyproject_config.py
alexluck-sift: when the pyproject.toml has a `tool.pytest.ini_options` section, the config.inipath will be set with the toml. If that is the case, we can just rely on pytest to tell us the toml file.

This function just finds the toml (if it is exists), it doesn't restrict the user from using an ini file to resolve the pytest settings.

python/lib/sift_client/_tests/pytest_plugin/test_disabled.py
alexluck-sift: correct, I wanted to have a clear set of rules about what types of settings are available from which surfaces. It didn't seem like this would be something that would change per-execution in a CLI (which is what I thought made sense for env vars other than credentials).

python/docs/guides/pytest_plugin/configuration.md
alexluck-sift: the test is run in CI, but the function can be run to produce the output to plop into the markdown in the event the test fails (that is, options update but docs not).

python/lib/sift_client/pytest_plugin.py
alexluck-sift: yeah, I was thinking of taking a pass at the structure of all of these functions/methods in the pytest_plugin.py. I'd like the pytest_plugin.py to contain only things users may want to export or read/understand and move everything else to _internal. Right now this py is a bit bloated. That would be a good time to revisit this kind of thing.

ian-sift (approved):

PR #609, opened 2026-06-02
Python(fix): add channel timeout

alexluck-sift:
## Summary

Synchronous SDK calls could hang indefinitely. A stalled gRPC connection, or a call issued after the client's background event loop had stopped during teardown, left the calling thread blocked forever because neither the RPC nor the blocking wait had a deadline. This adds a default request deadline plus two backstops so a stalled or post-shutdown call fails with a clear error instead of hanging.

## Changes

- **Default RPC deadline (primary).** Unary calls now carry a 60s deadline via the gRPC service config (`methodConfig.timeout`), alongside the existing retry policy. A stalled call fails with `DEADLINE_EXCEEDED` rather than hanging. Configurable per connection via `GrpcConfig(request_timeout=...)`, overridable per call, and disabled by passing `None`.
- **Bounded synchronous wait.** The sync wrapper's `.result()` now uses a deadline (RPC timeout plus a margin) and raises `TimeoutError` after cancelling the request. The gRPC deadline fires first, so this only trips if that fails.
- **Fail fast after close.** A call made once the background loop has stopped now raises a clear `RuntimeError` instead of scheduling onto a dead loop and blocking.

## Testing

- `_internal` unit suite passes, including new coverage for the sync deadline and the post-close guard.
- Confirmed the deadline lands in the channel service config: `60s` default, explicit override honored, omitted when disabled.

## Notes

The default of 60s sits above the retry budget (5 attempts, 5s max backoff). The service-config timeout applies to all methods via the wildcard matcher; this is safe because high-throughput streaming runs through the Rust stream bindings on a separate connection, not this channel.

### DIFF
--- a/python/lib/sift_client/_internal/sync_wrapper.py
+++ b/python/lib/sift_client/_internal/sync_wrapper.py
@@ -65,9 +65,7 @@ def generate_sync_api(cls: type[ResourceBase], sync_name: str) -> type:
             loop_running = loop.is_running()
         if not loop_running:
             coro.close()
-            raise RuntimeError(
-                "Sift client is closed; cannot make synchronous API calls."
-            )
+            raise RuntimeError("Sift client is closed; cannot make synchronous API calls.")
 
         timeout = getattr(client, "sync_call_timeout", None)
         future = asyncio.run_coroutine_threadsafe(coro, loop)
--- a/python/lib/sift_client/_internal/sync_wrapper.py
+++ b/python/lib/sift_client/_internal/sync_wrapper.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import concurrent.futures
 import inspect
 import sys
 from functools import wraps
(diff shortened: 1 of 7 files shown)


solidiquis (approved):

python/lib/sift_client/_internal/sync_wrapper.py
```diff
@@ -53,8 +54,29 @@ def __init__(self, *args, **kwargs):  # noqa: N807
...
+        if loop_running is None:
+            loop_running = loop.is_running()
+        if not loop_running:
+            coro.close()
+            raise RuntimeError("Sift client is closed; cannot make synchronous API calls.")
+
+        timeout = getattr(client, "sync_call_timeout", None)
+        future = asyncio.run_coroutine_threadsafe(coro, loop)
+        try:
+            return future.result(timeout=timeout)
```
wei-qlu, line 73: It seems like this timeout cap (75s) wrapping the coroutine will abort long-running sync processes like `wait_until_complete` / `wait_and_download` even if they're still polling, or if the user passes in a larger timeout_sec arg

PR #616, opened 2026-06-04
Python(feat): order-independent pytest report tree with early parent …

alexluck-sift:
### Why

The Sift pytest plugin builds a report tree where each test becomes a leaf step, nested under parent steps for its package, module, class, and parametrize axes. The previous design maintained those parents on a single shared stack, where a new step's parent is whatever is on top. One stack can only hold one open branch at a time, so sibling tests sharing a parent had to run back-to-back. To guarantee that, the plugin sorted every collected test item in `pytest_collection_modifyitems`.

That sort overrode the user's execution order. It was incompatible with `pytest-randomly`, `pytest-ordering`, dependency-ordered suites, and even pytest's own fixture-scope reordering (a session-scoped parametrized fixture can interleave tests across classes). Without the sort, an interleaved order tore the shared parent down and reopened it as a duplicate:

```
Old design, execution order [TestA::a1, TestB::b1, TestA::a2]:

  a1  stack: [mod, TestA, a1]
  b1  TestA != TestB -> pop+close TestA, push TestB   stack: [mod, TestB, b1]
  a2  TestB != TestA -> pop+close TestB, push TestA   (reopened from scratch)

  mod
  ├── TestA #1   └─ a1
  ├── TestB      └─ b1
  └── TestA #2   └─ a2      <- duplicate parent
```

A second limitation: every container step stayed `IN_PROGRESS` for the whole run and only resolved at session end, so an incrementally uploaded report showed all containers unresolved until the very end.

### What changed

Replace the stack-and-sort design with an identity-keyed parent registry. Parents are keyed by identity (`nodeid` for hierarchy parents; the test's param-stripped id plus axis frames for parametrize parents), created once, and reused by any descendant in any order:

```
New design, same interleaved order (any order works):

  a1  get-or-create TestA          leaf a1 -> TestA
  b1  reuse mod; get-or-create TestB    leaf b1 -> TestB
  a2  reuse mod; REUSE TestA            leaf a2 -> TestA

  mod
  ├── TestA   ├─ a1   └─ a2     <- created once, reused
  └── TestB   └─ b1
```

- The collection hook is now `pytest_itemcollected` (caches each item's chain) instead of `pytest_collection_modifyitems` (which sorted). The plugin never touches the items list, so it cannot conflict with another plugin's ordering hook.
- Each test's leaf is created under its resolved parent explicitly, rather than relying on stack position. This required explicit-parent step creation and parent-relative path numbering in `ReportContext` (`context_manager.py`).
- A parent closes as soon as the last test in its subtree finishes (tracked by counting descendant leaves at `pytest_collection_finish` and decrementing in `pytest_runtest_logfinish`), so containers resolve progressively. `finalize_parents` at session end is the backstop for anything still open (e.g. an aborted run).
- A parent's `end_time` is set to its last-descendant finish, so durations stay meaningful even though parents close lazily.

### Behavior notes

- Test execution order is preserved exactly as the user/other plugins set it.
- Container steps now resolve mid-run instead of all at session end.
- One accepted limitation: a test item injected after collection finishes is not counted, so its container could resolve slightly stale. This is rare, never crashes, and the report-level pass/fail verdict is unaffected.

### Testing

- Added `pytest-randomly` as a dev dependency (pinned to 3.15.0, the last release supporting Python 3.8, so randomization is active on the 3.8 CI test job).
- Added tests for order independence (interleaved execution produces no duplicate parents), early close (containers resolve before later siblings run), parent status propagation, and parent timing.
- Plugin suite passes in default order and under multiple random seeds; full non-integration suite is green; clean on ruff, mypy, and imports on Python 3.8 and 3.10.


python/lib/sift_client/pytest_plugin.py
```diff
@@ -359,28 +339,32 @@ def pytest_configure(config: pytest.Config) -> None:
...
+def pytest_itemcollected(item: pytest.Item) -> None:
+    """Cache each test item's hierarchy chain and parametrize path at collection.
+
+    This is a per-item hook, not ``pytest_collection_modifyitems`` — the plugin
+    never touches the ``items`` list or its order, so it cannot conflict with a
+    user's (or another plugin's) collection-ordering hook. The report tree is
+    built from an identity-keyed registry (see ``get_or_create_parent_chain``),
+    so item order is irrelevant to nesting; ``pytest-randomly``,
+    ``pytest-ordering``, and pytest's own fixture-scope reordering are all
+    preserved untouched.
```
ian-sift, line 351: idk if worth investigatin but claude claims pytest-rerunfailures will create duplicate parent steps which is probably... fine? idk how many users would use

ian-sift (approved):

python/lib/sift_client/pytest_plugin.py
alexluck-sift: ah interesting situation. Supporting that would require some new paradigm of "reopening" a previously evaluated step(s). How would we log past executions, etc. I think we can wait for a customer request on this since it is specific to a pytest plugin.

PR #617, opened 2026-06-05
rust(fix): fixes file attachement bug

Brandon-Shippy:
# Description

Fixes file attachment not working when importing files into sift via the CLI.
In `import/utils` when making the rest request to upload the zipped file we were not including `CONTENT_DISPOSITION` header to the request, as such our file names would not be preserved when ingested and not accessible via the run.

# Validation

Manual validation of all file types.


solidiquis (approved):
Could you fix the branch title before merging

PR #618, opened 2026-06-05
rust(feat): SiftStreamAutoRegister — inline flow registration wrapper

tsift:
Adds `SiftStreamAutoRegister<T>` to sift_stream and `SiftStreamAutoRegisterPy` to sift_stream_bindings.

Flows are registered with Sift automatically on first send. If a `FlowConfig` for the flow was provided during construction of the `SiftStreamAutoRegister` struct, that will be used to register to the `Flow`. This allows providing types such as enums and bitarrays, as well as channel descriptions, units, and other metadata. Otherwise, a minimal config is generated based on the channel-name and type (the only available information at ingest time).

Known flows can still be pre-registered via `IngestionConfigForm` at init time to eliminate first-send latency. Subsequent sends for any registered flow are local cache-hits with no network overhead.

### DIFF
--- a/rust/crates/sift_stream/examples/auto-register/main.rs
+++ b/rust/crates/sift_stream/examples/auto-register/main.rs
@@ -1,21 +1,22 @@
 // Demonstrates SiftStreamAutoRegister: inline flow registration without pre-declaring schemas.
 //
-// The stream is initialized with an empty flow list. On the first send for any flow name,
-// SiftStreamAutoRegister derives a FlowConfig from the Flow's channel values and registers
-// it with Sift automatically. Subsequent sends for the same flow name are cache-hits and
-// have no extra overhead.
+// Three flows show the full registration spectrum:
 //
-// Staged configs let you attach metadata (units, descriptions) to a flow without pre-declaring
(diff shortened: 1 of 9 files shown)


rust/crates/sift_stream/examples/auto-register/main.rs
```diff
@@ -0,0 +1,108 @@
...
+        .attach_run(RunForm {
+            name: run_name.clone(),
+            client_key: run_name,
+            ..Default::default()
+        })
+        .live_with_backups()
+        .build()
+        .await?;
+
+    let mut auto = SiftStreamAutoRegister::new(stream, vec![]);
```
marc-sift: Add a staged flow here just to show an example.

marc-sift (changes requested):

rust/crates/sift_stream/examples/auto-register/main.rs
tsift: Good idea! Added in 3b1141a0469d040bcaaec969e01543a8bcee6bac

marc-sift (approved):

PR #623, opened 2026-06-08
python(fix): keep test-results logging from hanging sync calls

wei-qlu:
## Summary

The test-results offline-log path acquired a file lock with no timeout on the shared event-loop thread; a contended or stale lock froze the loop and every synchronous API call hung indefinitely.

This change moves the log helpers to `filelock.AsyncFileLock` (floor raised to `~=3.15`):

- Lock waiting is cooperative, so a stuck lock parks only the waiting call, never the event loop.
- The wait is bounded at 60s and raises `TimeoutError` instead of hanging (the lib default waits forever).
- `run_in_executor=False` makes acquisition cancel-safe; a cancelled task could otherwise strand the lock for the process lifetime.

## Test plan

- Writer and reader fail fast with `TimeoutError` while the lock is held.
- Unrelated coroutines still complete while a call waits on the held lock.
- Cancels never strand the lock (pins `run_in_executor=False`).
- Read/parse round trip and the 500-append race test cover the data path.
- Passes local integration tests

### DIFF
--- a/python/lib/sift_client/_internal/low_level_wrappers/_test_results_log.py
+++ b/python/lib/sift_client/_internal/low_level_wrappers/_test_results_log.py
@@ -39,7 +39,9 @@ Ad-hoc writers to the same path are not protected.
 
 from __future__ import annotations
 
+import asyncio
 import atexit
+import functools
 import json
 import os
 import re
@@ -165,7 +167,7 @@ class ReplayResult:
     measurements: list[TestMeasurement] = field(default_factory=list)
 
 
-def log_request_to_file(
+async def log_request_to_file(
     log_file: str | Path,
     request_type: str,
     request: Any,
@@ -179,8 +181,8 @@ def log_request_to_file(
     mid-write partial final line. See the module docstring for the full
     concurrency model.
 
-    This is synchronous and blocks on the lock; async callers should offload it
-    off the event loop (see ``_LOG_IO_EXECUTOR``).
+    The blocking lock + file I/O runs on ``_LOG_IO_EXECUTOR``, so a slow or
+    stuck lock cannot block the event loop.
 
     Args:
         log_file: Path to the log file.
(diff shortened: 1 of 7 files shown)


python/lib/sift_client/_internal/low_level_wrappers/test_results.py
```diff
@@ -860,7 +881,10 @@ async def update_test_measurement(
 
         if log_file is not None or simulate:
             if log_file is not None:
-                log_request_to_file(log_file, "UpdateTestMeasurement", request)
+                await run_sync_function(
```
alexluck-sift: this is a lot dev overheard to remember to call each time. Can we instead change `log_request_to_file` to async and we can just call that directly? 

Same for the read path.

python/lib/sift_client/_internal/low_level_wrappers/_test_results_log.py
```diff
@@ -155,59 +170,95 @@ def log_request_to_file(
...
     request_json = json.dumps(request_dict, separators=(",", ":"))
     line = f"[{tag}] {request_json}\n"
-    with FileLock(str(log_path.with_name(log_path.name + ".lock"))):
-        with open(log_path, "a") as f:
-            # The inner ``with`` flushes and closes the file before the
-            # FileLock is released, so no reader sees a partial line.
-            f.write(line)
+    lock_path = log_path.with_name(log_path.name + ".lock")
+    try:
+        with FileLock(str(lock_path), timeout=timeout):
```
alexluck-sift: async support for filelock appears to have been added in newer releases. I think we can just update that and most of the other changes are entirely unneeded.  See v3.15.

alexluck-sift (changes requested):

python/lib/sift_client/_internal/low_level_wrappers/test_results.py
wei-qlu: updated, both helpers are async and call sites just await them directly

python/lib/sift_client/_internal/low_level_wrappers/_test_results_log.py
wei-qlu: updated to AsyncFileLock. Kept the 60s timeout (lib default waits forever) and set run_in_executor=False since executor-based acquire isn't cancel-safe; test added.

alexluck-sift (approved):

ian-sift (approved):

PR #624, opened 2026-06-09
python(feat): add auto register client for lazy flow config registration

tsift:
Adds a high-level `AutoRegisterStreamingClient` and factory method `IngestionAPIAsync.create_auto_register_streaming_client()` that wraps the `SiftStreamAutoRegisterPy` binding.  Flows are registered automatically on first send.


wei-qlu (dismissed):

python/lib/sift_client/resources/ingestion.py
```diff
@@ -194,6 +197,66 @@ async def create_ingestion_config_streaming_client(
...
+            disk_backup_policy: Disk backup policy for LIVE_WITH_BACKUPS or FILE_BACKUP mode.
+            checkpoint_interval_seconds: Checkpoint interval in seconds (LIVE_WITH_BACKUPS only).
+            enable_tls: Whether to enable TLS for the connection.
+            tracing_config: Configuration for SiftStream tracing.
+            staged_configs: Optional flow configs to use when auto-registering flows for the first
+                time. When a staged config exists for a flow, it is used for registration instead
+                of a minimal derived config, preserving units, descriptions, and other metadata.
+                The staged config is validated against the flow's channel names and types before
+                use; a mismatch raises `RuntimeError`. Each staged config is consumed after
+                successful registration.
```
marc-sift: > Each staged config is consumed after successful registration.

This sounds like rust speak that might not be nessecary here? Is there any impact to the user regarding this?

python/lib/sift_client/resources/ingestion.py
```diff
@@ -194,6 +197,66 @@ async def create_ingestion_config_streaming_client(
...
+            run: The run to associate with ingestion. Can be a Run, RunCreate, dict, or run ID string.
+            asset_tags: Tags to associate with the asset.
+            asset_metadata: Metadata to associate with the asset.
+            streaming_mode: Transport mode for the stream. Defaults to LIVE_WITH_BACKUPS.
+            retry_policy: Retry policy for LIVE_WITH_BACKUPS mode.
+            disk_backup_policy: Disk backup policy for LIVE_WITH_BACKUPS or FILE_BACKUP mode.
+            checkpoint_interval_seconds: Checkpoint interval in seconds (LIVE_WITH_BACKUPS only).
+            enable_tls: Whether to enable TLS for the connection.
+            tracing_config: Configuration for SiftStream tracing.
+            staged_configs: Optional flow configs to use when auto-registering flows for the first
```
marc-sift, line 244: We should add some info somewhere in this docstring why a user would want to use AutoRegisterStreamingClient. For example, if their ingestion config is too large and exceeds the grpc max message size.

marc-sift (changes requested):

python/lib/sift_client/resources/ingestion.py
tsift: Added: be8ad883662704ee65112c4fff1f1d9ac01f8582

python/lib/sift_client/resources/ingestion.py
tsift: Ah, yes good call out. Its an implementation detail so not helpful including there.
be8ad883662704ee65112c4fff1f1d9ac01f8582

marc-sift (approved):

PR #628, opened 2026-06-09
python(feat): add audit logging to pytest plugin

alexluck-sift:
Adds a DEBUG audit trace to the pytest plugin so field failures can be diagnosed from one file without a rerun.

- New `--sift-audit-log` / `sift_audit_log` option (`path | true | false`). On by default: DEBUG trace to a temp file, WARNINGs echoed to stdout. Set a path to pin the file, or `false` to disable.
- Columnar `key=value` lines trace config typos, credentials, connection, the resolved-settings snapshot (with `env`/`cli`/`ini`/`toml` provenance; API key redacted), status transitions, templates, step open/close, and the replay lifecycle.
- Per-session temp artifacts (JSONL log, tracking sidecar, audit log, replay sibling) now group under `<tmpdir>/sift_test_results/<random>/`; the replay subprocess gets its own `*.replay.log`. Temp logs are reclaimed on a clean finish only when audit logging is off.
- Terminal summary gains an `audit log` section; `render_report_tree` shows the final step tree with per-step status.
- Docs: a "Getting support" section in the plugin overview explains the trace, where to find it, and to zip the session directory when contacting Sift. `pyproject.toml` sets a stable `test_case` grouping key for integration runs.
- Tests added for configuration, incremental replay, and report context.
- Adds a fixture to assign report metadata that users can implement/override to have custom metadata handling

Example outputs:
[eyw4bltc-audit.replay.log](https://github.com/user-attachments/files/28814448/eyw4bltc-audit.replay.log)
[eyw4bltc-audit.log](https://github.com/user-attachments/files/28814449/eyw4bltc-audit.log)

### DIFF
--- a/python/lib/sift_client/_internal/pytest_plugin/audit_log.py
+++ b/python/lib/sift_client/_internal/pytest_plugin/audit_log.py
@@ -1,0 +1,237 @@
+"""DEBUG audit trace for the pytest plugin (file) plus WARNING echo (stdout).
+
+On by default: every session attaches two handlers to the ``sift_client`` root
+logger so plugin-behavior modules AND high-value SDK call sites land in one file
+(a temp file unless ``--sift-audit-log=<path>`` pins one), with warnings also
+echoed to stdout. Pass ``--sift-audit-log=false`` (or set ``sift_audit_log =
+"false"``) to turn it off. The replay subprocess gets its own sibling file via
+``replay_audit_path``.
+
+Handlers are removed at session end (``pytest_unconfigure`` ->
+``detach_audit_handlers``) so a process that runs many pytest sessions — the
+plugin's own test suite drives nested in-process sessions — doesn't accumulate
+handlers or leak one session's file into the next.
+
+TODO: levels are fixed (DEBUG file / WARNING stdout) and output is plain text.
+A configurable level, JSON lines, or rotation are follow-ups gated on real need.
+"""
+
+from __future__ import annotations
+
+import logging
+import sys
+import tempfile
+from pathlib import Path
+from typing import TYPE_CHECKING, Any
+
+if TYPE_CHECKING:
+    import pytest
+
+ROOT_LOGGER = "sift_client"
+# Columnar line for easy parsing: fixed-width timestamp, level, and namespace
+# columns, then the message — which every plugin call starts with a
+# left-justified ``EVENT_WIDTH`` event token followed by space-separated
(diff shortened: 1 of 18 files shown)


python/lib/sift_client/_internal/pytest_plugin/report.py
```diff
@@ -205,6 +209,40 @@ def resolve_initial_status(new_step: NewStep, item: pytest.Item) -> None:
...
+    step = new_step.current_step
+    if step is not None and step.error_info is not None and step.error_info.error_message:
+        return step.error_info.error_message.strip().splitlines()[0]
+    return ""
+
+
+def skip_or_xfail_reason(phase: Any) -> str:
+    """The skip/xfail reason text from a pytest phase, or empty when none.
+
+    xfail reasons ride on ``report.wasxfail``; plain skips store
```
ian-sift, line 231: what is wasxfail?

python/lib/sift_client/_internal/pytest_plugin/audit_log.py
```diff
@@ -0,0 +1,263 @@
...
+echoed to stdout. Pass ``--sift-audit-log=false`` (or set ``sift_audit_log =
+"false"``) to turn it off. The replay subprocess gets its own sibling file via
+``replay_audit_path``.
+
+Handlers are removed at session end (``pytest_unconfigure`` ->
+``detach_audit_handlers``) so a process that runs many pytest sessions — the
+plugin's own test suite drives nested in-process sessions — doesn't accumulate
+handlers or leak one session's file into the next.
+
+TODO: levels are fixed (DEBUG file / WARNING stdout) and output is plain text.
```
ian-sift, line 15: Is this not a blocker for rolling this out?

ian-sift (approved):

python/lib/sift_client/_internal/pytest_plugin/report.py
alexluck-sift: xfail lets users mark tests as expected to fail: https://docs.pytest.org/en/stable/how-to/skipping.html#xfail-mark-test-functions-as-expected-to-fail

So this comes from pytest and tells us if it was an xfail so we can record the status reason.

python/lib/sift_client/_internal/pytest_plugin/audit_log.py
alexluck-sift: I think we can come back to this if customers request it. Right now we write to tmp by default, so these will get cleaned up automatically.

PR #641, opened 2026-06-16
rust(feat): mcp rate-limiting

Brandon-Shippy:
# Description
Implements gRPC status handling for backoff and retries in the mcp. 

We construct each Service with a `Retry Policy` that outlines how to handle gRPC status codes. In here we implement a retry loop that takes in a closure. With this new structure we wrap our gRPC calls within the retry loop so we can handle the gRPC status codes when they come in.

When the retry layer gives up, we give the agent a soft-signal to not retry this tool without the user providing more context or the general shape of the request is different. We do handle the different tonic statuses differently.

Following the status codes laid out here: https://google.aip.dev/194 

`UNAVAILABLE` = Retry with backoff, then soft signal on exhaustion
`RESOURCE_EXHAUSTED, DEADLINE_EXCEEDED, INTERNAL, ABORTED, ALREADY_EXISTS, PERMISSION_DENIED, UNAUTHENTICATED, CANCELLED` = Immediate error return with soft signal

Errors without set soft-guidance are given a hard rmcp error code back:

`INVALID_ARGUMENT and OUT_OF_RANGE` = `INVALID_PARAMS`
`NOT_FOUND` = `RESOURCE_NOT_FOUND`
`UNIMPLEMENTED` = `METHOD_NOT_FOUND`
`FAILED_PRECONDITION, DATA_LOSS, UNKNOWN` = `INTERNAL_ERROR`

What this does not limit:
- Number of tools being used at once
- Total request count over a session

# Verification
- Mock tests
- Unit tests
- Pointed the mcp to an unreachable address and did manual testing to get errors back

### DIFF
--- a/rust/crates/sift_mcp/src/error/mod.rs
+++ b/rust/crates/sift_mcp/src/error/mod.rs
@@ -1,4 +1,4 @@
-use rmcp::model::{CallToolResult, Content, ErrorCode};
+use rmcp::model::{CallToolResult, ErrorCode};
 use tonic::{Code, Status};
 
 #[cfg(test)]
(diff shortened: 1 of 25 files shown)


rust/crates/sift_mcp/src/service/data/mod.rs
```diff
@@ -253,27 +255,42 @@ impl DataService {
...
-                    end_time: Some(end_time),
-                    page_size: PAGE_SIZE,
-                })
-                .await
-                .context("failed to get data")?;
+            let channel = self.channel.clone();
+            let queries = queries.clone();
+            let token = page_token.clone();
+            let start_time = start_time;
+            let end_time = end_time;
```
evan-sift: I don't see us overwriting `start_time` / `end_time`. am I missing something?

rust/crates/sift_mcp/src/error/mod.rs
```diff
@@ -1,11 +1,121 @@
...
+    BackendError,
+    Conflict,
+    AlreadyExists,
+    PermissionDenied,
+    Unauthenticated,
+    Cancelled,
+}
+
+impl AgentSignal {
+    fn from_code(code: Code) -> Option<Self> {
```
solidiquis: Prefer to use the std lib [From](https://doc.rust-lang.org/std/convert/trait.From.html):

```rust
impl From<Code> for AgentSignal {
    fn from(code: Code) -> Self {
        todo!()
    }
}
```

solidiquis: In this case you probably want [TryFrom](https://doc.rust-lang.org/std/convert/trait.TryFrom.html) where the error type can be `anyhow::Error`

solidiquis (changes requested):

rust/crates/sift_mcp/src/service/data/mod.rs
Brandon-Shippy: https://github.com/sift-stack/sift/pull/641/commits/cc4f6a7582ea686133d32cf334139068e6e48035

rust/crates/sift_mcp/src/error/mod.rs
Brandon-Shippy: https://github.com/sift-stack/sift/pull/641/commits/157e4398c7c668b6611dde1d91ec5155cf606852

solidiquis (approved):

evan-sift (approved):

PR #646, opened 2026-06-22
python(feat): add ABAC resource and principal attributes

wei-qlu:
## Summary

Adds the public API for ABAC resource and principal attributes to `sift_client`.

- `client.resource_attributes` manages attribute keys on entities (assets, channels, runs).
- `client.principal_attributes` manages keys on principals (users, user groups).

The attribute key is the entry point: create or fetch a key, define enum values, then assign values to entities or principals. `SET_OF_ENUM` assignments replace the full value set per target. Principal assignment accepts user IDs or emails, resolving emails to user IDs via REST since the gRPC `User` proto has no email field.

Also fixes `gen_pyi.py` to render enum parameter defaults as valid stub syntax.

## Test plan

- Unit tests for the types, low-level wrappers, and resource orchestration (get-or-create dedup, value/entity/email resolution, pagination)
- Manual QA checklist covers the live-backend flows (key/enum-value/assignment CRUD, archive cascade, email resolution, sync/async parity).


nathan-sift (approved):
lgtm. There are some more user friendly additions we can consider in the future to make working with the API easier, but I don't think any of that is necessary or in scope for this PR.

PR #650, opened 2026-06-23
Get data cache improvements

ian-sift:
## Summary
Replaces the unbounded in-process channel cache that OOMed a customer pod with a shared, bounded, on-disk cache, and speeds up `get_data` pagination on the side. Adds a public `client.cache` namespace for configuration and observability.
Motivating incident: a single 21-minute sustained pull ate ~5–7 GB on an 8 GiB pod and crashed it. Cause: `DataLowLevelClient.channel_cache` was a class-level attribute shared across every `SiftClient` in the process, with no eviction. The first commit on the branch bounds it; the rest of the branch rebuilds the cache around a real storage layer.
### Cache architecture
- **New shared on-disk store.** One `DiskCache` (`diskcache`-backed, SQLite under the hood) lives on `SiftClient`; every cache-aware resource writes to one global byte budget at one path with one LRU policy. The old class-level dict is gone.
- **Typed adapter on top.** `ChannelDataCache` wraps the shared store with channel-specific key namespacing (`channel:<run_id>:<id>`) and isinstance-checked reads so cross-adapter writes can't return the wrong shape downstream.
- **Run-scoped keys.** Cache buckets are keyed by `(channel_id, run_id)`. Same channel under two runs is two cache buckets — a query for run B no longer short-circuits on run A's cached entry.
- **In-memory tier removed.** Earlier drafts had a two-tier (memory + disk) layout; benchmarking showed the memory tier wasn't pulling its weight against `diskcache`'s built-in OS page cache, so the in-process layer was torn out. Reads always go through the disk path.
- **Oversize safety rail.** A single fetch larger than the byte cap is now skipped (with a deduped warning) rather than cache-busting every other entry trying to fit.
- **`ignore_cache=True` skips both reads and writes.** Pre-fix, "non-caching" workloads still appended to the shared cache on every call.
### Pagination perf
- Replaces the inline O(N²) per-page `pd.concat` loop in `_handle_pagination` / `_merge_pages` with one batched concat per channel. Bench numbers from same-shape inputs:
  - 10 pages × 10k rows: **6.3× faster** (22 ms → 3.5 ms)
  - 50 pages × 10k rows: **26× faster** (488 ms → 19 ms)
  - Customer-reported 145M-point pulls scale closer to **~80×** at the tail.
### Public API additions (`client.cache`)
| Method | Purpose |
| --- | --- |
| `client.cache.enable(*, path=None, max_bytes=None)` | Open / reconfigure the on-disk cache. No args = use defaults. |
| `client.cache.disable()` | Close the handle. Reads miss, writes drop. Directory is left intact. |
| `client.cache.clear(path=None)` | Delete a cache directory entirely (refuses paths without the diskcache marker file as a typo safety net). |
| `client.cache.stats() -> CacheStats` | Read-only snapshot (path, used/cap bytes, entry count, channel entry count). `str(stats)` prints a notebook-friendly summary. |
`CacheStats` and `CacheNamespace` live in `sift_client.util.cache` and are re-exported from the top-level `sift_client` package so type hints work and `mkdocstrings` picks them up.
## Configuration
Defaults are opt-out — caching is on at first use, no user code required:
| Knob | Default | Notes |
| --- | --- | --- |
| Enabled | **on** | Call `client.cache.disable()` to turn off. |
| Path | `<tempfile.gettempdir()>/sift-data-cache` | Process-shared; a fresh process reads prior-session entries. |
| Byte cap | **4 GiB** | `diskcache` LRU evicts once approached. |
| Failure mode (default path) | warn + degrade to no-cache | Read-only `/tmp`, restricted containers, etc. — `get_data` still works, just always goes to the wire. |
| Failure mode (explicit path) | raise | User asked for it specifically; the failure surfaces. |
```python
# Off (no on-disk state)
client.cache.disable()
# Custom location / size
client.cache.enable(path="/data/sift-cache", max_bytes=2 * 1024 ** 3)
# Remove a stale or corrupted cache directory
client.cache.clear()                  # default tmp path
client.cache.clear("/data/sift-cache")
# Inspect
print(client.cache.stats())
# Sift cache:
#   path:     /tmp/sift-data-cache
#   used:     142.3 MiB / 4.0 GiB (3.5%)
#   entries:  487 (487 channel)
```
### Memory
- The OOM regression goes away by construction: cache state is on disk, nothing accumulates in process across `get_data` calls. The customer's identical-workload re-run on the workaround branch stayed flat at 1–1.5 GiB; this branch matches that without the manual purge.
### Latency
- **Per-call CPU savings** from the batched concat scale with page count: 6×–26× on the bench, more on long-tail pulls.
- **Wire savings** kick in on the second-and-onward call covering the same `(channel, run, window)` — those are served from disk instead of paged from the API. Cross-process savings work too (default path persists across runs).
### Disk usage
- Bounded by `max_bytes` (default 4 GiB). LRU eviction past the cap.
- Total includes SQLite overhead; `client.cache.stats().size_bytes` is the same number diskcache's eviction loop uses.
## Behavior changes worth flagging
- **`ignore_cache=True` no longer writes** to the cache. Code paths that relied on the cache being warmed by an `ignore_cache=True` call need to drop that flag.
- **Resource-level cache knobs are gone.** Previously each cache-aware resource exposed its own enable/disable methods; configuration is now global at `client.cache.*` because the store is shared.
- **Class-level cache state is gone.** Two `SiftClient` instances in the same process no longer share cache state via class attributes; they share via the on-disk store at the configured path.
- **Empty results aren't cached** (`get_data` returning zero rows now re-fetches on the next identical call instead of cache-hitting an empty frame).

## Verification
Test suite updates: default-path lazy init + silent fallback, `disable`/`enable` reconfigure-in-place semantics, oversize entry skip+warn (with dedup), run-id key scoping, bitfield element grouping, `ignore_cache=True` skipping writes, cross-process cache reuse via the same path, prefix-scoped `clear`, `stats()` for disabled / empty / populated / foreign-prefix shapes.


python/lib/sift_client/_internal/low_level_wrappers/data.py
```diff
@@ -298,22 +391,47 @@ async def get_channel_data(
...
-            channel_data=ret_data, start_time=start_time, end_time=end_time, run_id=run_id
-        )
+        if not ignore_cache:
+            self._update_cache(
+                channel_data=ret_data, start_time=start_time, end_time=end_time, run_id=run_id
+            )
 
         return ret_data
 
+    def _merge_pages(
```
ian-sift, line 831: Replaces a per-page ``pd.concat(...).groupby(...)`` loop that was
        O(N²) in the number of pages — each iteration copied the cumulative
        DataFrame — with a single batched concat per channel. At realistic
        pagination depths the speedup is large: 200 pages of 10k rows each
        drops from ~11 s to ~130 ms in the bench.

python/lib/sift_client/client.py
```diff
@@ -136,6 +136,7 @@ def __init__(
         rest_url: str | None = None,
         connection_config: SiftConnectionConfig | None = None,
         app_url: str | None = None,
+        data_cache_max_bytes: int | None = None,
```
alexluck-sift: We should avoid adding any resource specific settings at the client level. Instead, can we set this on the specific resource?

python/lib/sift_client/_internal/low_level_wrappers/data.py
```diff
@@ -34,17 +35,106 @@
...
+    def invalidate(self, channel_id: str) -> None:
+        prior = self._entries.pop(channel_id, None)
+        if prior is not None:
+            self._total_bytes -= prior.size_bytes
+
+    def clear(self) -> None:
+        self._entries.clear()
+        self._total_bytes = 0
+
+    def _evict_until_under_bound(self) -> None:
```
alexluck-sift: Is there a need for this specific behavior, or does the cache disabling solve our current need?

I am weary to add more complexity to the in memory cache when longer term we would want to use a on-disk cache.

ian-sift: part of the request was for configurable cache size. I tested a diskcache(fastest i could find) implementation which added 10-15seconds/GB. And there is some concerns there about adhering to FedRamp policy re: storing user data on disk. Also some concerns w/ cleanup w/ interrupted processes (i think mitigated by storing in /tmp which should be cleaned up by OS or pod teardown).

python/CHANGELOG.md
```diff
@@ -7,6 +7,53 @@ This project adheres to [Semantic Versioning](http://semver.org/).
...
+```python
+client.channels.configure_data_cache(max_bytes=128 * 1024 * 1024)  # 128 MiB cap
+client.channels.configure_data_cache(max_bytes=0)                  # disable caching
+```
+
+`configure_data_cache` may be called at any time; if the cache is already populated, the new bound is applied immediately and excess entries are evicted.
+
+`ignore_cache=True` on `client.channels.get_data(...)` now also skips writing into the cache, matching its read-side bypass semantics. Previously a "non-caching" workload still appended to the shared cache on every call, which still caused increased memory usage.
+
+The internal `DataLowLevelClient.channel_cache` is no longer a class attribute. Any external code that relied on `DataLowLevelClient.channel_cache.channels.clear()` as a workaround should remove it — the bounded cache no longer requires manual purging.
```
alexluck-sift: This can be omitted since the low level client is private/internal.

python/CHANGELOG.md
```diff
@@ -7,6 +7,53 @@ This project adheres to [Semantic Versioning](http://semver.org/).
...
+
+#### On-disk channel data cache (opt-in)
+
+The channel data cache can now optionally persist to disk, surviving process restarts. The disk tier is a second-chance layer beneath the in-memory cache: on a memory miss, `get_data` checks disk before going to the wire. Re-running the same workload in a new session picks up the previously-cached windows for free.
```
alexluck-sift: This should be the new default and in-memory opt-in. Additionally, do we have any reason to keep the in-memory if we are adding an on-disk? The performance of on-disk should be more than sufficient compared to querying over the network.

If we need to keep the in-memory for some reason, I don't think we should be using an in-memory AND on-disk cache.

python/CHANGELOG.md
```diff
@@ -7,6 +7,53 @@ This project adheres to [Semantic Versioning](http://semver.org/).
...
+```
+
+To remove a stale cache directory from a previous session:
+
+```python
+client.channels.clear_data_cache_on_disk()                   # default tmp path
+client.channels.clear_data_cache_on_disk("/data/sift-cache") # custom path
+```
+
+`clear_data_cache_on_disk` refuses to delete directories that don't look like a sift channel data cache (missing the `diskcache` marker), so a typo'd path won't wipe unrelated data.
```
alexluck-sift: is this comment really necessary?

python/lib/sift_client/client.py
```diff
@@ -148,6 +148,11 @@ def __init__(
                 Set this for on-prem or custom deployments whose API host can't be
                 mapped to a frontend automatically; see the ``app_url`` property.
                 A value here takes precedence over ``connection_config.app_url``.
+
+        Resource-specific knobs live on the resource itself. For example,
```
alexluck-sift: I don't think this comment is necessary here and just one more thing to maintain as the resources are updated.

python/lib/sift_client/resources/channels.py
```diff
@@ -64,6 +65,114 @@ def __init__(self, sift_client: SiftClient):
         self._low_level_client = ChannelsLowLevelClient(grpc_client=self.client.grpc_client)
         self._units_low_level_client = UnitsLowLevelClient(grpc_client=self.client.grpc_client)
         self._data_low_level_client = None
+        # Caller-supplied cache size; ``None`` means "use the wrapper default
```
alexluck-sift: It would make sense to move diskcache outside of the individual resource. We may want to use this for other things in the future and tying it directly to the channel API will make this more difficult.

python/uv.lock
```diff
@@ -1,5 +1,5 @@
 version = 1
-revision = 3
+revision = 2
```
alexluck-sift: I think you may be on an older uv version

alexluck-sift: Might need to add  `required-version = ">=0.8"` to our pyproject.toml `tool.uv` to ensure we keep the version

python/pyproject.toml
```diff
@@ -350,6 +351,12 @@ ignore_errors = true
 [[tool.mypy.overrides]]
 module = "nptdms"
 ignore_missing_imports = true
+
+# diskcache ships without inline type hints or PEP 561 marker. Used by the
+# channel data cache's optional on-disk tier.
+[[tool.mypy.overrides]]
+module = "diskcache"
+ignore_missing_imports = true
 ignore_errors = true
```
alexluck-sift, line 368: looks like this insert unintentionally removed ignore_errors from the nptdms overrides

python/lib/sift_client/_internal/low_level_wrappers/data.py
```diff
@@ -34,17 +41,347 @@
...
+
+    def disable_disk(self) -> None:
+        """Close the disk tier (if open). Does not touch the disk contents.
+
+        Use ``sift_client.clear_data_cache_on_disk(path)`` to remove a
+        directory from disk.
+        """
+        self._close_disk()
+
+    def get(self, channel_id: str) -> ChannelCacheEntry | None:
```
alexluck-sift: when a single channel is larger than the cache size, it essentially breaks the cache by wiping the entire thing on every call. Perhaps instead of putting it into the cache if that is the case, we just skip adding it or cache busting and emit a warning.

python/CHANGELOG.md
ian-sift: in memory and disk cache seems very standard in computing no? I can remove for now though to get this over the line faster

python/lib/sift_client/_internal/disk_cache_config.py
alexluck-sift, line 1: Looks good! One more tweak to make this more reusable and only a single cache for users to manage/be concerned about:

## Proposal

Split `ChannelCache` into a shared storage layer and a typed adapter, with ownership on the client. Composition, not inheritance.

**1. Extract `DiskCache`, one concrete shared store** (`_internal/disk_cache.py`, sibling to `DiskCacheConfig`). Lift the diskcache mechanics out of `ChannelCache` as-is: the `diskcache.Cache` lifecycle, byte cap and LRU eviction, the marker-guarded `clear_disk` classmethod, the no-op-when-disabled behavior, and the oversized skip+warn (keyed on the full namespaced key rather than channel ID). Keep it key/value-agnostic: the caller hands it `size_bytes`, and diskcache pickles the value, so the store never needs to know it's holding a DataFrame.

**2. Own one instance on the client.** Create it lazily on `SiftClient` (`client._disk_cache`), configured from the single `DiskCacheConfig`. Resources hold a reference to it rather than constructing their own. The default path drops the "channel" prefix (`.../sift-data-cache`), so `max_bytes` is one global disk budget and LRU spans every resource that uses the store.

**3. Write `ChannelDataCache`, a concrete adapter over the shared store.** It holds `client._disk_cache`, namespaces keys (`f"channel:{id}"`), keeps `ChannelCacheEntry`, computes `size_bytes` via `memory_usage(deep=True)`, and delegates everything else. Key namespacing keeps resources from colliding in the shared store. A second resource adds its own adapter and prefix without touching the storage layer; the `DiskCacheAdapter` Protocol gets written then, from two real shapes.

**4. Move the public knobs to the client.** Since the store is shared, `enable_data_cache_disk` / `disable_data_cache_disk` / `clear_data_cache_on_disk` go on the client (or a `client.cache` namespace), not on `channels`. The API is still `[Unreleased]`, so move it clean with no shim or back-compat.

python/lib/sift_client/_internal/disk_cache.py
```diff
@@ -0,0 +1,336 @@
...
+            :attr:`DEFAULT_DISK_MAX_BYTES`. Ignored when ``disk_path``
+            is ``None``.
+    """
+
+    #: Default directory for the shared cache. Lives under
+    #: :func:`tempfile.gettempdir` so it survives across sessions of the
+    #: same user but doesn't pollute the home directory. The suffix is
+    #: fixed so multiple ``SiftClient`` instances naturally share the
+    #: same store and pick up each other's prior sessions.
+    DEFAULT_DISK_PATH: str = os.path.join(tempfile.gettempdir(), "sift-data-cache")
```
alexluck-sift, line 88: Generating a temp directory each initialization defeats a primary purpose of the disk cache which is cross-session caching. We should choose a directory that is stable

python/lib/sift_client/_internal/cache_namespace.py
```diff
@@ -0,0 +1,114 @@
...
+    Default policy: disk caching is **opt-out** (the ``DiskCacheConfig`` is
+    constructed with ``enabled=True``). Users who don't want any state on
+    disk call :meth:`disable_disk` to silence it; users who want a custom
+    location or byte cap call :meth:`enable_disk` with arguments.
+    """
+
+    def __init__(self, client: SiftClient):
+        self._client = client
+
+    def enable_disk(
```
alexluck-sift: can we remove `_disk` suffix from these? seems unnecessary since there is no alternatives to the disk cache provided

python/lib/sift_client/client.py
```diff
@@ -230,6 +252,52 @@ def rest_client(self) -> RestClient:
...
+            target_path = config.path or DiskCache.DEFAULT_DISK_PATH
+            try:
+                self._disk_cache = DiskCache(
+                    disk_path=target_path,
+                    disk_max_bytes=config.max_bytes,
+                )
+            except Exception:
+                if not config.using_default_path:
+                    raise
+                logger.warning(
```
alexluck-sift: Should we be using a proper warning here instead? If a consumer isn't connecting to our logger, they get no warning.

python/lib/sift_client/_internal/disk_cache.py
```diff
@@ -0,0 +1,336 @@
...
+    Args:
+        disk_path: Directory to persist to. ``None`` keeps the store
+            disabled. A previously-populated directory is reused, so a
+            fresh process reading the same path sees existing entries.
+        disk_max_bytes: Byte cap on the store. ``None`` falls back to
+            :attr:`DEFAULT_DISK_MAX_BYTES`. Ignored when ``disk_path``
+            is ``None``.
+    """
+
+    #: Default directory for the shared cache. Lives under
```
alexluck-sift: what is this format of docstring? It isn't picked up by my IDE

python/lib/sift_client/_internal/low_level_wrappers/data.py
```diff
@@ -40,11 +41,114 @@ class ChannelCacheEntry(BaseModel):
...
+        Type-checks the raw value before returning so a row written by a
+        different adapter (or a corrupt entry that survived) reads as a
+        miss instead of being handed back as the wrong type.
+        """
+        raw = self._store.get(self._key(channel_id))
+        if not isinstance(raw, ChannelCacheEntry):
+            return None
+        return raw
+
+    def put(self, channel_id: str, entry: ChannelCacheEntry) -> None:
```
alexluck-sift: This also needs to be scoped per run since runs act as a label/dimension

python/lib/sift_client/_internal/disk_cache.py
```diff
@@ -0,0 +1,336 @@
...
+        self._disk_path: str | None = None
+        self._disk_max_bytes: int | None = None
+        if disk_path is not None:
+            self._open_disk(
+                str(disk_path),
+                disk_max_bytes if disk_max_bytes is not None else self.DEFAULT_DISK_MAX_BYTES,
+            )
+
+    @classmethod
+    def clear_disk(cls, path: str | os.PathLike[str] | None = None) -> None:
```
alexluck-sift, line 117: This may have issues with deleting open caches. I think we can make this an instance method where it first closes and then clears the cache

python/lib/sift_client/_internal/disk_cache.py
ian-sift: This is stable.  
> As long as the system's environment variables remain the same between restarts, gettempdir() will reliably return the exact same folder path (e.g., /tmp on Linux or C:\Users\Name\AppData\Local\Temp on Windows)

https://docs.python.org/3/library/tempfile.html#tempfile.gettempdir

python/lib/sift_client/_internal/low_level_wrappers/data.py
```diff
@@ -204,24 +322,21 @@ def _update_cache(
                     # So we just don't update the cache.
```
alexluck-sift: Likely an existing bug, but because this function uses channel name (instead of id), bitfield elements aren't going to work correctly. This function should be id only so that that behavior is fixed. `name_id_map` can probably go away.

python/lib/sift_client/_internal/low_level_wrappers/data.py
```diff
@@ -204,24 +322,21 @@ def _update_cache(
...
-                    pd.concat([self.channel_cache.channels[channel_id].data, data])
-                    .groupby(level=0)
-                    .last()
-                )
-                self.channel_cache.channels[channel_id].start_time = min(
-                    suggested_start_time, self.channel_cache.channels[channel_id].start_time
-                )
-                self.channel_cache.channels[channel_id].end_time = max(
-                    end_time, self.channel_cache.channels[channel_id].end_time
+            existing = self.channel_cache.get(channel_id)
```
alexluck-sift: From claude:
Write amplification on incremental pulls. _update_cache re-concats and re-pickles the full merged
frame on every fetch (data.py:325-339). Sequential incremental reads of one long run rewrite the whole           
accumulated frame each time, roughly O(n²). Partly acknowledged by the TODO at data.py:194-195.

python/lib/sift_client/_internal/disk_cache.py
ian-sift: this is a call that users have to call explicitly (we don't automatically call anywhere). It should be something they can call to clear cache accumulated from previous sessions and if they somehow are running parallel processes, the managing of process-unique caches should be managed via the configured path by the user.

python/lib/sift_client/util/cache.py
```diff
@@ -0,0 +1,114 @@
...
+
+if TYPE_CHECKING:
+    import os
+
+    from sift_client.client import SiftClient
+
+logger = logging.getLogger(__name__)
+
+
+class CacheNamespace:
```
alexluck-sift, line 110: Would be nice to have a function that a user could call to see current cache location, size, maybe some other stats about how many channels, etc.?

python/lib/sift_client/_internal/low_level_wrappers/data.py
ian-sift: this is kinda a complicated fix. added a draft at https://github.com/sift-stack/sift/pull/664 if we wanna include that here

python/lib/sift_client/_internal/disk_cache.py
alexluck-sift: This is called by the CacheNamespace which if no path is provided, it will delete the default one which may be open

python/lib/sift_client/util/cache.py
```diff
@@ -0,0 +1,239 @@
...
+        pct = f" ({self.size_bytes / self.max_bytes * 100:.1f}%)" if self.max_bytes else ""
+        return (
+            "Sift cache:\n"
+            f"  path:     {self.path}\n"
+            f"  used:     {_format_bytes(self.size_bytes)} / {cap}{pct}\n"
+            f"  entries:  {self.entry_count} ({self.channel_entries} channel)"
+        )
+
+
+class CacheNamespace:
```
alexluck-sift, line 110: This should be outside of _internal so it is picked up by docs (since we expose it on the SiftClient directly)

python/lib/sift_client/_internal/disk_cache.py
ian-sift: ya my point is I believe that would be the expectation esp if the expectation is for the cache to pick up session to session. If you have multiple processes you can't get away from specifying unique paths per process if you ever wanna call clear cache and only have it affect that process' cache so I don't see how that behavior changes if you make it an instance method

alexluck-sift (commented):
Review of the cache rewrite. The architecture is solid and the tests are thorough; the pagination rewrite looks correct. Comments below focus on the cache write path and the concurrency model. The run-scoped empty-result caching (first comment) is the one I'd resolve before merge.

python/lib/sift_client/_internal/low_level_wrappers/data.py
```diff
@@ -105,122 +587,121 @@ async def _get_data_impl(
...
+                end_time=seg_end,
+            )
+
+        # Empty refs for channels that were queried this call but came
+        # back with zero rows. One ref per fetched range so the
+        # coverage claim matches what we actually asked the wire about
+        # — overclaiming would let a future query for an adjacent
+        # range hit the cache and miss data that actually exists.
+        empty_df = pd.DataFrame()
+        for cid, ranges in fetched_ranges_per_channel.items():
```
alexluck-sift, line 705: Run-scoped empty results get cached here, which can permanently mask data that arrives later on an ongoing run.

The old `_update_cache` deliberately skipped caching empty results when `run_id` was set (the `continue`, with the note that absence before the run started isn't assertable). This empty-ref loop has no `run_id` guard, so a run-scoped query that currently returns no rows writes an empty `SegmentRef` over the fetched window. With default args that window is `[epoch, now()]`.

Repro: `get_channel_data(c1, run_id="R", T0, T2)` before the run has produced `[T1,T2]` data returns empty and records coverage for `[T0,T2]`. After the run ingests that data, the same query becomes a full cache hit (`gaps == []`), returns zero rows, and never refetches.

This also reads as the opposite of the PR description ('Empty results aren't cached ... now re-fetches'). The run-scoped path has no test. Suggest a regression test with `run_id` set, an empty first response and a non-empty second, asserting the second call still hits the wire.

python/lib/sift_client/_internal/low_level_wrappers/data.py
```diff
@@ -35,16 +36,483 @@
...
+        Per-fetch disk write is O(this segment) — no rewrite of any
+        already-cached segment.
+
+        Write order is segment-first, then index: an interrupted update
+        leaves an unreachable orphan segment (harmless; LRU-evicts) but
+        never leaves the index pointing at a missing segment. Empty
+        puts skip the segment write entirely and update the index
+        directly.
+        """
+        idx = self._load_index(channel_id, run_id) or SegmentIndex()
```
alexluck-sift, line 255: The index read-modify-write (`_load_index` then mutate then `_write_index`, also in `_compact_bucket`) is not atomic, and the default cache path is shared across every `SiftClient` in a process and across processes. `diskcache` makes each individual `set` atomic, but nothing wraps the get plus set, and there is no `cache.transact()` or lock.

Concurrent `get_data` calls on the same `(channel, run)` bucket last-writer-wins on the index: writer A's segment body lands but its ref is dropped, leaving an orphan plus a redundant refetch. Both writers can also read the same `next_seg_id`, write the same `seg:<n>` key, and leave the surviving ref (carrying A's claimed range) pointing at B's data. The crash-safety comments cover single-writer interrupt ordering but not concurrent writers. Worth wrapping the index update in `transact()`, or documenting the store as single-writer.

Related design point on cache size: `max_bytes` is currently session-defined (each client sets the cap in-process, default 4 GiB) while the store persists cross-session and cross-process at a fixed path. Two processes opening the same directory with different caps fight over `size_limit` (last opener wins). Since the store is already shared, its budget should be too. Consider a per-directory sidecar (for example `sift-cache-meta.json` holding `max_bytes` and a schema version, read on open and overwritten only by an explicit `enable(max_bytes=...)`), and/or an env var so a deployment can set it once across imports.

python/lib/sift_client/_internal/low_level_wrappers/data.py
```diff
@@ -35,16 +36,483 @@
...
+
+            frame = self._load_segment(channel_id, run_id, ref.seg_id)
+            if frame is None:
+                # Evicted by diskcache LRU (or index/segment got out of
+                # sync). Treat the segment's range as uncovered so the
+                # caller refetches it.
+                continue
+
+            present_ranges.append(clamped)
+            sliced = frame[start_time:end_time]  # type: ignore[misc]
```
alexluck-sift, line 208: This label slice raises `KeyError: Value based partial slicing on non-monotonic DatetimeIndexes ... is not allowed` if a stored segment's `DatetimeIndex` is not monotonic, crashing the read path instead of degrading to a gap. Single-frame segments are stored verbatim with no `sort_index()` (only the bitfield multi-frame branch sorts via `groupby`). The SDK cannot request descending order today (`order_by` is not wired to the request), so this only fires if the server ever returns an unsorted page, but there is no defensive sort or try/except. Cheap guard: sort on store, or wrap the slice and treat a failure as a gap.

python/lib/sift_client/_internal/disk_cache.py
```diff
@@ -0,0 +1,364 @@
...
+
+    def close(self) -> None:
+        """Release the disk file handle. Safe to call when disabled."""
+        self._close_disk()
+
+    def _open_disk(self, path: str, max_bytes: int) -> None:
+        import diskcache
+
+        os.makedirs(path, exist_ok=True)
+        self._disk = diskcache.Cache(
```
alexluck-sift: This opens a SQLite handle (plus a connection pool) that is only closed via an explicit `disable`, `enable`, or `clear`. There is no `__del__`, context manager, or `SiftClient.close()`, so a service that builds many transient clients leaks a handle per client against the shared directory, which compounds the concurrent-writer concern above. Consider closing on client teardown or exposing a `close()`.

python/lib/sift_client/_internal/low_level_wrappers/data.py
```diff
@@ -35,16 +36,483 @@
...
+            before :meth:`_compact_bucket` folds them into one.
+            Without compaction a long incrementally-pulled run grows
+            ``len(idx.segments)`` without bound, so every full-range
+            ``get_range`` would load and concat all segments —
+            swapping the old O(n^2) write for unbounded read fan-out.
+            The cap bounds that fan-out at a known constant.
+
+    """
+
+    KEY_PREFIX: str = "channel:v2:"
```
alexluck-sift: The constant is `channel:v2:`, but the docstring above (around line 97) still calls it the `v1` suffix and gives `channel:v2:` as the example bumped value, so following that instruction would 'bump' to the value already live. The same `v1` wording appears in `test_data.py`. Doc-only, but the versioning story does not match the code.

python/lib/sift_client/_internal/low_level_wrappers/data.py
```diff
@@ -267,51 +781,73 @@ async def get_channel_data(
...
+    def _merge_pages(
+        self,
+        pages: list[list[Any]],
+        *,
+        initial: dict[str, pd.DataFrame],
+    ) -> dict[str, pd.DataFrame]:
+        """Flatten paged channel data + any cached slices into one DataFrame per channel.
+
+        ``initial`` carries any cached slices already populated by
+        ``_check_cache``. Cached entries are folded in as the first frame for
```
alexluck-sift: Stale reference: `_check_cache` is removed in this PR. Cached slices now come from `get_range` inline in `get_channel_data`.

python/lib/sift_client/_internal/disk_cache.py
```diff
@@ -0,0 +1,364 @@
...
+        """
+        self._oversized_warned.clear()
+        if self._disk is not None:
+            self._disk.clear()
+
+    def close(self) -> None:
+        """Release the disk file handle. Safe to call when disabled."""
+        self._close_disk()
+
+    def _open_disk(self, path: str, max_bytes: int) -> None:
```
alexluck-sift: Suggestion: let the size cap persist across sessions instead of re-asserting it per open.

`max_bytes` is currently session-defined (each client passes it, defaulting to 4 GiB) while the store persists at a shared path. Since `_open_disk` passes `size_limit` on every open, an existing cache created at one cap gets silently resized whenever another client opens it with a different one.

`diskcache` already persists `size_limit` in its own Settings table, so we can lean on that: pass `size_limit` only when the caller gives an explicit `max_bytes`, or when the directory is fresh (to establish the 4 GiB default), and omit it otherwise so `diskcache` reuses the stored value. Then read `self._disk.size_limit` back into `_disk_max_bytes` so `stats()` reports the true cap.

Net effect: 4 GiB becomes the default only for fresh caches; an existing cache keeps its cap until an explicit `enable(max_bytes=...)`. This needs threading the optional `max_bytes` through `__init__`/`enable` into `_open_disk` (today it is coalesced to the default too early), a changelog note, and a test that reopens a 2 GiB cache and asserts it stays 2 GiB.

python/lib/sift_client/_internal/low_level_wrappers/data.py
ian-sift: i think this level of concurrency support is blowing up the scope alot. I think nominally processes could just specify their own cache location for now and the concurrency helpers/architecture alluded to here should be scoped to a different ticket

alexluck-sift: agreed

alexluck-sift (dismissed):

alexluck-sift (approved):

PR #651, opened 2026-06-23
rust(feat): update asset tool mcp

Brandon-Shippy:
# Description
adds `update_asset` tool. With this tool you can update metadata and tags on an asset. Will be helpful when building skills for specific workflows, like merging assets.

# Validation
- Unit tests
- E2E testing locally with moving and merging assets


solidiquis (approved):
Can you just fix the branch name to match the convention?

PR #659, opened 2026-06-26
rust(chore): update protos

evan-sift:
Update protos from latest

Protos commit SHA: b3cb55ae03


protos/sift/rule_evaluation/v1/rule_evaluation.proto
```diff
@@ -7,6 +7,7 @@ import "google/api/field_behavior.proto";
 import "google/protobuf/timestamp.proto";
 import "protoc-gen-openapiv2/options/annotations.proto";
 import "sift/common/type/v1/resource_identifier.proto";
+import "sift/reports/v1/reports.proto";
```
wei-qlu: `import "sift/reports/v1/reports.proto";` in `rule_evaluation.proto` has to be removed too since it was used in something that was stripped out

Not really a good way to remove these since there can be cascading dependencies

wei-qlu (approved):

PR #662, opened 2026-06-27
rust(feat): version cli top level command

Brandon-Shippy:
# Description
Adds command `sift-cli version` which hits githubs release tags and tells the user whether there is a new release or if they are on the newest version

Adds new crate, semver, for parsing and versioning semantics

Replaces clap's default version command. 

# Verification
**Latest**
Screenshot 2026-06-29 at 10 52 34 PM
**On alpha build**
Screenshot 2026-06-30 at 1 56 28 PM
**Out of date**
Screenshot 2026-06-30 at 1 58 42 PM


rust/crates/sift_cli/src/cmd/version.rs
```diff
@@ -0,0 +1,78 @@
...
+pub async fn run() -> Result<ExitCode> {
+    let current_str = env!("CARGO_PKG_VERSION");
+    let current = Version::parse(current_str)?;
+
+    let mut out = Output::new();
+    out.line(format!("sift-cli {}", current_str.bold()));
+
+    match fetch_latest().await {
+        Ok(Some(latest)) => {
+            out.line(format!("Latest release: {}", latest.to_string().bold()));
```
solidiquis, line 29: We shouldn't use alpha versions as the latest release. Prefer to locate the last stable version and use that as the latest release.

Furthermore if the user isn't on the latest version it'd be cool if we can grab the curl/powershell command based on their OS from the release pages and drop it for them to use.

solidiquis: You can manually verify by isntalling an old version of the cli

solidiquis (changes requested):

rust/crates/sift_cli/src/cmd/version.rs
Brandon-Shippy: https://github.com/sift-stack/sift/pull/662/commits/1065fa1f2e01c3afc8d689899472b83e63c239b9

solidiquis (approved):

PR #663, opened 2026-06-29
python(feat): ULog data imports

wei-qlu:
### Summary

Adds PX4 ULog (`.ulg`) as a data import format in the Python client. ULog files are self-describing, so import is one call with no schema:

```python
client.data_import.import_from_path("flight.ulg", asset=asset)
```

- Detection runs client-side with pyulog; the import service has no ULog detect endpoint. Requires the new `ulog` extra: `pip install sift-stack-py[ulog]`.
- With no config, every channel in the file is imported. Use `detect_config` to inspect the channels first and drop, rename, or retype them before importing.
- Channels are selected by `message_name`, `instance`, and `field_name`. The derived `channel` property gives the full name, e.g. `"sensor_accel_0.x"`, which is also the default Sift channel name.
- Config also supports `relative_start_time` (anchors the timeline; required when the log has no GPS fix), `info_keys` / `param_keys` (imported as run metadata), and `parse_error_policy`.

### Test

- Unit tests: detection rules on synthetic `.ulg` files (all scalar types, multi-instance topics, logged strings, corrupt records) and config round-trips through the proto.
- Manual QA against dev: imported values and timestamps match pyulog's decode; selective import picks the right instance and applies renames, units, and run metadata; bad selectors, and missing time anchors


python/lib/sift_client/_internal/util/ulog.py
```diff
@@ -0,0 +1,234 @@
...
+class _ScanResult:
+    """Logged topics and log-string markers found while scanning records."""
+
+    def __init__(self) -> None:
+        self.subscriptions: list[tuple[str, int]] = []
+        self.has_untagged_logs = False
+        self.log_tags: set[int] = set()
+
+
+def _scan_subscriptions(path: Path) -> _ScanResult:
```
marc-sift: Per our offline discussion, this manuall parsing isn't needed. Let's just use pyulog to get the subscriptions and add a warning about the pyulog library silently dropping malformed topics.

marc-sift (changes requested):

python/lib/sift_client/_internal/util/ulog.py
wei-qlu: Updated to use `pyulog`

python/lib/sift_client/sift_types/data_import.py
```diff
@@ -841,10 +849,167 @@ def _to_proto(self) -> Hdf5ConfigProto:
...
+            (the ``channel`` property), e.g. ``"sensor_accel_0.x"``.
+    """
+
+    message_name: str
+    instance: int = 0
+    field_name: str = ""
+    name: str = ""
+
+    @property
+    def channel(self) -> str:
```
marc-sift: Should this be called `ulog_name` or `ulog_channel` or something like that to distinguish it from a sift channel?

python/lib/sift_client/sift_types/data_import.py
```diff
@@ -841,10 +849,167 @@ def _to_proto(self) -> Hdf5ConfigProto:
...
+
+class UlogImportConfig(ImportConfigBase):
+    """Configuration for importing a PX4 ULog (``.ulg``) file.
+
+    ULog files describe their own channels. Leave ``data`` empty to import every
+    detected channel, or call ``detect_config`` and edit the returned ``data``
+    list to skip, rename, retype, or annotate channels before importing.
+
+    Attributes:
+        data: Channel selections. Empty imports all detected channels with
```
marc-sift: "If empty, will  import..."

python/lib/sift_client/_internal/util/ulog.py
```diff
@@ -0,0 +1,167 @@
...
+                    for i in range(array_size):
+                        walk(f"{prefix}{field_name}[{i}].", field_type)
+                else:
+                    walk(f"{prefix}{field_name}.", field_type)
+
+    walk("", message_name)
+    return flattened
+
+
+def detect_ulog_fields(ulog: ULog) -> dict[str, DetectedUlogChannel]:
```
marc-sift: Do you need `DetectedUlogChannel`?  There's enough information here to create the `UlogDataColumn` and can just return that.

marc-sift (changes requested):

python/lib/sift_client/sift_types/data_import.py
wei-qlu: renamed to `default_channel_name`, it's the original that was created during parsing

python/lib/sift_client/sift_types/data_import.py
wei-qlu: updated

python/lib/sift_client/_internal/util/ulog.py
wei-qlu: Nope, cleaned up and creates the UlogDataColumn with data type conversion inline

marc-sift (approved):

PR #664, opened 2026-06-29
draft move to segment based cache indexing

ian-sift:


alexluck-sift (commented):
Four notes on the segment cache, inline. The compaction one (on the index write) is the only one I'd block on before this is on by default, since read cost grows without bound on a long run rather than staying a corner case. The key-prefix versioning is a cheap simplification worth folding in now. Write-side dedup and empty-range coverage are real but can be follow-ups, and the empty-range change only earns its keep if no-data queries turn out common.

python/lib/sift_client/_internal/low_level_wrappers/data.py
```diff
@@ -110,77 +123,238 @@ def __init__(self, store: DiskCache):
...
+        idx.segments.append(
+            SegmentRef(
+                seg_id=seg_id,
+                start_time=start_time,
+                end_time=end_time,
+                size_bytes=size_bytes,
+            )
+        )
+        idx.next_seg_id += 1
         self._store.put(
```
alexluck-sift: The segment model fixes write amplification, but nothing ever coalesces segments, so per-bucket count grows without bound. A long run pulled incrementally writes one segment per fetch, and every full-range read then loads and concats all of them. That swaps the old O(n^2) write for unbounded read fan-out, which I'd want bounded before this is on by default.

The fix is a cap plus compaction. When a bucket's segment count crosses a `MAX_SEGMENTS_PER_BUCKET`, load its segments, coalesce overlapping and adjacent ranges, write the merged segment, delete the old keys, and rewrite the index. A flat cap shrinks the constant, but the re-pickle keeps it quadratic. For amortized O(n), trigger compaction geometrically: fold the recent small segments into a base segment once their combined size reaches the base, the way an LSM tree levels. That also holds read fan-out to a few segments. A flat cap is fine for a first cut if we pick a low number and add a segment-count counter, so we can see real fragmentation before tuning.

python/lib/sift_client/_internal/low_level_wrappers/data.py
```diff
@@ -110,77 +123,238 @@ def __init__(self, store: DiskCache):
...
+        entry = ChannelCacheEntry(
+            data=data,
+            start_time=start_time,
+            end_time=end_time,
+            size_bytes=size_bytes,
+        )
+
+        # Segment first.
+        seg_key = self._segment_key(channel_id, run_id, seg_id)
+        self._store.put(seg_key, entry, size_bytes=size_bytes)
```
alexluck-sift: Overlapping segments are real, but I don't think this needs its own mechanism. In steady state `get_range` only fetches gaps, so a refetch of a covered range is rare, and the few that occur get coalesced by the compaction proposed on the index-write thread. If we want a cheap guard anyway, `put_segment` can skip the write when the new `[start, end]` is already contained in an existing `SegmentRef`, which is an index-only check with no segment load. I'd hold even that until compaction is in.

python/lib/sift_client/_internal/low_level_wrappers/data.py
```diff
@@ -110,77 +123,238 @@ def __init__(self, store: DiskCache):
...
+            end_time: Claimed upper bound; same semantics as start.
+
+        Per-fetch disk write is O(this segment) — no rewrite of any
+        already-cached segment.
+
+        Write order is segment-first, then index: an interrupted update
+        leaves an unreachable orphan segment (harmless; LRU-evicts) but
+        never leaves the index pointing at a missing segment.
         """
+        if len(data) == 0:
```
alexluck-sift: Right now an empty fetch is skipped here, so a repeat of a no-data range hits the wire again. Record it instead as a `SegmentRef` with `seg_id = None`, meaning the range is covered but has no backing segment. `get_range` counts the ref's range toward coverage and skips the load, and `_compute_gaps` then drops it from the gaps, so there's no refetch. That keeps one coverage representation (the ref list) rather than a parallel `empty_ranges` field, and the coverage stays in the index we already load, so reads don't pay a segment load per empty range. For run-scoped queries, claim only the exact window you queried, since you can't assert absence outside it. Needs the same version bump as the rest.

python/lib/sift_client/_internal/low_level_wrappers/data.py
```diff
@@ -110,77 +123,238 @@ def __init__(self, store: DiskCache):
...
             if key.startswith(self.KEY_PREFIX):
                 self._store.invalidate(key)
 
+    # --- internal helpers ---
+
+    def _load_index(self, channel_id: str, run_id: str | None) -> SegmentIndex | None:
+        raw = self._store.get(self._index_key(channel_id, run_id))
+        if not isinstance(raw, SegmentIndex):
+            return None
+        if raw.schema_version != self.SCHEMA_VERSION:
```
alexluck-sift: The `schema_version` field only guards the index shape; segments still rely on the `isinstance` check. Fold the version into the key prefix instead, e.g. `channel:v1:<run>:<id>:...`. A bump then changes every key, so old indexes and old segments both go unreachable and LRU-evict together, and the `schema_version` field and this mismatch-discard branch go away. One version, both entry shapes.

PR #666, opened 2026-06-29
python(refactor): group access control APIs under client.access_control

wei-qlu:
## Summary

Groups access-control APIs under:

```python
client.access_control
client.async_.access_control
```

This keeps resource and principal attributes together under one access-control surface, with room for future APIs.

- Added sync and async `access_control` namespaces.
- Moved resource and principal attribute APIs under `client.access_control`.
- Improved assignment ergonomics:
  - `assign()` accepts key objects or key IDs.
  - Resource assignments accept resource objects, raw resource IDs, or `ResourceAttributeEntity`.
  - Principal assignments accept user IDs, and user emails for `USER` principals.

### Example

```python
key = client.access_control.resource_attributes.get_or_create_key(
    "licenses",
    ResourceAttributeKeyType.SET_OF_ENUM,
)
licenses = key.get_or_create_enum_values(["LICENSE_A", "LICENSE_B"])
key.assign_to([channel], value=licenses)
```

## Validation

- Unit tests and Manual QA of creating resource and principal attributes through the new namespace.


python/lib/sift_client/resources/access_control/principal_attributes.py
alexluck-sift: Is this intended to be user set/viewed? If so, should it not be a function arg?

Same with the endpoint stuff below, these should all be in the low level client. They are not concerns of the high level resource.

alexluck-sift: a lot of newly introduced private functions, should these go into _internal?

python/lib/sift_client/resources/access_control/principal_attributes.py
```diff
@@ -412,8 +419,6 @@ async def unarchive_assignments(
                 batch, principal_type=principal_type.value
             )
 
-    # ───────── User resolution ─────────
-
     async def resolve_user_id(self, email: str) -> str:
```
alexluck-sift: This should be a full fledged Users resource and types. We shouldn't implement it within another resource since it is common functionality/behavior that would benefit the entire SDK

python/lib/sift_client/resources/access_control/resource_attributes.py
```diff
@@ -74,7 +87,99 @@ def __init__(self, sift_client: SiftClient):
...
+            )
+
+        return {
+            resource_id: ResourceAttributeEntity(
+                entity_id=resource_id, entity_type=next(iter(entity_types))
+            )
+            for resource_id, entity_types in matched_types.items()
+        }
 
     async def get_key(self, *, key_id: str) -> ResourceAttributeKey:
```
alexluck-sift: The pattern of `get_<thing>`, `list_<thing>`, etc. isn't consistent with the rest of the client which is `resource --> CRUD method`. It would be more inline with the rest of the client if the namespacing was something like `attrtibutes.keys.get()` etc.

See `CONTRIBUTING.md` for the nested resource pattern

python/lib/sift_client/sift_types/principal_attribute.py
alexluck-sift, line 1: principal and resource attributes share the same effective primitives. Rather than defining these twice, we should see if we can have a single set which both principal and resources can use. 

A common `AttributeKey`, `AttributeEnumValue`, `AttributeAssignment` types would simplify things and reduce confusion/user burden.

At the resource level, this should also be done or at the very least leverage a common mixin that captures the common functionality instead of duplication.

Here is a more detailed set of suggestions around this provided by Claude:
1. Unify the two sides behind a shared core. Principal and resource attributes are the same three-tier primitive (Key → EnumValue → Assignment) pointed at different targets, but the code duplicates the machinery twice rather than sharing it. Introduce a generic base (AttributeKey, AttributeEnumValue, AttributeAssignment) with two thin specializations, or at least a shared mixin for the identical logic (_enum_value_id, _chunks, _assignment_id, _resolve_key, the value-mapping helper). This removes the drift risk and the copy-paste.
2. Harmonize the vocabulary across the two halves so knowledge transfers from one to the other:
  - Key value type: value_type (principal) vs key_type (resource). Pick one.
  - Assignment type: PrincipalAttributeValue vs ResourceAttribute. Pick one naming pattern.
  - Value-type enum: PrincipalAttributeValueType vs ResourceAttributeKeyType. Align.
3. Resolve the "assignment" terminology collision. One concept currently has three names: the verb assign, the noun assignment (in method names), and the type ResourceAttribute / PrincipalAttributeValue. list_assignments() returns ResourceAttribute objects. Align the type name and the method vocabulary, or bridge them explicitly in the class docstrings.
4. Decide and document the scope of access_control. The namespace exposes only attribute tagging. The policies (Cedar) and roles/v2 services that turn tags into actual access decisions are not surfaced in the new client. Either state clearly that decisions are configured elsewhere, or plan policies as a sibling under the same namespace so the name matches what the surface does.
5. Smooth principal_type the way resource IDs are already smoothed. The resource side infers a bare ID's type by querying assets/channels/runs (resource_attributes.py _resolve_resource_ids). The principal side instead makes the caller thread principal_type=USER_GROUP through assign, list_assignments, archive, and get_assignment. Infer it, or fold user/group into single calls, so the two halves behave symmetrically.

python/lib/sift_client/resources/access_control/resource_attributes.py
```diff
@@ -354,15 +456,16 @@ async def list_assignments(
...
+                Pass a resource object, resource ID, or ``ResourceAttributeEntity``.
             include_archived: If True, include archived assignments.
             filter_query: Explicit CEL query.
             order_by: Field and direction to order by.
             limit: Maximum number of assignments to return.
             page_size: Results to fetch per request.
         """
-        if entity is not None:
-            resolved = _resolve_entity(entity)
+        if resource is not None:
```
alexluck-sift, line 571: When `resource` is set, `key` and `filter_query` are silently dropped by this early return. Someone who passes a resource and a key will assume both applied and get back a wider set than they expected. Raise a `ValueError` when `resource` is combined with the other filters, since rejecting the call is clearer than ignoring an argument.

python/lib/sift_client/resources/access_control/principal_attributes.py
alexluck-sift, line 1: `get_assignment`: `principal_type` defaults to `USER`, so fetching a user-group assignment by ID returns nothing unless the caller passes `principal_type=PrincipalType.USER_GROUP`. The ID alone identifies the assignment, so a correct call ends up looking like a missing record. Can we infer the type from the ID, or drop the default and require it so a wrong guess fails loudly?

python/lib/sift_client/resources/access_control/resource_attributes.py
alexluck-sift, line 1: `update_key` / `create_key`: the rest of the client updates through typed models like `AssetUpdate` and `ReportUpdate` with field masks, but keys update through loose kwargs here. That reads fine for two fields, but it breaks the pattern and won't carry field-mask semantics as keys grow. The same applies to the principal side. If skipping the model layer is deliberate, worth a note in the PR. A common `AttributeKeyUpdate` would also serve both sides once the primitives are shared.

alexluck-sift, line 1: `assign` (and `assign_to` on the key type): `value` is typed as `Any` and dispatched on the key's type at runtime, so nothing tells a caller whether to pass a bool, an int, a single enum, or a list until they read the docstring or hit the error. Typed overloads keyed on the value type would let the checker catch the common mistakes.

alexluck-sift, line 1: `archive_enum_value`: this returns a migration count while the neighboring methods return other things. `archive_assignments` returns `None`, `archive_key` returns the key, `assign` returns a list, and `check_key_archive_impact` returns an int. A caller can't predict the shape without reading each one. Align where it's reasonable, and where a return is genuinely unusual like this count, call it out in the docstring so it isn't a surprise.

python/lib/sift_client/resources/access_control/__init__.py
alexluck-sift, line 1: The docs preview builds, but there's no worked example past the snippet in the PR description. One end-to-end walkthrough covering create a key, define enum values, and assign to a user and to an asset would cover what most people reach for first. Fine as a follow-up if you'd rather land the refactor first.

python/lib/sift_client/resources/access_control/resource_attributes.py
```diff
@@ -23,26 +25,34 @@
...
-    """Resolve an entity-like value to a ResourceAttributeEntity."""
-    if isinstance(entity, ResourceAttributeEntity):
-        return entity
-    if isinstance(entity, Asset):
-        return ResourceAttributeEntity.for_asset(entity._id_or_error)
-    if isinstance(entity, Channel):
-        return ResourceAttributeEntity.for_channel(entity._id_or_error)
-    if isinstance(entity, Run):
-        return ResourceAttributeEntity.for_run(entity._id_or_error)
+ResourceLike = Union[ResourceAttributeEntity, Asset, Channel, Run, str]
```
alexluck-sift: **Make the caller's intent explicit**

A bare `str` in `resources` doesn't say what it refers to, so the client has to guess the type. `_resolve_resource_ids` runs that guess for you: three `list_()` calls against assets, channels, and runs per batch to work out what the ID is. Guessing intent is the wrong job for the client. We should only accept an input when the type is unambiguous: an object, which says its own type, or a raw ID paired with the type the caller meant.

The resolution being slow and quietly fragile (channels aren't fetched with `include_archived`, so an archived channel ID looks like a bad ID; the ambiguity check can't even fire if IDs don't collide across types) is a symptom of guessing, not the main reason to drop it.

Fix: drop bare `str` and let the reference name its own type.

```python
ResourceLike = ResourceAttributeEntity | Asset | Channel | Run   # no bare str
```

Objects already carry their type. Raw IDs go through the factories we already have:

```python
key.assign_to([channel, Resource.for_asset("ast_1")], value=licenses)
```

`assign` then resolves nothing over the network:

```python
entities = [_as_entity(r) for r in resources]  # pure, reads each object's own type
```

`_resolve_resource_ids` and the three list calls go away. Same move on the principal side kills the `principal_type` argument, since `PrincipalRef.user(...)` / `PrincipalRef.user_group(...)` each know what they are.

python/lib/sift_client/sift_types/resource_attribute.py
```diff
@@ -145,7 +150,7 @@ def __str__(self) -> str:
 
 
 class ResourceAttribute(BaseType[ra_pb.ResourceAttribute, "ResourceAttribute"]):
```
alexluck-sift: Two things about how this type surfaces its data.

The value is spread across `enum_value_id`, `boolean_value`, and `number_value`, so a caller checks three fields to find the one that's set. A single `value` property that returns the populated one (typed off the key) would match how the rest of the client hides proto shapes. Keep the raw fields if you want, but give people one obvious accessor.

`key`, `enum_value`, and `entity` are optional and populated only when the proto carried the details, so `attr.key.display_name` works on some results and hits `None` on others. Either always resolve them or document exactly when each is set. Same note on `PrincipalAttributeValue`.

python/lib/sift_client/sift_types/resource_attribute.py
```diff
@@ -274,69 +283,75 @@ def create_enum_value(
...
+    ) -> list[ResourceAttribute]:
+        """Assign a value to one or more resources for this key.
 
         Args:
-            entities: Entities to assign to. Each may be a ``ResourceAttributeEntity``
-                or an ``Asset``/``Channel``/``Run`` instance.
+            resources: Resources to assign to. For currently supported resource types, pass
+                ``Asset``, ``Channel``, or ``Run`` objects, their IDs, or
+                ``ResourceAttributeEntity`` when you already know the resource type.
             value: The value to assign. For ``SET_OF_ENUM`` keys, a list of enum values
```
alexluck-sift, line 340: Worth stating how `SET_OF_ENUM` lands in the returned data. Each `ResourceAttribute` holds one enum value, so a set of three on one resource comes back as three assignments, and `assign` returns the list it created. That mapping is fine, but nothing in the type says a resource's set is spread across rows, so `list_assignments` reads like duplicates. Document it, or add a grouped view keyed by resource. Same for principal `SET_OF_ENUM`.

python/lib/sift_client/sift_types/resource_attribute.py
```diff
@@ -37,7 +40,7 @@ class ResourceAttributeKeyType(Enum):
 
 
 class ResourceAttributeEntityType(Enum):
-    """Kind of Sift entity a resource attribute can be assigned to."""
+    """Kind of Sift resource a resource attribute can be assigned to."""
 
     UNSPECIFIED = ra_pb.RESOURCE_ATTRIBUTE_ENTITY_TYPE_UNSPECIFIED
```
alexluck-sift: These enums expose the proto `UNSPECIFIED` zero value, which no caller should pass. Reject it in `create_key` / `assign` (and the principal equivalents), or leave it out of the public enum, so it can't be set by accident.

python/lib/sift_client/resources/access_control/resource_attributes.py
alexluck-sift, line 1: `list_keys` filters on name only, and `list_assignments` on key or resource only, but the types carry more. Keys have `created_date`, `modified_date`, `created_by_user_id`, and `description`; assignments have `created_date` and `created_by_user_id`. Every other `list_` in the client exposes `created_after` / `created_before`, `modified_after` / `modified_before`, `created_by` / `modified_by`, and `description_contains` through the shared `_build_time_cel_filters` and `_build_common_cel_filters` helpers. So `assets.list_(created_after=..., created_by=...)` works, but the same call here doesn't, even though the data is on the type. Same gap on both principal list methods. If the filter backend doesn't support these yet, CONTRIBUTING says to extend the API rather than drop them on the client side.

alexluck-sift, line 1: Small polish: docstring depth is uneven compared to the rest of the client. `list_keys` has full Args/Returns, but `get_key`, `find_key`, `create_enum_value`, and the archive methods are one-liners. Sibling resources like `assets` document every method. Worth leveling these up while the surface is still being shaped. Applies to the principal file too.

python/lib/sift_client/resources/access_control/resource_attributes.py
alexluck-sift: Related: whatever naming we land on, `list_` should keep the trailing underscore the rest of the client uses to avoid shadowing the builtin. Today it's `list_keys` / `list_enum_values` / `list_assignments`. Under the nested shape these become `keys.list_()` and so on, which resolves it cleanly.

alexluck-sift (changes requested):

python/lib/sift_client/resources/access_control/principal_attributes.py
wei-qlu: No, they are internal limits and now moved to the low level client

wei-qlu: moved to _internal and merged duplicate helpers into a shared private module

python/lib/sift_client/resources/access_control/principal_attributes.py
wei-qlu: Added `client.users` with get, list_, find, and resolve_ids. Principal attributes now use it for email lookups

python/lib/sift_client/resources/access_control/resource_attributes.py
wei-qlu: key updates now go through `ResourceAttributeKeyUpdate` and `PrincipalAttributeKeyUpdate`

wei-qlu: Kept the count since it matches `check_key_archive_impact`, and noted it on the returns docstring on both sides.

wei-qlu: `value` is now typed as `ResourceAttributeValueLike` / `PrincipalAttributeValueLike` on `assign` and `assign_to`

python/lib/sift_client/resources/access_control/resource_attributes.py
wei-qlu: Restructured, both attributes now expose `keys`, `enum_values`, and `assignments` as nested sub-resources with CRUD methods

python/lib/sift_client/resources/access_control/resource_attributes.py
wei-qlu: Updated, raises an error when `resource` is set with the other args

python/lib/sift_client/resources/access_control/principal_attributes.py
wei-qlu: Updated `principal_type` to be a required arg on `assignments.get`, `archive`, and `unarchive`, ID doesn't encode the type so we can't infer

python/lib/sift_client/resources/access_control/resource_attributes.py
wei-qlu: updated, every method on both files now include the full arg/return docstrings

python/lib/sift_client/sift_types/resource_attribute.py
wei-qlu: Removed `UNSPECIFIED` from the public enums on both attributes

python/lib/sift_client/resources/access_control/__init__.py
wei-qlu: Created a [ticket](https://linear.app/siftstack/issue/ENG-12839/add-end-to-end-access-control-walkthrough-runnable-example-docs-page) for a follow-up

python/lib/sift_client/resources/access_control/resource_attributes.py
wei-qlu: Updated both sides. 

Resources: bare IDs are gone, pass objects or ResourceAttributeEntity.for_*(...).` resolve_resource_ids` is removed. 

Principals: added PrincipalRef.user(...) / PrincipalRef.user_group(...), which removes principal_type from assign; mixed batches split into one RPC per type, and emails still work since an email names its own type.

python/lib/sift_client/sift_types/resource_attribute.py
wei-qlu: Added docs to both

python/lib/sift_client/sift_types/resource_attribute.py
wei-qlu: Documented the row-per-value mapping on the assignment types

python/lib/sift_client/resources/access_control/resource_attributes.py
wei-qlu: Added the standard filters to all six list methods, scoped to what each service's CEL actually supports. Made tickets for remaining server gaps: resource keys lack modified/user filters and resource enum values only filter on name and created date. 

Will update as they are implemented

python/lib/sift_client/sift_types/principal_attribute.py
wei-qlu: vocabulary aligned, naming collision gone, shared logic in one internal module, both sides symmetric with self-typed references. Scope note added to the namespace docstring

Kept two type sets instead of a common AttributeKey, etc. on purpose since the protos diverge, and distinct types catch cross-side mixups statically.

python/lib/sift_client/resources/users.py
```diff
@@ -0,0 +1,125 @@
...
+        with no matching user are omitted.
+
+        Args:
+            emails: The login emails to resolve.
+        """
+        wanted = list(dict.fromkeys(email for email in emails if email))
+        if not wanted:
+            return {}
+        users = await self.list_(names=wanted)
+        return {user.name: user._id_or_error for user in users}
```
alexluck-sift: `resolve_ids` keys the map by the server's canonical `user.name`, but `_resolve_email_refs` looks it up by the caller's original string. If the backend normalizes email case, `assign(key, ["Alex@Sift.com"])` finds the user in the list call and then misses on the exact-key lookup, raising "No user found" for a user that exists. Suggest matching case-insensitively (normalize both sides before the lookup). Can you confirm whether the backend canonicalizes email case?

python/lib/sift_client/resources/users.py
```diff
@@ -0,0 +1,125 @@
...
+        Returns a mapping of email to user ID for the emails that were found. Emails
+        with no matching user are omitted.
+
+        Args:
+            emails: The login emails to resolve.
+        """
+        wanted = list(dict.fromkeys(email for email in emails if email))
+        if not wanted:
+            return {}
+        users = await self.list_(names=wanted)
```
alexluck-sift: `resolve_ids` lists with the default `include_inactive=False`, so a principal attribute can't be assigned to a deactivated user by email even though the user exists. If that's intended, worth a docstring note; otherwise pass `include_inactive=True` for resolution.

python/CHANGELOG.md
```diff
@@ -63,29 +63,45 @@ Breaking change: `client.reports.create_from_template` now takes `report_templat
...
+key = client.access_control.principal_attributes.keys.get_or_create(
+    "licenses",
+    PrincipalAttributeValueType.SET_OF_ENUM,
+)
 licenses = key.get_or_create_enum_values(["LICENSE_A"])
 key.assign_to(["user@example.com"], value=licenses)
 ```
 
-Keys, enum values, and assignments each support create, get, list, update, and archive operations. For `SET_OF_ENUM` keys, an assignment replaces the full value set on each target.
+Keys, enum values, and assignments are managed through the nested `keys`, `enum_values`, and `assignments` APIs on each side (for example `client.access_control.resource_attributes.assignments.list_()`), each supporting the standard create, get, list, update, and archive operations. For `SET_OF_ENUM` keys, an assignment replaces the full value set on each target, and each enum value in the set is returned as its own assignment.
```
alexluck-sift: Non-blocking: this says each sub-resource supports "create, get, list, update, and archive," but `enum_values` has no `get`/`update` and `assignments` has no `update`. Suggest "the relevant create/get/list/update/archive operations," or spelling it out per sub-resource.

python/lib/sift_client/sift_types/principal_attribute.py
```diff
@@ -118,7 +154,9 @@ class PrincipalAttributeValue(BaseType[pa_pb.PrincipalAttributeValue, "Principal
     boolean_value: bool | None
     number_value: int | None
     key: PrincipalAttributeKey | None
+    """Full key details. Always set in responses."""
```
alexluck-sift: Non-blocking: this documents `key` as "Always set in responses," but the resource-side equivalent says "Only set when the server includes them," and both populate conditionally on `HasField("key")`. Please align this with the resource wording.

python/lib/sift_client/resources/access_control/principal_attributes.py
```diff
@@ -0,0 +1,714 @@
...
+        """
+        resolved_key = await resolve_key(
+            key,
+            key_cls=PrincipalAttributeKey,
+            getter=lambda key_id: self._low_level_client.get_key(key_id),
+        )
+        refs = await self._resolve_email_refs([_as_principal_ref(p) for p in principals])
+        create_kwargs = attribute_value_kwargs(resolved_key.value_type, value)
+
+        grouped: dict[PrincipalType, list[str]] = {}
```
alexluck-sift, line 544: Non-blocking: `create` groups by principal type and returns assignments in group order, not input order, so a caller pairing input to output positionally would mispair. Either preserve input order or note in the docstring that order isn't guaranteed.

python/lib/sift_client/resources/access_control/_common.py
```diff
@@ -0,0 +1,62 @@
...
+    return await getter(key)
+
+
+def attribute_value_kwargs(value_type: Enum, value: Any) -> dict[str, Any]:
+    """Map a user-supplied value to the batch-create value kwargs for a key's value type.
+
+    Dispatches on the enum member name, which is identical for the resource and
+    principal value-type enums (ENUM, BOOLEAN, NUMBER, SET_OF_ENUM).
+    """
+    if value_type.name == "SET_OF_ENUM":
```
alexluck-sift, line 44: Non-blocking: an empty list passes the `SET_OF_ENUM` check and sends an RPC that clears the set on each target. If that's the intended "clear" path, worth documenting; otherwise reject an empty list.

python/lib/sift_client/resources/access_control/principal_attributes.py
```diff
@@ -0,0 +1,714 @@
...
+        """Batch unarchive assignments of the given principal type.
+
+        Args:
+            assignments: The assignments or assignment IDs to unarchive.
+            principal_type: The kind of principal the assignments apply to.
+        """
+        ids = [id_of(a) for a in assignments]
+        await self._low_level_client.unarchive_values(ids, principal_type=principal_type.value)
+
+    async def _resolve_email_refs(self, refs: list[PrincipalRef]) -> list[PrincipalRef]:
```
alexluck-sift, line 693: Non-blocking: duplicate principals aren't de-duped before the batch call (`resolve_ids` dedups, but this re-expands one resolved ref per input), so repeated emails/refs send duplicate IDs to `batch_create_values`. Worth de-duping the resolved list.

python/lib/sift_client/sift_types/principal_attribute.py
```diff
@@ -243,76 +294,96 @@ def create_enum_value(
...
+        principals: list[PrincipalRef | User | str],
+        *,
+        value: PrincipalAttributeValueLike,
+    ) -> list[PrincipalAttributeAssignment]:
         """Assign a value to one or more principals for this key.
 
         Args:
-            principals: Principal IDs to assign to. For ``USER`` principals, an entry
-                containing ``@`` is treated as an email and resolved to a user ID.
+            principals: Principals to assign to. Pass ``PrincipalRef.user(...)`` /
```
alexluck-sift, line 324: Non-blocking: `assign_to` lists the accepted inputs but omits the "bare IDs are rejected" caveat that `assignments.create` documents. Minor consistency fix so the two docstrings agree.

alexluck-sift (changes requested):
Re-reviewed after the latest round. All my earlier feedback is addressed, and I ran a full audit of the current surface: the touched-area unit tests (151) pass, and mypy and pyright are clean. Sync/async parity, exports, and client wiring all check out. Overall the PR is in strong shape.

Two items with real user impact are inline as requested changes (both small). The rest are non-blocking and prefixed as such.

One extra note, non-blocking, on test coverage: the suite passes but a few branches are unexercised. A NUMBER key given a `bool`, a `SET_OF_ENUM` key given a scalar, the falsy numeric value `0` (only `5` is tested), the `modified_after` filter, and an assertion that `UNSPECIFIED` stays out of the public enums. Cheap to add and they guard real edges.

python/lib/sift_client/resources/users.py
wei-qlu: Confirmed the backend does not normalize email case, so I made resolve_ids match case-insensitively: exact match first, then a case-insensitive fallback, and it raises if two users differ only by case.

python/lib/sift_client/resources/users.py
wei-qlu: not intended: resolve_ids now includes inactive users, so assigning to a deactivated user by email works, updated the docstring

python/CHANGELOG.md
wei-qlu: Reworded to "the relevant create, get, list, update, and archive operations"

python/lib/sift_client/sift_types/principal_attribute.py
wei-qlu: Aligned with the resource wording

python/lib/sift_client/resources/access_control/principal_attributes.py
wei-qlu: Documented, order isn't guaranteed to match input order

python/lib/sift_client/resources/access_control/_common.py
wei-qlu: Now rejects an empty list, clearing a set is done by archiving its assignments

python/lib/sift_client/resources/access_control/principal_attributes.py
wei-qlu: De-duped after resolution now on both sides

python/lib/sift_client/sift_types/principal_attribute.py
wei-qlu: Added the same bare-ID caveat to `assign_to`

alexluck-sift (approved):

PR #669, opened 2026-07-01
rust(feat): soft approval gates on mcp calls

Brandon-Shippy:
# Description
Uses the destructive flag on tools that are destructive actions. This flag is read by the harness and makes sure to ask the user if they want to make the tool call, assuming the client honors the hints. If the user has YOLO mode on, or has already selected "accept all tool calls like this" options when using the mcp this prompt will be skipped. 

This acts as a soft approval wall, which the user can bypass if they decide to. The way these prompts are shown to the user depends on the harness they are using, and may look different across different agents.

Note: This is closer to hints as outlined in the [MCP tool spec](https://modelcontextprotocol.io/specification/2025-11-25/server/tools)

# Validation
Soft confirmation on `list_assets`
Screenshot 2026-07-01 at 11 48 41 AM
Tool confirmation on imports
Screenshot 2026-07-01 at 12 10 27 PM
Annotation confirmation
Screenshot 2026-07-01 at 12 30 14 PM
Updating tags on an asset
Screenshot 2026-07-01 at 12 33 40 PM

### DIFF
--- a/rust/crates/sift_mcp/CLAUDE.md
+++ b/rust/crates/sift_mcp/CLAUDE.md
@@ -46,7 +46,8 @@ Non-negotiables. These hold regardless of context pressure:
 3. Do not write retry logic anywhere except `policy::with_retry`. Call it; don't reinvent it.
 4. Set `read_only_hint` honestly. Read tools are `true`; anything that writes is `false`. Write
    tools must also set `destructive_hint` and `idempotent_hint` — see Step 4 for the mapping
-   (additive vs. update vs. state-flip).
+   (additive vs. update vs. state-flip). These fields follow the MCP tool annotations spec:
+   <https://modelcontextprotocol.io/specification/2025-11-25/server/tools>.
 5. Do not mirror the API one-to-one. Run Step 0 before writing anything.
 
 ---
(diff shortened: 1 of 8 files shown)


rust/crates/sift_mcp/CLAUDE.md
```diff
@@ -44,7 +44,9 @@ Non-negotiables. These hold regardless of context pressure:
    objects or enums. (See "Keep parameters flat" below for why.)
 2. Every service function gets a test. Testing tools and other code is encouraged.
 3. Do not write retry logic anywhere except `policy::with_retry`. Call it; don't reinvent it.
-4. Set `read_only_hint` honestly. Read tools are `true`; anything that writes is `false`.
+4. Set `read_only_hint` honestly. Read tools are `true`; anything that writes is `false`. Write
```
evan-sift, line 47: perhaps worth include the MCP tool spec link here as well

evan-sift (dismissed):

solidiquis (dismissed):

rust/crates/sift_mcp/CLAUDE.md
Brandon-Shippy: https://github.com/sift-stack/sift/pull/669/commits/20aae1ab0dae6629446681e8110fc8722f865fb0

solidiquis (approved):

PR #674, opened 2026-07-02
rust(feat): ULog data imports

wei-qlu:
## Summary
Adds `sift-cli import ulog`. 

ULog files upload through `CreateDataImportFromUpload` with `ulog_config`; `--preview` runs client-side channel detection using a hand-rolled, dependency-free header parser whose channel naming, type mapping, and edge-case behavior match the server importer and the Python client detector. Flags cover timestamp anchoring (`--relative-start-time`), run metadata (`--info-key`, `--param-key`), and recoverable-error handling (`--parse-error-policy`).

## Test plan
- Unit tests covering the config builder, type map, field flattening, skip rules, sync recovery, and malformed-input tolerance.
- Manual QA: Preview, imports, client-side validation, and parse error policy.


rust/crates/sift_cli/src/cmd/import/ulog/detect_ulog_config.rs
```diff
@@ -0,0 +1,292 @@
+//! Client-side PX4 ULog (`.ulg`) channel detection for `--preview`.
```
Brandon-Shippy: Nit: think we can drop all comments in here for consistency

rust/crates/sift_cli/src/cmd/import/ulog/import.rs
```diff
@@ -0,0 +1,130 @@
...
+    let relative_start_time = match &args.relative_start_time {
+        Some(start) => {
+            let dt = DateTime::parse_from_rfc3339(start)
+                .context("--relative-start-time is not valid RFC3339")?;
+            Some(Timestamp::from(dt.to_utc()))
+        }
+        None => None,
+    };
+
+    // Send only one run identifier; --run-id takes precedence over --run.
```
Brandon-Shippy: How is the ULog backend handling the import?

On other imports we can pass both and the backend will prefer `run_id` over `run_name`. 
If ULog matches that contract we can just set `run_name: args.common.run.clone().unwrap_or_default(),` like we do for other imports. As a result dropping the need for this check

rust/crates/sift_cli/src/cmd/import/ulog/import.rs
```diff
@@ -0,0 +1,130 @@
...
+
+    Ok(UlogConfig {
+        asset_name: args.common.asset.clone(),
+        run_name,
+        run_id: args.common.run_id.clone().unwrap_or_default(),
+        relative_start_time,
+        info_keys: args.info_key.clone(),
+        param_keys: args.param_key.clone(),
+        parse_error_policy: ProtoUlogParseErrorPolicy::from(args.parse_error_policy).into(),
+        // Empty `data` tells the server to import every detected channel.
```
Brandon-Shippy: Nit: think we can drop this comment for consistency with other imports

Brandon-Shippy (changes requested):

rust/crates/sift_cli/src/cmd/import/ulog/detect_ulog_config.rs
wei-qlu: removed

rust/crates/sift_cli/src/cmd/import/ulog/import.rs
wei-qlu: removed

rust/crates/sift_cli/src/cmd/import/ulog/import.rs
wei-qlu: `run_id` and `run_name` are mutually exclusive (enforced server-side) and should be handled client-side, I have the check at the command level. This now aligns with the client as well

I need to update the protos

Brandon-Shippy (approved):

PR #678, opened 2026-07-10
Python(fix): fix get data pagaination and add progress bars

alexluck-sift:
https://github.com/user-attachments/assets/cc73b200-53ca-4914-a995-e210bc893293

## Summary
Adds progress bars to `get_data` and archive unzip, and fixes a pagination defect in `get_data` where the `limit` argument was silently ignored.

## Progress bars
`get_data` shows a two-line bar: the top line fills by channel (`3/8`), and the second line tracks points loaded and throughput live as each page arrives, e.g. `1,024,000 pts · 218,400 pts/s · voltage, current (+3 more)`.
On completion the second line becomes a past-tense summary, e.g. `Loaded 716,834 points (avg 211,921/s)`.
Point totals include cache hits, while the rate covers only fetched points, so a warm call reads `Loaded 716,834 points from cache` with no rate and a mixed call reads `Loaded N points: F fetched (avg …/s), C cached`.
Archive extraction shows unzip progress. Both no-op when stdout cannot back a bar (e.g. a detached process or `pythonw`), default on for sync callers and off for async, and are overridable per call or globally via `sift_client.config.show_progress`.

## get_data pagination fix
While adding the bar I found `get_data` paging one proto at a time and ignoring `limit`. `limit` is documented as a maximum data-point count, but it was passed to the shared paginator as a maximum number of returned protos. Each page returns exactly one proto holding up to `page_size` points, so the loop counted protos, never bounded points, and paged to the end of the run regardless of `limit`. The fix routes the point budget through `page_size` and stops after `ceil(limit / page_size)` pages.

| Request | Wire calls | Max points/channel | Wall time |
|---|---|---|---|
| `limit=1000`, before | ~800 | 1,000,000 (limit ignored) | tens of seconds |
| `limit=1000`, after | 8 | 1,000 | 0.40s |
| `limit=100`, after | 8 | 100 | 0.43s |
| `limit=50000`, after | 8 | 50,000 | 1.98s |
| no limit (full export) | 1,108 | 1,000,000 | ~145s, unchanged |

`limit` now bounds points per channel and each channel resolves in a single round-trip. The unbounded export path still pages to exhaustion via the token.

## Deserialize once
To count points for the progress line without parsing the wire payload twice, each page is deserialized once in the paginator and handed to `_merge_pages` as frames, rather than parsed once to count and again to merge.

## Tests
67 data tests pass. New: 5 for the limit path (single-page budget, ceil pages, no-limit exhaustion, over-budget termination, zero budget) and 3 for the progress line (live points total, fetched/cached split, all-cached shows no rate). The two core limit tests fail against the old proto-counting logic and pass against the fix. The `_merge_pages` tests were updated to pass pre-deserialized frames for the single-parse change.

### DIFF
--- a/python/lib/sift_client/_internal/low_level_wrappers/data.py
+++ b/python/lib/sift_client/_internal/low_level_wrappers/data.py
@@ -824,12 +824,31 @@ class DataLowLevelClient(LowLevelClientBase, WithGrpcClient):
             title="Fetching channel data",
             disable=not show_progress,
         ) as bar:
+            # Fetches run concurrently, so there is no single "current"
+            # channel. Track the set in flight and name them as a group,
+            # dropping each as it returns.
+            in_flight: list[str] = []
+
+            def _render() -> None:
+                if not in_flight:
+                    bar.text("")
+                    return
+                shown = in_flight[:3]
+                extra = len(in_flight) - len(shown)
+                text = ", ".join(shown)
+                if extra:
+                    text += f" (+{extra} more)"
+                bar.text(text)
 
             async def _tick(task: asyncio.Task, label: str) -> Any:
-                result = await task
-                bar.text(label)
-                bar()
-                return result
+                in_flight.append(label)
+                _render()
+                try:
+                    return await task
+                finally:
+                    in_flight.remove(label)
+                    _render()
+                    bar()
 
             # Keep ``gather`` (not ``as_completed``) so pages stay in
             # submission order: ``_merge_pages`` lets later-positioned
--- a/python/lib/sift_client/_internal/low_level_wrappers/data.py
+++ b/python/lib/sift_client/_internal/low_level_wrappers/data.py
@@ -599,25 +599,12 @@ class DataLowLevelClient(LowLevelClientBase, WithGrpcClient):
         max_points: int | None,
         on_page: Callable[[int], None] | None = None,
     ) -> list[dict[str, pd.DataFrame]]:
-        """Page one ``GetData`` query, bounding the result by *points*.
+        """Page one ``GetData`` query up to a point budget.
 
-        ``GetData`` returns one ``ChannelData`` proto per page holding up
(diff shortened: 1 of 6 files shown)


python/lib/sift_client/_internal/low_level_wrappers/data.py
```diff
@@ -589,6 +590,55 @@ async def _get_data_impl(
...
+    ) -> list[Any]:
+        """Page one ``GetData`` query, bounding the result by *points*.
+
+        ``GetData`` returns one ``ChannelData`` proto per page holding up
+        to ``page_size`` points, and keeps issuing a ``next_page_token``
+        while more data exists. ``max_points`` (the caller's ``limit``) is
+        a data-point budget, but a page is one proto regardless of how
+        many points it carries, so the budget drives ``page_size`` and
+        the number of pages, not a proto count. Sizing each page to the
+        budget means it is met in a single round-trip.
```
ian-sift: nit: this comment is kinda wordy confusing

python/lib/sift_client/_internal/low_level_wrappers/data.py
```diff
@@ -589,6 +590,55 @@ async def _get_data_impl(
...
+
+        Each page advances every channel in ``kwargs["channel_ids"]`` by
+        the same ``page_size`` points, so the page count covers the
```
ian-sift: maybe also a misread on my part but just confirming calling next page on a query where some channels stop having data over the same time range is fine?

i.e. page 1 has 10 points for channel1 and 5 for channel2, page 2 has 10 more points for channel1 and none for channel2, page 3 has some points for channel1 and none for channel2 etc

alexluck-sift: right now each getData request only contains one channel because of an existing endpoint bug. But we should definitely revisit once we change how we call the backend

ian-sift (approved):

PR #679, opened 2026-07-10
Replace bash dev script with justfile for task automation

alexluck-sift:
## Summary
Migrate from a custom bash script (`scripts/dev`) to a `justfile` for development task automation. This provides a more maintainable and discoverable way to run common development tasks.

## Key Changes
- **Removed** `python/scripts/dev` bash script (251 lines) containing all development task definitions
- **Added** `python/justfile` with equivalent recipes for all development tasks, including:
  - Environment setup: `bootstrap`, `pip-install`
  - Code quality: `fmt`, `fmt-check`, `lint`, `lint-fix`, `mypy`, `mypy-win`, `pyright`, `check`
  - Testing: `test`, `test-floor`, `test-ceiling`, `test-pytest-compat`, `test-integration`, `test-all`
  - Code generation: `gen-stubs`, `mypy-stubs`, `gen-extras`
  - Documentation: `docs`, `docs-build`
- **Updated** CI workflow (`.github/workflows/python_ci.yaml`) to use `just` commands instead of `uv run` directly
- **Updated** pre-push git hooks to invoke `just` recipes instead of `bash ./scripts/dev`
- **Updated** documentation references in test README to point to `just` commands

## Notable Implementation Details
- Added `just` setup step in CI workflow via `extractions/setup-just@v4`
- Preserved all original functionality and command behavior
- Added `fmt-check` recipe for CI-specific format checking (non-destructive)
- Maintained complex multi-step recipes like `test-pytest-compat` using bash script blocks
- Improved discoverability with `default` recipe that lists all available tasks
- Simplified CI commands by delegating to `just` recipes rather than raw `uv run` calls


.pre-commit-config.yaml
```diff
@@ -0,0 +1,106 @@
...
+        pass_filenames: false
+
+      # --- push stage (thorough) ---
+      - id: python-check
+        name: just python::check (lint + types)
+        entry: just python::check
+        language: system
+        files: ^python/
+        pass_filenames: false
+        always_run: true
```
wei-qlu: should `always_run` be false so it only runs for python changes?

alexluck-sift: Good catch — `always_run: true` was overriding the `files: ^python/` filter, so this ran on every push regardless. Removed it in 9178818; the filter alone now gates it, matching the other push-stage hooks.

---
_Generated by [Claude Code](https://claude.ai/code)_

python/justfile
```diff
@@ -0,0 +1,159 @@
...
+    # stubs (e.g. fcntl) as if we were on Windows. Catches imports that would
+    # only fail at runtime on Windows.
+    uv run mypy --platform=win32 lib
+
+# Run 'pyright lib' for type checking
+pyright:
+    uv run pyright lib
+
+# Run lint-fix, fmt, mypy, mypy-win, pyright sequentially
+check: lint-fix fmt mypy mypy-win pyright
```
wei-qlu: Non-blocking: `just check` assumes the dev-all extra is already synced, so a plain uv sync (which prunes extras) will make it and the pre-push hook fail with: 

Image


The old `fmt-lint.sh` self-synced to avoid this, and `stubs-check`/`extras-check` already do `uv sync --extra dev-all`, so adding it to check would keep them consistent

A normal `just bootstrap` workflow will cover this but just in case

alexluck-sift: Fixed in 8913b01. Added a `_sync-dev` step (`uv sync --extra dev-all --quiet`) as `check`'s first dependency, matching what `stubs-check`/`extras-check` already do. Verified from a pruned env (`uv sync` then `just python::check`): it re-syncs and the checks run clean instead of failing with command-not-found.

---
_Generated by [Claude Code](https://claude.ai/code)_

wei-qlu (dismissed):

wei-qlu (approved):

PR #683, opened 2026-07-14
python(feat): add public sift.canvas.v1 proto package

ian-sift:
## Summary

- New `protos/sift/canvas/v1/canvas.proto` hosting the `CanvasCellExecutionStatus` enum.

## Test plan

- [x] `scripts/gen.sh python` regenerates cleanly.
- [x] `python -c "from sift.canvas.v1 import canvas_pb2; print(list(canvas_pb2.CanvasCellExecutionStatus.items()))"` prints all 10 enum values with their stable integer identifiers.
- [x] `uv lock` refreshes without drift.
- [ ] CI passes.


python/pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "sift_stack_py"
-version = "0.19.0"
+version = "0.19.1"
```
alexluck-sift: do we need to publish this immediately? otherwise we should hold off on versioning (and this isn't a bug fix so patch might be inappropriate)

protos/sift/canvas/v1/canvas.proto
```diff
@@ -0,0 +1,19 @@
+syntax = "proto3";
+
+package sift.canvas.v1;
```
alexluck-sift, line 3: please build all of the gen code for Go, Rust, etc.

python/CHANGELOG.md
```diff
@@ -5,6 +5,16 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [v0.20.0] - July 14, 2026
+
+### What's New
+
+#### Public `sift.canvas.v1` package
```
alexluck-sift: don't think we need to mention added protos here

python/pyproject.toml
ian-sift: i guess we can circle back in azimuth whenever next release happens

alexluck-sift (approved):

PR #686, opened 2026-07-16
rust(feat): mcp gate destructive flag

Brandon-Shippy:
# Description
Require a `--allow-destructive` flag on mcp launch to enable all destructive actions. Adds a method that checks if the flag is turned on, otherwise it gives a useful Error. Also updates developer agent files to ensure when new tools are built the agents have the context to use this method.

Full command is `sift-cli mcp --allow-destructive`

# Verification

Updating Tags w/o flag enabled
Screenshot 2026-07-15 at 11 32 35 PM

Updating Tags with flag enabled
image


solidiquis (approved):
Beautiful

PR #688, opened 2026-07-17
python(fix): fixing time type

adamtoth-dev:
# Summary
The request should convert to protobuf time stamp instead of keeping as Python datetime object.

### DIFF
--- a/python/lib/sift_client/_internal/low_level_wrappers/data.py
+++ b/python/lib/sift_client/_internal/low_level_wrappers/data.py
@@ -19,7 +19,7 @@ from sift.data.v2.data_pb2_grpc import DataServiceStub
 
 from sift_client._internal.disk_cache import DiskCache
 from sift_client._internal.low_level_wrappers.base import LowLevelClientBase
-from sift_client._internal.time import to_timestamp_nanos
+from sift_client._internal.time import to_timestamp_nanos, to_timestamp_pb
 from sift_client._internal.util.progress import alive_bar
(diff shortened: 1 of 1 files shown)


python/lib/sift_client/_internal/low_level_wrappers/data.py
```diff
@@ -580,8 +580,8 @@ async def _get_data_impl(
         request_kwargs: dict[str, Any] = {
             "queries": queries,
             "sample_ms": 0,
-            "start_time": start_time,
-            "end_time": end_time,
+            "start_time": to_timestamp_pb(start_time) if start_time else None,
```
alexluck-sift, line 583: I'm confused why start_time is optional, but not end time?

adamtoth-dev: It comes from the definition of the function and the by default the start time is not required, gets set to 0 and the end time get set to now.
It is just a safe guard.

alexluck-sift: so there is no way to query this and get ALL time? e.g. unbound? That seems excessively limiting

alexluck-sift (approved):

python/lib/sift_client/_internal/low_level_wrappers/data.py
adamtoth-dev: Discussed offline, the current behaviour and path is that unless the endtime is provided, we default to the current time (now) as endtime.

PR #690, opened 2026-07-20
python(bug): surface get_data page_size parameter

adamtoth-dev:
## Summary

- Exposes `page_size` on `ChannelsAPIAsync.get_data` so callers can tune how many data points are fetched per gRPC request (default 10,000, server cap 1,000,000).
- Raises the gRPC max receive message length to 50 MB to accommodate large pages — without this, responses near the server cap would hit the default 4 MB limit and fail.

### DIFF
--- a/python/lib/sift_client/resources/channels.py
+++ b/python/lib/sift_client/resources/channels.py
@@ -265,6 +265,7 @@ class ChannelsAPIAsync(ResourceBase):
         start_time: datetime | None = None,
         end_time: datetime | None = None,
         limit: int | None = None,
+        page_size: int | None = None,
         ignore_cache: bool = False,
         show_progress: bool | None = None,
     ) -> dict[str, pd.DataFrame]:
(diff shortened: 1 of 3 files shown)


python/lib/sift_client/resources/channels.py
```diff
@@ -276,6 +277,9 @@ async def get_data(
             start_time: The start time to get data for.
             end_time: The end time to get data for.
             limit: The maximum number of data points to return. Will be in increments of page_size or default page size defined by the call if no page_size is provided.
+            page_size: Number of data points to fetch per request. Defaults to 10,000.
```
alexluck-sift, line 280: any reason not to simply increase this default page size if that is our recommendation?

adamtoth-dev: Discussed with Data team, they feel more comfortable to leave it as the default for now. We want to monitor the effect of this on the data service.

alexluck-sift (dismissed):

alexluck-sift (approved):

PR #694, opened 2026-07-28
ENG-13258: Encourage tool calls to use `limit` parameter when listing out entities

lineville:
Resolves [ENG-13258](https://linear.app/siftstack/issue/ENG-13258/limit-mcp-tool-calls) by updating all of the list tool calls to more aggressively limit results. Right now a vague prompt like "What are all my assets in Sift?" can results in a tool call that returns a large amount of data which can exceed the conversation's context window. This [eval run](https://app.development.siftstack.com/test-results/b0a6144c-cb07-416e-bac7-a36f0d4574e6#1) shows this example of a failing turn because the `list_assets` call pointing at dev returned over 1700 assets which is about 1.4M tokens (exceeding opus 4.8's 1M context window).

> [!IMPORTANT]
> I've pinned the ruff version in the rust-ci workflow which started failing as of this PR now that the latest version of rust fails 38 new tests which all pass on the previous version of ruff.


.github/workflows/rust_ci.yaml
```diff
@@ -83,7 +83,7 @@ jobs:
           set -e
           python3 -m venv .venv
           source .venv/bin/activate
-          pip install maturin pytest pytest-asyncio ruff
+          pip install maturin pytest pytest-asyncio ruff==0.15.20
```
lineville, line 86: The latest version of ruff was starting to throw a whole bunch of errors that were passing with this version 0.15.20 of ruff, so pinning to this.

Brandon-Shippy (changes requested):
The agent advice is great. 

We need to enforce the limit on the paging() function in `service/common/mod.rs` so that a rogue agent who is ignoring advice cannot cause unbounded queries.

Perhaps, set something like a `pub const DEFAULT_LIMIT` and edit the function to clamp any incoming limit from the agent in case it falls outside the range we want. Either way you decide to do it I think we need it bounded in code, not just agent advice.

Brandon-Shippy (approved):

rust/crates/sift_cli/assets/skills/claude-code/SKILL.md
```diff
@@ -33,7 +33,9 @@ to combine them when working with Sift.
 1. **Sift MCP server** — started by `sift-cli mcp`. The preferred surface for
    agents. Exposes structured, authenticated tools:
    - `list_assets`, `list_runs`, `list_channels`, `list_reports`, `list_rules`,
-     `list_rule_versions`, `list_annotations`: discover what exists.
+     `list_rule_versions`, `list_annotations`: discover what exists. Pass `limit`
+     (start at 200, max 1000). Omitting it defaults to 200 and values above 1000
+     clamp to 1000, so raise `limit` when a result comes back capped.
```
evan-sift, line 38: just because we _can_ do 200 to 1000 doesn't mean that I think we should. wdyt about defaulting from 50 to 100, and if we REALLY need to, users can break out and go higher. 

this goes back to us wanting to have this be more like a 'search' endpoint than a list one

rust/crates/sift_mcp/src/tool/assets/mod.rs
```diff
@@ -45,15 +45,16 @@ impl SiftMcpServer {
...
+              - `limit`: max items to return. Start at 200 and only raise it if the result is capped
+                and you still need more. Values are clamped to `1..=1000`; omitting it defaults to 200.
 
             Errors:
               - `INVALID_PARAMS` if `filter` is not a valid CEL expression or `order_by` references an unknown field.
               - `INTERNAL_ERROR` for upstream gRPC failures.
 
             Guidance:
-              - Always prefer narrowing the filter over relying on `limit` — very large unfiltered listings can be slow.
+              - Narrow with `filter` whenever you know what you're looking for; pass `limit` regardless, so an
```
evan-sift, line 56: this verbiage could probably be stronger. We can eval it, but might make sense to have this be a `search_*` (or even `get_*`) pattern instead of a `list_*` pattern to encourage searching

PR #704, opened 2026-07-30
rust(feat): add sift-cli agent lifecycle for Claude and Codex

evan-sift:
## Changes

Adds `sift-cli agent install|update|doctor|uninstall`, replacing the removed project-scoped skill writer with four stateless, user-scoped commands. State is derived from the installed files and each client's own MCP config, so there is no state file to drift.

Registers through each client's native CLI (`claude mcp`, `codex mcp`) rather than editing their config files. A Codex registration counts as Sift-managed only when both the transport and the surrounding metadata match what this CLI writes, so a hand-tuned entry is reported as a conflict instead of being silently replaced. Native replacement is remove-then-add, and a failed add restores the previous entry.

Install and update preflight every target and refuse to touch an unmanaged same-name skill or MCP entry. Skill install mirrors the embedded bundle exactly, so files a previous release left behind are removed. Registrations default to read-only; update preserves whatever mode is installed and rejects a mixed fleet rather than choosing for the user.

Points the MCP server's instructions and destructive-tool refusal at these commands, and makes `agent update` refuse to run on an outdated CLI, printing the installer for the newer release. Release lookup now skips drafts and prereleases.

## Verification

- 11 new `cmd::agent` tests plus the `cmd::version` release-filter test

## Note for reviewers

Rollback of a partial multi-client install is covered here by the native replacement test. The cross-client transaction tests arrive with the JSON-backed clients in the next PR, which can be driven directly without a fake process runner.

### DIFF
--- a/rust/crates/sift_cli/src/cli/mod.rs
+++ b/rust/crates/sift_cli/src/cli/mod.rs
@@ -62,6 +62,11 @@ pub enum Cmd {
     #[command(subcommand)]
     Install(InstallCmd),
 
+    /// Manage the Sift integration for detected AI coding clients
+    #[cfg(feature = "mcp")]
+    #[command(subcommand)]
+    Agent(AgentCmd),
+
     /// Start the Sift MCP server
     #[cfg(feature = "mcp")]
     #[command(hide = true)]
@@ -92,6 +97,43 @@ pub enum InstallCmd {
     /// Install or print shell completions for sift-cli
     #[command(subcommand)]
     Completions(CompletionsCmd),
+}
+
+/// Manage Sift's release-matched skill and MCP sidecar as one bundle.
+#[cfg(feature = "mcp")]
+#[derive(Subcommand)]
+pub enum AgentCmd {
+    /// Install every detected client in safe mode
+    Install(AgentInstallArgs),
+
+    /// Refresh every detected client while preserving its access mode
+    Update(AgentUpdateArgs),
+
+    /// Check the CLI version, skill files, and access modes
+    Doctor,
+
+    /// Remove Sift-owned agent files and registrations
+    Uninstall,
+}
+
+#[cfg(feature = "mcp")]
+#[derive(clap::Args)]
+pub struct AgentInstallArgs {
+    /// Enable tools that modify or archive resources for every detected MCP client
+    #[arg(long)]
+    pub allow_destructive: bool,
+}
+
+#[cfg(feature = "mcp")]
+#[derive(clap::Args)]
+pub struct AgentUpdateArgs {
+    /// Enable tools that modify or archive resources for every detected MCP client
+    #[arg(long, conflicts_with = "read_only")]
+    pub allow_destructive: bool,
+
+    /// Disable destructive tools for every detected MCP client
+    #[arg(long, conflicts_with = "allow_destructive")]
+    pub read_only: bool,
 }
 
 #[derive(Subcommand)]
--- a/rust/crates/sift_cli/src/cmd/agent/files.rs
+++ b/rust/crates/sift_cli/src/cmd/agent/files.rs
@@ -1,0 +1,7 @@
+use std::{fs, path::Path};
+
+pub(super) fn remove_empty_parent(path: &Path) {
+    if let Some(parent) = path.parent() {
+        let _ = fs::remove_dir(parent);
+    }
+}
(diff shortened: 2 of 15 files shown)


rust/crates/sift_cli/src/cli/mod.rs
```diff
@@ -94,6 +99,43 @@ pub enum InstallCmd {
...
+#[derive(Subcommand)]
+pub enum AgentCmd {
+    /// Install every detected client in safe mode
+    Install(AgentInstallArgs),
+
+    /// Refresh every detected client while preserving its access mode
+    Update(AgentUpdateArgs),
+
+    /// Check the CLI version, skill files, and access modes
+    Doctor,
```
lineville, line 113: I've always thought anything that shipped with a `doctor` was a signal it had gotten too complex and brittle 😄 I don't think that's the case for us (hopefully), but maybe it's still handy to have this doctor. I'd be okay leaving it off but if you've found it useful I'll leave it up to you.

rust/crates/sift_cli/src/cmd/agent/files.rs
```diff
@@ -0,0 +1,7 @@
+use std::{fs, path::Path};
+
+pub(super) fn remove_empty_parent(path: &Path) {
```
lineville, line 3: What's this used for? Seems a little strange to be all by itself with one utility to delete a folder.

rust/crates/sift_cli/src/cmd/agent/skill.rs
```diff
@@ -0,0 +1,286 @@
...
+fn unused_backup_path(path: &Path) -> Result<PathBuf> {
+    let parent = path
+        .parent()
+        .filter(|parent| !parent.as_os_str().is_empty())
+        .ok_or_else(|| anyhow!("{} has no parent directory", path.display()))?;
+    let name = path
+        .file_name()
+        .and_then(|name| name.to_str())
+        .unwrap_or("sift");
+    for index in 0..1000 {
```
lineville: Where does this 1000 come from, feel like a magic number potentially?

rust/crates/sift_cli/src/cmd/version.rs
```diff
@@ -60,37 +64,34 @@ pub async fn run() -> Result<ExitCode> {
...
 
-async fn fetch_latest() -> Result<Option<Version>> {
+pub(crate) async fn fetch_latest() -> Result<Option<Version>> {
     let client = ClientBuilder::new()
         .user_agent(USER_AGENT)
         .timeout(Duration::from_secs(5))
         .build()?;
 
-    let mut releases: Vec<GithubRelease> = client
+    let releases: Vec<GithubRelease> = client
```
lineville, line 73: Nice ✨ I like muts irl but not in my rust code

rust/crates/sift_cli/src/cli/mod.rs
lineville: The more I read, I'm thinking `doctor` is more useful for agents than humans. Seems like it's mostly being used as a verification method post-install which seems sensible to me.

rust/crates/sift_cli/src/cmd/version.rs
```diff
@@ -108,3 +109,35 @@ fn install_command(latest: &Version) -> String {
...
+
+#[cfg(test)]
+mod tests {
+    use super::{GithubRelease, latest_stable};
+    use semver::Version;
+
+    #[test]
+    fn latest_release_excludes_drafts_and_prereleases() {
+        let releases = vec![
+            GithubRelease {
```
lineville, line 121: Will we need to bump these tags periodically?

evan-sift: this a test func, the goal is that it filters out the draft / alpha / etc.

rust/crates/sift_cli/src/cmd/agent/files.rs
evan-sift: we do add to it in the next PR https://github.com/sift-stack/sift/pull/705/changes#diff-40355c005042a9bf40fbbcc8174ba9063686dee2fd97db016166dee1ba4e6782

rust/crates/sift_cli/src/cmd/agent/skill.rs
evan-sift: unfortunately, my local harness stomped over the git history... anyway here is the explanation for it. a test cases was added as well that makes sure to remove any unneeded fallback candidates as well  https://github.com/sift-stack/sift/pull/704/changes#diff-a8181c3dbe445b4777a97d1d981b5bccfdbcbc2ce1b33a18598a508be00cde15R20-R23

lineville (dismissed):

lineville (approved):

PR #705, opened 2026-07-30
rust(feat): add Cursor and OpenCode agent support

evan-sift:
## Changes

Extends the agent lifecycle to the two clients that keep MCP registrations in their own JSON config rather than behind a CLI. Cursor uses `~/.cursor/mcp.json` under `mcpServers`; OpenCode uses `~/.config/opencode/opencode.json` under `mcp` with a local command array. Both join Codex on the shared `~/.agents/skills/sift` directory.

Merges into the existing config instead of rewriting it, so unrelated servers and top-level settings survive. Writes go through a temp file and rename with permissions carried over, so an interrupted write cannot truncate a user's config. Refuses to follow a symlinked config or skill directory. An entry carrying keys this CLI does not write is a conflict rather than something to overwrite, and an OpenCode `opencode.jsonc` is reported as unavailable because comments would not survive a rewrite.

Detects these two clients by config directory as well as `PATH`, since both are commonly installed as editors with no CLI on `PATH`.

Snapshots every client's prior config before the first write, so a failure partway through a multi-client install restores all of them plus the skill directories.

## Verification

- 21 `cmd::agent` tests, including the cross-client rollback and preflight cases this PR makes testable

### DIFF
--- a/rust/crates/sift_cli/src/cmd/agent/config.rs
+++ b/rust/crates/sift_cli/src/cmd/agent/config.rs
@@ -1,14 +1,15 @@
 use std::{
     ffi::{OsStr, OsString},
-    io,
-    path::Path,
+    fs,
+    io::{self, ErrorKind},
+    path::{Path, PathBuf},
     process::Command,
 };
 
 use anyhow::{Context, Result, anyhow};
-use serde_json::Value;
+use serde_json::{Map, Value, json};
 
-use super::{AccessMode, Environment, Harness};
+use super::{AccessMode, Environment, Harness, files};
 
 #[derive(Debug)]
 struct CommandOutput {
@@ -51,7 +52,16 @@ struct NativeInspection {
 #[derive(Debug)]
 pub(super) struct Snapshot {
     harness: Harness,
-    previous: Option<NativeEntry>,
+    contents: SnapshotContents,
+}
+
+#[derive(Debug)]
+enum SnapshotContents {
+    Native(Option<NativeEntry>),
+    Json {
+        path: PathBuf,
+        contents: Option<Vec<u8>>,
+    },
 }
 
 #[derive(Debug, Clone, Eq, PartialEq)]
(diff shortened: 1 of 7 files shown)


rust/crates/sift_cli/src/cmd/agent/config.rs
```diff
@@ -68,31 +78,83 @@ pub(super) fn inspect(harness: Harness, environment: &Environment) -> Result<Sta
...
-    let inspection = inspect_native(harness, environment, &SystemRunner)?;
-    let previous = match inspection.state {
-        State::Missing => None,
-        State::Current(_) | State::ManagedDrift(_) => inspection.entry,
-        State::Conflict(detail) | State::Unavailable(detail) => {
-            return Err(anyhow!(
-                "{} MCP registration cannot be snapshotted: {detail}",
-                harness.label()
-            ));
+    let contents = match harness {
```
lineville, line 81: Are there other harnesses we should be worried about like Pi, GitHub Copilot etc? I worry that this isn't generic enough and could limit our ability to work with other harnesses maybe, unless I'm missing something here?

rust/crates/sift_cli/src/cmd/agent/config.rs
```diff
@@ -68,31 +78,83 @@ pub(super) fn inspect(harness: Harness, environment: &Environment) -> Result<Sta
...
+                    Err(error).with_context(|| format!("failed to inspect {}", path.display()))
+                }
+            },
+        },
+    }
 }
 
 fn inspect_with(harness: Harness, environment: &Environment, runner: &dyn Runner) -> Result<State> {
-    Ok(inspect_native(harness, environment, runner)?.state)
+    match harness {
```
lineville, line 153: Same idea, it feels like it could get tricky to support an unlimited number of harnesses like this.

lineville (commented):
Overall looks solid, my main concern is that a lot of this is specific to 4 known harnesses. How do we expect this to handle other coding harnesses, or do we just want to accept that these are the handful we support and can validate?

rust/crates/sift_cli/src/cmd/agent/config.rs
evan-sift: I thought so at first, but then just went by the top 3 + opencode.

I tried to add in Pi, but it requires the MCP adapter, and I didn't want to enforce that up front. 

WDYT about:
- a `--skills-path` param to allow users to install the skills wherever they want 
- something like `--mcp-config` that just outputs the raw JSON for you to manually add into your agent (or event with some other text output in the CLI)

evan-sift: thinking on it more. honestly, I don't think that this is for Beta scope, we can revisit it later

lineville (dismissed):

lineville (approved):

PR #706, opened 2026-07-30
rust(feat): add profile-aware agent installation

evan-sift:
## Changes

Lets the agent bundle target a named Sift profile. The global `--profile` flag was already parsed but hidden; this unhides it and threads it into the MCP arguments each client is registered with, so a registration points at one profile rather than whatever the default resolves to at runtime.

Widens the managed registration from an access mode to access plus profile. Both are read back out of the client configs, so update preserves the installed profile the same way it already preserved read-only or destructive access, with no state file. Adds `--default-profile` to move every client back to the default, mutually exclusive with `--profile`.

Only the exact argument shapes this CLI writes count as managed, now including `mcp --profile <name>` and `mcp --profile <name> --allow-destructive`. A profile that is empty or starts with `-` is rejected rather than treated as managed.

Mixed profiles across detected clients are an error, matching the existing treatment of mixed access modes: doctor reports them and update refuses to guess.

## Verification

- 25 `cmd::agent` tests

## Note for reviewers

No CHANGELOG entry is added here, matching the original branch. If `--profile` / `--default-profile` should be called out for release, that line is worth adding.

### DIFF
--- a/rust/crates/sift_cli/src/cmd/agent/mod.rs
+++ b/rust/crates/sift_cli/src/cmd/agent/mod.rs
@@ -35,6 +35,44 @@ impl AccessMode {
     }
 }
 
+#[derive(Debug, Clone, Eq, Hash, PartialEq)]
+pub(super) enum Profile {
+    Default,
+    Named(String),
+}
+
+impl Profile {
+    fn from_option(profile: Option<String>) -> Self {
+        match profile {
+            Some(profile) => Self::Named(profile),
+            None => Self::Default,
+        }
+    }
+
+    fn label(&self) -> String {
+        match self {
+            Self::Default => "default profile".to_string(),
+            Self::Named(profile) => format!("profile '{profile}'"),
+        }
+    }
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub(super) struct Registration {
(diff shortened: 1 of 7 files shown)


rust/crates/sift_cli/src/cmd/agent/mod.rs
```diff
@@ -272,11 +330,36 @@ pub async fn doctor() -> Result<ExitCode> {
...
+        println!("[error] Detected MCP clients use mixed profiles.");
+        unhealthy = true;
+    }
+    let expected_profile = expected_profile.map(Profile::Named);
+    let unexpected_profile = expected_profile.as_ref().is_some_and(|expected| {
+        profiles.len() != environment.harnesses.len()
+            || profiles.iter().any(|installed| installed != expected)
+    });
+    if unexpected_profile {
+        println!(
```
lineville, line 357: What's the case where this happens? Someone has two harnesses running on their machine and each picks up a different profile?

lineville (dismissed):
Looks good just want to understand what this mixed profile issue is? How does one create this issue of mixed profiles that we're protecting against?

lineville (dismissed):

lineville (approved):

PR #710, opened 2026-07-31
rust(feat): Improve search behavior in list tools

lineville:
Improves the behavior of the MCP's searching capabilities by providing clearer guidance on how to use CEL match expressions to pull less data and be better and finding what the prompter is looking for.


rust/crates/sift_cli/assets/skills/sift/SKILL.md
```diff
@@ -37,13 +37,17 @@ to combine them when working with Sift.
      (start at 50, max 200). Omitting it defaults to 50 and values above 200
      clamp to 200, so raise `limit` when a result comes back capped.
    - `list_report_rule_summaries`: per-rule pass/fail/open breakdown for a report.
+   - Searching any of those lists: prefer a pattern over an exact `==`.
+     `name.matches("(?i)rover")` is RE2 and case-insensitive; `contains`,
+     `startsWith`, and `endsWith` are case-sensitive. An empty result usually
+     means the pattern was too narrow, so retry a shorter fragment before you
+     tell the user that nothing exists. Each tool's description lists its own
+     filterable fields. When a request is too vague to filter on, sample with a
+     small `limit` and ask the user to narrow it rather than guessing.
```
evan-sift:
```suggestion
     small `limit` (e.g. 5) and ask the user to narrow it rather than guessing.
```

rust/crates/sift_mcp/CLAUDE.md
```diff
@@ -499,6 +499,37 @@ When you add or update a list tool:
...
 
+### Search guidance on the `filter` bullet
+
+A field list does not tell an agent *how* to search it, and the default failure is an exact `==`
+against a value the caller only half knows. Every tool that takes a `filter` therefore closes its
+`filter` bullet with two lines naming that resource's own text fields:
+
+```
+                Prefer a pattern over `==`: `name.matches(\"(?i)rover\")` is RE2, case-insensitive.
+                `contains`/`startsWith`/`endsWith` are case-SENSITIVE. Empty result: retry a shorter fragment.
```
evan-sift: open discussion, but perhaps we should start with what we do want. This phrasing is a bit passive (and most models should be smart enough), but wondering if we want something like this: (I workedshopped this locally for a bit)

```
When filtering or searching, use `name.matches("(?i)rover")`, not `==`. Use
`==` only for an exact value from a prior result.
`contains`/`startsWith`/`endsWith` are case-SENSITIVE: `contains("Rover")`
silently misses `rover-01`. An empty result is not proof of absence — retry
once with a shorter fragment.
```

evan-sift: we'll then want to update all the instances in the tools with whatever we land on

lineville: Updated all of the relevant tools based on your suggestion in [8f01f0e](https://github.com/sift-stack/sift/pull/710/commits/8f01f0eafc83dd1841562a75ab17faa765b53ff1)

evan-sift (approved):

PR #712, opened 2026-08-01
rust(feat): split the sift skill into a router and reference files

evan-sift:
breakup skills for increased token efficiency and better agent directions


rust/crates/sift_cli/assets/skills/sift/SKILL.md
```diff
@@ -25,269 +24,83 @@ description: >-
...
+
+- **Search a list.** Filter with a pattern rather than an exact match. Each
+  tool's description names its own filterable fields. When the request is too
+  vague to filter on, sample with a small `limit` and ask the user to narrow
+  it. Do not guess.
+- **Attribute something to a person.** Resolve the person with `list_users`,
+  then filter another list on `created_by_user_id`. For "runs I created", pass
+  `me: true`. Never guess which listed user is the caller.
+- **Produce numbers.** `get_data` writes a Parquet file. `sql` then queries it.
+  Add `upload_dataset` when the result belongs back in Sift.
```
evan-sift, line 74: Have local SQL is great, and definitely useful for analysis. But, we'll also want to support querying `get_data` with adhoc calc channel expressions. The goal there is so that users can save back CC's into the app if it makes sense for them to do so (and so that the data locally stays as close to what the generated source of truth would be when requesting from Sift). 

Seems like the `get_data` MCP tool does not currently support a CEL expression mode, which is fine. Probably a GA task. But just calling this out here. 

viz @rowenwu @solidiquis @lineville

rust/crates/sift_cli/assets/skills/sift/references/cli.md
```diff
@@ -0,0 +1,111 @@
...
+
+`sift-cli` and the Python library import several file types directly. The CLI
+supports CSV, Parquet (flat-dataset and channel-per-row), TDMS, HDF5, ULog,
+and `sift_stream` backups.
+
+If the user's file type is not supported by the CLI or MCP server, you have
+three options:
+
+1. Transform the data into a CSV or Parquet file and import that with the CLI.
+2. Transform the data into a Parquet file and upload it with the MCP
```
lineville: 1 and 2 seem a bit ambiguous to me, if it can transform the data into a Parquet file is there a reason to opt to import it with the CLI vs MCP? Maybe we could be more prescriptive here if we have a preference.

rust/crates/sift_cli/assets/skills/sift/references/integration-code.md
```diff
@@ -0,0 +1,21 @@
+# Writing integration code against Sift
+
+When a user wants help integrating their own systems with Sift, consult the
+references below and write code against `sift_client` for Python or
```
lineville, line 4: Should we make it clear if there's a preference on python vs rust or to just infer which makes most sense based on context?

lineville (dismissed):

rust/crates/sift_cli/assets/skills/sift/references/integration-code.md
evan-sift: Might make sense to include something specific like:

> Infer based on the user's current environment. If both Python and Rust are installed, ask the user which language they want Sift-related code to be written in.

rust/crates/sift_cli/assets/skills/sift/references/cli.md
evan-sift: this is cleaned up now to be more clear.

lineville (dismissed):

lineville (approved):

PR #718, opened 2026-08-04
rust(feat): add report tooling for external rules

Brandon-Shippy:
# Description
- Adds three MCP tools for report templates:
    - list_report_templates
    - create_report_template
    - update_report_template — update a template's name, description, tags, rule set, metadata, or
  archive state
  
Bundles in the follow bug fix:
- Fixes create_report so it now starts the report's evaluation. Reports created through the MCP now progress past the initial state as expected. 
- `list_runs` and `list_test_reports` were missing guidance about the default of finding archived data is false.

# Verification

- Unit tests
- Manual testing, all tools returned properly

Testing of all the tools:

`create_report_template`:
image

`list_report_templates`:
Screenshot 2026-08-04 at 3 41 36 PM

`update_report_template`:
Screenshot 2026-08-04 at 3 42 45 PM


rust/crates/sift_mcp/src/tool/test_reports/mod.rs
```diff
@@ -83,6 +83,7 @@ impl SiftMcpServer {
             Guidance:
               - To audit a single run, resolve the report first (by `name`, `run_id`, or `test_case`), then pass
                 its `test_report_id` to the step and measurement tools.
+              - Use `is_archived == false` to exclude archived test reports unless they're explicitly needed.
```
evan-sift: should invert this. By _**default**_ we should NOT query things that are archived. The user needs to explicitly ask to get `is_achived == true` to query things that are.

rust/crates/sift_cli/assets/skills/sift/SKILL.md
```diff
@@ -76,6 +76,14 @@ exists.
...
 - **Answer a question about how Sift works.** Call `search_docs`. Do not answer
   from memory, and cite the page you used.
+- **Run rules against a run.** Find or author rules with `list_rules` /
+  `create_rule`. To reuse the same rule set across many runs, bundle standard
+  rules (`is_external: false`) into a template with `create_report_template`,
+  then call `create_report` with `report_template_id`. For a one-off — or for
+  ad-hoc rules (`is_external: true`, which the API also calls "external" but
+  cannot be attached to a template) — skip the template and pass `rule_ids`,
+  `rule_client_keys`, or `rule_version_ids` directly to `create_report`. Track
+  progress via `list_report_rule_summaries`.
```
evan-sift:
```suggestion
- **Evaluate rules against a run.** Find rules with `list_rules` and author rules
  with `create_rule`. To reuse the same rule set across many runs, bundle
  standard rules (`is_external: false`) into a template with `create_report_template`,
  then call `create_report` with `report_template_id`. For a one-off — or for
  ad-hoc rules (`is_external: true`, which the API also calls "external" but
  cannot be attached to a template) — skip the template and pass `rule_ids`,
  `rule_client_keys`, or `rule_version_ids` directly to `create_report`. Track
  progress via `list_report_rule_summaries`.
```

rust/crates/sift_mcp/src/tool/test_reports/mod.rs
evan-sift: (should be applied everywhere)

rust/crates/sift_mcp/src/tool/report_templates/mod.rs
```diff
@@ -0,0 +1,367 @@
...
+                `report_template_id`, `tag_id`, `tag_name`, `client_key`, `metadata`, `name`, `is_archived`.
+                Reference metadata entries as `metadata.{key}` (e.g. `metadata.owner == \"qa-team\"`).
+              - `order_by`: optional comma-separated `FIELD_NAME[ desc]` list. Orderable fields: `created_date`,
+                `modified_date`. Default sort is `created_date desc` (newest first). Example:
+                `\"created_date desc,modified_date\"`.
+              - `limit`: max items to return. Start at 50 and only raise it if the result is capped and you still
+                need more. Values are clamped to `1..=200`; omitting it defaults to 50.
+              - `organization_id`: optional. Required only when the caller belongs to multiple organizations;
+                scopes the listing to that org. Omit for single-organization users.
+
```
evan-sift, line 84: Importantly, let's make sure that the tool param text is updated with the current meta. `order_by` and `filter` looks out of date with the latest. `filter` specifically needs to document the regex matching for searching:
```
                When filtering or searching, use `name.matches(\"(?i)avionics\")`, not `==`. Use `==` only for an
                exact value from a prior result. `contains`/`startsWith`/`endsWith` are case-SENSITIVE:
                `contains(\"Avionics\")` silently misses `avionics-power-limit`.
```

Brandon-Shippy: https://github.com/sift-stack/sift/pull/718/commits/6638bb472d7b3880e829e109dc4a56b39f11f01a

rust/crates/sift_cli/assets/skills/sift/SKILL.md
Brandon-Shippy: https://github.com/sift-stack/sift/pull/718/commits/6638bb472d7b3880e829e109dc4a56b39f11f01a

rust/crates/sift_mcp/src/tool/test_reports/mod.rs
Brandon-Shippy: https://github.com/sift-stack/sift/pull/718/commits/6638bb472d7b3880e829e109dc4a56b39f11f01a

Double check my updated wording. I think it addresses this

evan-sift (approved):

PR #720, opened 2026-08-05
rust(feat): add top level job commands cli

Brandon-Shippy:
# Description
- Adds `sift-cli get jobs`, `get job <ID>`, `status job <ID>`, and `wait job <ID> [ID ...]`.
- Import no-wait path now prints the job id so scripts can capture it.
- mdBook + Sift skill updated.
- Changes `wait_until_completion` to now instead use the more general `poll_until_terminal` of which is the loop that we use with all the jobs commands

# Verification
- Manual Verification
- Unit Tests

`get jobs` command. Can filter and set it fetch the most recent X runs. 
Screenshot 2026-08-05 at 2 45 05 PM

`get job` command. Can input job id to query.
Screenshot 2026-08-05 at 2 46 08 PM

`wait job` command. Blocks i/o until job id is finished or errored.
Screenshot 2026-08-05 at 2 47 00 PM

Imports now give the job id for users to get and be able to query in CI or other automation workflows.
Screenshot 2026-08-05 at 2 51 15 PM


rust/crates/sift_cli/src/cmd/jobs/mod.rs
```diff
@@ -0,0 +1,257 @@
...
+    let mut any_failed = false;
+    for handle in handles {
+        let (job_id, outcome) = handle.await.context("wait task panicked")?;
+        match outcome {
+            Ok(Some(job)) => {
+                let status = job.job_status();
+                let line = format!("{}: {}", job_id.cyan(), status_label(status));
+                match status {
+                    JobStatus::Finished => Output::new().line(line).print(),
+                    JobStatus::Cancelled => Output::new().line(line).print(),
```
wei-qlu: A cancelled job here would have `any_failed` stay false and result in a 0 exit code, contradicting what `jobs.md` and `--help` are saying, is this intentional?

Brandon-Shippy: https://github.com/sift-stack/sift/pull/720/commits/1d2970b95585b3e88e38943827f2f09951fae0f5

Youre right

PR #721, opened 2026-08-05
rust(chore): 0.4.0 prep work

Brandon-Shippy:
# Description
- Un feature flag mcp
- update changelog
- bump cargo.toml version number

Will update this changelog depending on if anything else gets added in before release

### DIFF
--- a/rust/crates/sift_cli/CHANGELOG.md
+++ b/rust/crates/sift_cli/CHANGELOG.md
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [v0.4.0] - August 5, 2026
+
 ### What's New
 
 - Added `sift-cli agent install`, `update`, `doctor`, and `uninstall` for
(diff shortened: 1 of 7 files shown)


rust/crates/sift_cli/Cargo.toml
```diff
@@ -1,9 +1,9 @@
 [features]
-mcp = []
+test-reports = ["sift_mcp/test-reports"]
```
lineville, line 2: Maybe I missed this, but why is `test-report` a new feature here?

rust/crates/sift_cli/CHANGELOG.md
```diff
@@ -5,17 +5,46 @@ This project adheres to [Semantic Versioning](http://semver.org/).
...
+  reusable set of rules that any `create_report` call can inherit via
+  `report_template_id`.
+- Improved MCP filter guidance: agents default `is_archived == false` on
+  every list call, use regex/case-insensitive matching
+  (`name.matches("(?i)...")`) instead of `==` for text search, and always
+  pass `limit` to keep results within the context window.
+- MCP tool calls no longer proceed on an empty list result. Empty lists
+  surface to the user for guidance instead of triggering a follow-up call
+  with an empty ID.
+- Added an opt-in `test-reports` Cargo feature on `sift_mcp` (forwarded from
```
lineville, line 40: What's the reason we want this to be opt-in and not just included by default?

lineville (approved):
Looks good to me! Was mostly just curious about the decision for test-reports to be opt-in, other than that notes look solid.

rust/crates/sift_cli/Cargo.toml
Brandon-Shippy: We are feature flagging it away, it got edited here since at the same time we are un-feature flagging mcp from the cli

rust/crates/sift_cli/CHANGELOG.md
Brandon-Shippy: Its not a GA feature yet and should only be surfaced to customers who have access to it. Otherwise they will see tooling that their API key cannot access

PR #722, opened 2026-08-07
rust(feat): --allow-create flag mcp

Brandon-Shippy:
# Description
Adds a new `--allow-create` flag on the mcp. This makes it so the user has to explicitly opt-in to have tools that create data in sift. 
Also updates documentation, skills, and tests.
Updates the skills to give stricter guidance around respecting user policy via the flags

# Verification
Updated the tests around it
Manual verification of all the modes

Screenshot 2026-08-07 at 4 02 59 PM

### DIFF
--- a/rust/crates/sift_mcp/src/tool/assets/test.rs
+++ b/rust/crates/sift_mcp/src/tool/assets/test.rs
@@ -36,7 +36,7 @@ async fn server_with_mock_and_flag(
         SiftMcpServer::new(
             channel,
             String::from("https://app.test.local"),
-            allow_destructive,
+            false,
             allow_destructive,
         ),
         handle,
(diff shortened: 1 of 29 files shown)


rust/crates/sift_mcp/src/tool/report_templates/test.rs
```diff
@@ -36,6 +36,7 @@ async fn server_with_mock_and_flag(
             channel,
             String::from("https://app.test.local"),
             allow_destructive,
+            allow_destructive,
```
evan-sift: should one of these be `allow_create`?

rust/crates/sift_mcp/src/tool/assets/test.rs
```diff
@@ -37,6 +37,7 @@ async fn server_with_mock_and_flag(
             channel,
             String::from("https://app.test.local"),
             allow_destructive,
+            allow_destructive,
```
evan-sift: same here

evan-sift (approved):

PR #723, opened 2026-08-10
feat: MCP check for updates

evan-sift:
- automatically check for sift-cli updates on first tool use
- flags for enabling / disabling update checker
- provides helpful scripts to run to update to the latest version when out of date

codex update
claude update

### DIFF
--- a/rust/crates/sift_cli/src/cmd/mcp.rs
+++ b/rust/crates/sift_cli/src/cmd/mcp.rs
@@ -4,11 +4,13 @@ use std::{
 };
 
 use anyhow::{Error, Result};
+use semver::Version;
 use sift_rs::Credentials;
+use tokio::sync::watch;
 use tracing_subscriber::EnvFilter;
 
 use crate::cli::McpArgs;
-use crate::cmd::Context;
+use crate::cmd::{Context, version};
 use crate::util::tty::Output;
 
 pub async fn report_startup_error(error: Error) -> Result<ExitCode> {
@@ -38,6 +40,12 @@ pub async fn run(ctx: Context, args: McpArgs, app_uri: String) -> Result<ExitCod
         "starting Sift MCP server"
     );
 
+    let update_check = if args.disable_update_check {
+        None
+    } else {
+        Some(start_update_check())
+    };
+
     let credentials = Credentials::Config {
         uri: ctx.grpc_uri,
         apikey: ctx.api_key,
@@ -48,6 +56,8 @@ pub async fn run(ctx: Context, args: McpArgs, app_uri: String) -> Result<ExitCod
         app_uri,
         args.allow_create,
         args.allow_destructive,
+        env!("CARGO_PKG_VERSION").to_string(),
+        update_check,
     )
     .await
     {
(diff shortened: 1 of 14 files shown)


rust/crates/sift_cli/src/cmd/mcp.rs
```diff
@@ -61,3 +67,168 @@ pub async fn run(ctx: Context, args: McpArgs, app_uri: String) -> Result<ExitCod
...
+            }
+        };
+        log_update_notice(&update_check);
+        return ready_update_check(update_check);
+    }
+
+    let (sender, receiver) = watch::channel(sift_mcp::UpdateCheck::Checking {
+        current_version: current_version.clone(),
+    });
+    tokio::spawn(async move {
```
Brandon-Shippy, line 113: suggestion: `write_release_cache` only runs on `Ok(Some(latest))`, so a failed check and a machine that never checked are identical in terms of their state on the cache. Anywhere that can't reach `api.github.com`, for whatever reason, takes the miss branch on every MCP server start and eats up to 5s on the first `check_for_updates` of every session.

Worth recording failures as `{"latest_version": null, "checked_at": ...}` on a short TTL, maybe 15m, so we can skip the fetch and report unavailable right away. `read_release_cache_at` needs a third state for that.

Brandon-Shippy (commented):
Looks good, some non-blocking comments.

rust/crates/sift_cli/src/cmd/mcp.rs
```diff
@@ -61,3 +67,168 @@ pub async fn run(ctx: Context, args: McpArgs, app_uri: String) -> Result<ExitCod
...
+            return ready_update_check(sift_mcp::UpdateCheck::Unavailable {
+                current_version,
+                message: "Could not parse the installed sift-cli version. Run `sift-cli --version` for details."
+                    .to_string(),
+            });
+        }
+    };
+
+    if let Some(cached) = version::read_release_cache(&current) {
+        let update_check = if cached.is_outdated {
```
Brandon-Shippy: nit: The `is_outdated` is redundant. `update_check_from_versions` already returns `Current` when `latest <= current`, so same value would be produced.

rust/crates/sift_mcp/src/server/mod.rs
```diff
@@ -2,14 +2,39 @@ use rmcp::{
...
-use crate::policy::RetryPolicy;
+#[cfg(test)]
+use crate::UpdateCheck;
+use crate::{UpdateCheckReceiver, policy::RetryPolicy};
+
+#[cfg(test)]
+mod test;
+
+const UPDATE_CHECK_INSTRUCTIONS: &str = concat!(
+    "Call `check_for_updates` once before the first other Sift tool in each session. ",
```
lineville: What does first "other" mean here? This is aiming to have it check for updates at the beginning of each session not each tool call correct?

rust/crates/sift_cli/src/cmd/mcp.rs
```diff
@@ -61,3 +67,168 @@ pub async fn run(ctx: Context, args: McpArgs, app_uri: String) -> Result<ExitCod
...
+            sift_mcp::UpdateCheck::Current {
+                current_version,
+                latest_version: cached.latest_version.to_string(),
+            }
+        };
+        log_update_notice(&update_check);
+        return ready_update_check(update_check);
+    }
+
+    let (sender, receiver) = watch::channel(sift_mcp::UpdateCheck::Checking {
```
lineville, line 110: We want this to be async in the background rather than blocking correct? I think this is fine, but just double checking we didn't just want to do the version check synchronously before anything else to be even firmer on enforcing the latest version.

lineville (dismissed):

rust/crates/sift_cli/src/cmd/mcp.rs
evan-sift: yeah, I didn't want it blocking, otherwise that could add annoying latency to the first request

rust/crates/sift_mcp/src/server/mod.rs
evan-sift: other = anything that isn't `check_for_updates`. I can make clearer

rust/crates/sift_cli/src/cmd/mcp.rs
evan-sift: ddc3cb0588acd4f52eb367746efa724c839078f6

rust/crates/sift_cli/src/cmd/mcp.rs
evan-sift: ddc3cb0588acd4f52eb367746efa724c839078f6

rust/crates/sift_mcp/src/server/mod.rs
evan-sift: ddc3cb0588acd4f52eb367746efa724c839078f6

lineville (approved):

Brandon-Shippy (approved):

PR #725, opened 2026-08-11
rust(chore): update rmcp to 3.1.2 and make compliant with 2026-07-28

evan-sift:


rust/crates/sift_mcp/src/server/mod.rs
```diff
@@ -93,7 +94,14 @@ impl ServerHandler for SiftMcpServer {
                 .unwrap_or(b.name.as_ref());
             a_key.cmp(b_key)
         });
-        Ok(ListToolsResult::with_all_items(tools))
+        let mut result = ListToolsResult::with_all_items(tools);
+        if context
+            .protocol_version()
+            .is_some_and(|version| version >= ProtocolVersion::V_2026_07_28)
```
lineville, line 100: It seems odd to me to reference a particular version directly like this, is this just to handle backwards compatibility? If its above that version you have to wrap the result in this cache scope and if it's older you just wrap it with `Ok`?

lineville (dismissed):
Approved as a hot fix. I do wonder if we should permanently have that particular protocol version in source like that, but if it fixes the missing tools we should ship this

Brandon-Shippy (approved):
lgtm
In the future it might be worth us just dropping our own `list_tools()` and instead letting rmcp's macro auto-gen it. So when a protocol updates we dont need to sync the fields by hand

rust/crates/sift_mcp/src/server/mod.rs
evan-sift: yeah this is purely for backwards compat

lineville (approved):

PR #726, opened 2026-08-11
chore: change ci runners to reduce x86_64 macOS build time

evan-sift:
when updating ALL runners to be better, here are the rough diffs:

```
┌────────────────────────────────┬──────────┬─────────┐
│              Job               │   Old    │  Final  │
├────────────────────────────────┼──────────┼─────────┤
│ Linux (RunsOn 8cpu, on-demand) │ 16.5 min │ 6m08s   │
├────────────────────────────────┼──────────┼─────────┤
│ arm64 macOS                    │ 16 min   │ 8m37s   │
├────────────────────────────────┼──────────┼─────────┤
│ x86_64 macOS (arm64 cross)     │ 41 min   │ 11m26s  │
├────────────────────────────────┼──────────┼─────────┤
│ Windows (unchanged runner)     │ 30 min   │ 28m10s  │
├────────────────────────────────┼──────────┼─────────┤
│ Release wall clock             │ ~42 min  │ ~29 min │
└────────────────────────────────┴──────────┴─────────┘
```

so I think if we just update the `x86_64 macOS (arm64 cross)` runner, that will be sufficient, since now we're bottlenecked by windows. reverted the changes I had for Linux and `arm64 macOS`  since the ~16m fall within the 30m bounds now

### DIFF
--- a/dist-workspace.toml
+++ b/dist-workspace.toml
@@ -33,7 +33,7 @@ github-build-setup = "includes/build-setup.yml"
 # Windows stays on the default GitHub runner: the RunsOn stack in this org does
 # not serve Windows jobs (verified on PR #726, the job was never picked up).
 [dist.github-custom-runners]
-x86_64-unknown-linux-gnu = "runs-on,runner=8cpu-linux-x64,image=ubuntu22-full-x64"
+x86_64-unknown-linux-gnu = "runs-on,runner=8cpu-linux-x64,image=ubuntu22-full-x64,spot=false"
 
 # Build the Intel macOS binary on an arm64 runner. The M-series runners are
 # much faster than the Intel `macos-*-intel` pool, and Rosetta 2 runs any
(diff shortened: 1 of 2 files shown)


jared-sift:
@evan-sift fyi, `8cpu-windows-x64` is not a valid runner spec

dist-workspace.toml
```diff
@@ -17,6 +17,17 @@ targets = ["aarch64-apple-darwin", "x86_64-apple-darwin", "x86_64-unknown-linux-
 install-path = "CARGO_HOME"
 # Whether to install an updater program
 install-updater = false
+# Build only sift_cli with `--package` instead of the whole workspace. The
+# workspace build compiles sift_stream_bindings (pyo3), which is not part of
+# the CLI release and fails to link when the apple target is cross-compiled.
+precise-builds = true
```
evan-sift, line 23: is this correct?

lineville: I found the docs on [precise-builds](https://axodotdev.github.io/cargo-dist/book/reference/config.html#precise-builds) and it seems like it probably will result in smaller builds based on this description. But the docs on [github-custom-runners](https://axodotdev.github.io/cargo-dist/book/reference/config.html#github-custom-runners) don't seem to show options for runner and host, but maybe they just aren't documented well?

solidiquis (approved):

PR #729, opened 2026-08-13
rust(chore): add in anonymous telemetry event names for mcp tools

evan-sift:
mcp anonymous telemetry


rust/crates/sift_mcp/src/lib.rs
```diff
@@ -97,19 +100,71 @@ pub async fn run_with_update_check(
...
+            allow_destructive,
+            cli_version,
+            update_check,
+            client_event_reporter: client_event::ClientEventReporter::default(),
+        },
+    )
+    .await
+}
+
+pub async fn run_with_client_events(
```
lineville, line 118: Is this a separate wrapper because we might not always have it running with client events being emitted? Are we supporting a variant where it doesn't send events?

lineville (approved):
Is it documented somewhere how we can view these events?

rust/crates/sift_mcp/src/lib.rs
evan-sift: not atm, we're always sending anonymous events, but this just gives us a way to lookup which event maps to which tool

PR #731, opened 2026-08-18
rust(feat): sift cli get assets

Brandon-Shippy:
# Description
Adds `sift-cli get asset` for listing assets. Returns in a table by default with hyperlinks to the asset in explore. By default we list out the most recent 50 assets by `modified_on` date
It has a few flags:
    - `--filter` takes a CEL expression
    - `--limit` lets you edit how many rows you get back
    - `--output-format` lets you determine the output format, current flags are text and json
   

# Verification
- Unit tests
- Manual testing

### DIFF
--- a/rust/crates/sift_cli/src/cli/mod.rs
+++ b/rust/crates/sift_cli/src/cli/mod.rs
@@ -51,6 +51,10 @@ pub enum Cmd {
     #[command(subcommand)]
     Export(ExportCmd),
 
+    /// Get commands to discover and pull data from Sift
+    #[command(subcommand)]
+    Get(GetCmd),
+
     /// Ping the Sift API to verify credentials and connectivity
     Ping,
 
@@ -309,6 +313,12 @@ pub enum ImportCmd {
     /// Run without a subcommand to import; use `ls` to list files without importing.
     #[command(name = "backups")]
     Backup(BackupArgs),
+}
+
+#[derive(Subcommand)]
+pub enum GetCmd {
+    /// Get assets
+    Asset(GetAssetArgs),
 }
 
 #[derive(Subcommand)]
@@ -867,4 +877,11 @@ impl DocArgs {
     fn default_addr() -> SocketAddr {
         "0.0.0.0:3000".parse().unwrap()
     }
+}
+
+#[derive(clap::Args)]
+pub struct GetAssetArgs {
+    /// Filter option for filtering search
+    #[arg(long)]
+    pub filter: Option<String>,
 }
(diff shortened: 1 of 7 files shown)


wei-qlu (approved):
looks good to me

rust/crates/sift_cli/src/cmd/get/assets.rs
```diff
@@ -0,0 +1,79 @@
...
+
+const ASSET_ID_CHAR_LENGTH_WHITESPACE_LENGTH: usize = 38;
+
+pub async fn run(ctx: Context, args: GetAssetArgs) -> Result<ExitCode> {
+    let grpc_channel = create_grpc_channel(&ctx)?;
+
+    let ListAssetsResponse { assets, .. } = AssetServiceClient::new(grpc_channel)
+        .list_assets(ListAssetsRequest {
+            filter: args.filter.unwrap_or_default(),
+            order_by: "modified_date desc".to_string(),
```
solidiquis, line 28: Should this be exposed as an option

rust/crates/sift_cli/src/cmd/get/assets.rs
```diff
@@ -0,0 +1,79 @@
...
+    cmd::Context,
+    util::{
+        api::create_grpc_channel,
+        app_uri::normalize_app_uri,
+        explore_url::build_explore_url,
+        tty::{Output, hyperlink, link_style, stdout_is_tty},
+    },
+};
+
+const ASSET_ID_CHAR_LENGTH_WHITESPACE_LENGTH: usize = 38;
```
solidiquis, line 20: We should use an ASCII table crate instead

rust/crates/sift_cli/src/cli/mod.rs
```diff
@@ -868,3 +878,25 @@ impl DocArgs {
...
+
+#[derive(Clone, Copy, ValueEnum, Debug, PartialEq, Eq)]
+#[value(rename_all = "lowercase")]
+pub enum OutputFormats {
+    Text,
+    Json,
+}
+
+#[derive(clap::Args)]
+pub struct GetAssetArgs {
```
solidiquis, line 890: We will be using these in a lot of other `get` subcommands so this should be a reusable struct that we flatten

rust/crates/sift_cli/src/cmd/get/assets.rs
Brandon-Shippy: https://github.com/sift-stack/sift/pull/744

rust/crates/sift_cli/src/cmd/get/assets.rs
Brandon-Shippy: https://github.com/sift-stack/sift/pull/744

rust/crates/sift_cli/src/cli/mod.rs
Brandon-Shippy: https://github.com/sift-stack/sift/pull/744

PR #732, opened 2026-08-18
rust(feat): add a fields projection and item count to list_* tools

rowenwu:
## Description

Two changes to the `list_*` tools, both about not making the caller do work the server can do.

### 1. `fields` projection

List tools return complete objects with no way to request a subset of fields. A wide listing therefore blows the tool-result token limit, spills to a temp file, and has to be read back and parsed before the caller can use it — roughly three extra operations for a result it already received.

Reproduced against a development environment: `list_channels` scoped to one asset at `limit: 200` returns 91,008 characters, about 470 per channel. `limit` does not help, because it caps row count rather than payload width, and there is no offset or page token exposed to walk the remainder. A caller that wants every channel name has no way to ask for only the names.

Adds an optional `fields` array to every list tool:

- `list_channels`, `list_runs`, `list_assets`, `list_annotations`
- `list_rules`, `list_test_reports`, `list_test_steps`, `list_test_measurements` (they share `ListParams`, so they get it too rather than accepting a parameter they would ignore)

Behavior:

- Omitted, or an empty array, returns the full object. No change for existing callers.
- Names match case-insensitively and ignore underscores, so `asset_id` and `assetId` both resolve. Callers carry names between tools that spell them differently.
- A requested name that matched nothing on any item comes back in `unmatched_fields` beside the results, so a typo is visible instead of silently narrowing the response. Proto3 omits fields at their default value, so a name absent from only some items is not reported.
- The injected `url` field is projectable like any other key.

Measured on a 120-channel asset: 52,990 characters without the projection, 3,310 with `fields: ["name"]`. The same task drops from 11 tool calls to 2, because the unprojected response has to be recovered in filtered chunks.

### 2. `count` on every list response

Counting a JSON array by eye is arithmetic the caller should not have to do, and it goes wrong. Observed on a 120-row response: the caller reported a total of 122, having estimated the count from the payload before enumerating it, then explained away its own correct enumeration of 120 rather than trusting it.

Every list body now carries `count`:

```json
{ "count": 120, "channels": [ … ] }
```

Exact, free to compute here, and already accounting for `limit` and any `fields` projection.

Each tool description documents both parameters, since that is the only documentation the calling model gets.

### Scope

MCP-side only. The server still fetches full objects over gRPC and pages through all of them, because no List RPC accepts a read field mask (`FieldMask` appears in the protos only as `update_mask` on Update requests). A read field mask on the List RPCs is tracked separately as backend work. The token limit applies to the tool result, so this change alone removes the overflow.

Not covered: `list_reports`, `list_report_rule_summaries`, `list_rule_versions` and `list_users` have their own params structs. Happy to extend if wanted.

## Verification

- `cargo build -p sift_mcp`, `cargo test -p sift_mcp` (247 passed), `cargo clippy -p sift_mcp --all-targets` clean, `cargo fmt --check` clean.
- 13 new tests: projection helpers in `tool/common/test.rs` (key selection, naming conventions, unmatched reporting, per-item absence, non-object passthrough, empty list, body shape, counts) and end-to-end tool tests in `tool/channels/test.rs` and `tool/runs/test.rs`.
- Exercised end to end against a development environment: two calls, a 3,310-character response, 120 rows carrying only the `name` key, no truncation.
- Onboarding docs updated in `agents/mcp.md`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rust/crates/sift_mcp/src/tool/annotations/mod.rs
```diff
@@ -96,6 +97,9 @@ impl SiftMcpServer {
                 added `url` field with the annotation's Sift web link (`<host>/annotation/<annotation_id>`). `url`
                 is omitted when the host can't be derived. Surface these links to the user when presenting
                 annotations.
+              - `count`: how many items the response carries. Read this instead of
```
lineville: Does including `count` save some work? We're we making extra tool calls to get the counts before? I remember discussing with @evan-sift before about hitting the db for total counts of entities and that being prohibitively expensive from a db performance perspective, I don't think we're doing that here, but just trying to understand how we're getting counts now?

rowenwu: No DB work — `count` is `items.len()` on the vector we already built and are about to serialize. One field on a response we were sending anyway: no extra RPC, no `COUNT(*)`, no additional paging. Your and Evan's concern about total-entity counts doesn't apply here, because this deliberately isn't one.

On what it saves: no extra tool calls were happening. What was happening is the model doing the arithmetic itself and getting it wrong. Handed a complete 120-row response, an agent reported 122 — it estimated from the payload before enumerating, then talked itself out of its own correct list of 120 rather than trusting it. So the saving is a number the caller was getting wrong, not a round trip it was making.

Your question did surface a real gap, though, and I've pushed a doc fix for it. `count` is the size of *this page*, not the number of items matching the filter. Since `paging()` caps at `limit` (default 50, max 200) and the response says nothing about being capped, an agent could read `count: 50` off a truncated listing and report it as a total — the same fabrication one step removed. All eight descriptions now say `count` is not a match total, and that `count` reaching your `limit` means assume there are more.

That heuristic is weak — it's wrong in the exact-multiple case, which is when a confident wrong answer is most likely. The real fix is `has_more`, and the information already exists: at the `results.truncate(record_limit)` break the service knows whether the upstream page token was still non-empty. We just discard it. I kept it out of this PR because all 13 `list_*` services return `Result<Vec<T>>`, so threading a flag through touches every one plus their tests, and this PR is already carrying a projection, a count, and a merge with main. Filed separately; happy to pull it in here instead if you'd rather review it as one change.

rowenwu: Pulled it into this PR — `has_more` now ships alongside `count`.

Services return a `Page<T>` (items plus the flag) instead of a bare `Vec<T>`, so the value comes from the truncation the service actually performs rather than being inferred in the tool layer. That distinction matters: inferring it from `count == limit` is wrong precisely when the true total is a multiple of the limit, which is when a confident wrong answer is most likely.

The nice side effect is in `get_data`. It was already hand-rolling this guess to avoid writing a Parquet file with silently missing channels:

```rust
// A selection that fills the record cap may have been truncated…
if channels.len() >= common::PAGE_SIZE as usize {
```

That now reads `page.has_more`, which retires a false positive: a selection matching exactly the cap with nothing left upstream used to be rejected as incomplete when it was whole. Both directions have tests now — one for real truncation, one for the full-page-with-nothing-left case that should get through.

Scope: the eight list tools that carry `count`. The other five list services (`users`, `reports`, `report_templates`, `rule_versions`, `report_rule_summaries`) still return `Vec` and can adopt `Page` when they need it.

272 tests pass, clippy and fmt clean. The two "too many arguments" clippy warnings are pre-existing on main in `report_templates/mod.rs` and `lib.rs`.

lineville: I'm surprised it was getting the math wrong of just the items, but if this makes that computation more deterministic sounds good!

lineville (dismissed):

rust/crates/sift_mcp/src/tool/common/mod.rs
```diff
@@ -65,6 +66,112 @@ pub(crate) fn with_urls<T: Serialize>(
...
+    let (items, unmatched) = match fields.as_deref() {
+        Some(fields) if !fields.is_empty() => project_fields(items, fields),
+        _ => (items, Vec::new()),
+    };
+
+    let mut body = serde_json::Map::new();
+    body.insert("count".to_string(), Value::from(items.len()));
+    body.insert("has_more".to_string(), Value::Bool(has_more));
+    body.insert(key.to_string(), Value::Array(items));
+    if !unmatched.is_empty() {
```
lineville, line 181: An empty result will put all of the fields in `unmatched_fields`, which may lead to it misleading the agent into attempting to respell the field names. I think in `list_body` we want to skip filling in unmatched fields if we got an empty response so that an empty page doesn't get confused with a misspelling.

rust/crates/sift_mcp/src/tool/common/mod.rs
```diff
@@ -65,6 +66,112 @@ pub(crate) fn with_urls<T: Serialize>(
         .collect()
 }
 
+/// Normalize a field name for matching so `asset_id`, `assetId` and `AssetId`
+/// all address the same key. Callers carry names between tools that spell them
+/// differently, and a name that quietly matches nothing is the failure this
+/// projection exists to avoid.
+fn normalized(name: &str) -> String {
```
lineville, line 80: I think this might result in multiple source keys normalizing to the same key, which means the first key will always win, might be worth just documenting that at least if it's intended.

rust/crates/sift_mcp/src/tool/rules/mod.rs
```diff
@@ -84,6 +90,12 @@ impl SiftMcpServer {
                 Example: `\"created_date desc,modified_date\"`.
               - `limit`: max items to return. Start at 50 and only raise it if the result is capped
                 and you still need more. Values are clamped to `1..=200`; omitting it defaults to 50.
+              - `fields`: optional array of field names to keep on each item, e.g.
+                `[\"name\"]`. Omit it for the full object. Names match case-insensitively
+                and ignore underscores, so `asset_id` and `assetId` both work. Any name
```
lineville: it also ignores hyphens as well right? Can we update the handful of places we mention this to say hyphens as well

rust/crates/sift_mcp/src/tool/common/mod.rs
rowenwu: Good catch, and it's worse than misleading — I confirmed it with a throwaway test before fixing:

```
project_fields(Vec::new(), &["name", "channel_id"])
→ unmatched = ["name", "channel_id"]
```

A field only counted as matched if it appeared on some item, so with zero items nothing could match. An agent that filtered down to an empty result would be told its field names were wrong and would plausibly retry with different spellings, chasing a typo that was never there.

Fixed by returning early: nothing came back, so nothing can be said about the names, and `unmatched_fields` is omitted entirely. Two tests — an empty page reports none, and a genuine typo on a non-empty page still does, so the guard doesn't swallow the signal the field exists for.

rust/crates/sift_mcp/src/tool/common/mod.rs
rowenwu: Right, and it's worth writing down. `asset_id` and `assetId` both reduce to `assetid`, and the lookup takes the first match in key order, so one would silently win if an object ever carried both spellings.

Intended, in the sense that picking one deterministically beats erroring on an ambiguity the caller can't see. Documented on `normalized()` now, including that proto3 JSON emits one casing per field so our current shapes don't hit it — the note is there for whoever changes the serializer.

rust/crates/sift_mcp/src/tool/rules/mod.rs
rowenwu: It does, yes — `normalized()` filters both `_` and `-`, and every description said only underscores. Updated all eight to say underscores and hyphens with `asset-id` as a third example, plus the doc comment on the normalizer.

Added a test that walks `asset_id`, `asset-id`, `AssetId` and `ASSET_ID` against the same key so the claim in the docs is actually enforced.

rust/crates/sift_mcp/src/tool/common/mod.rs
lineville: Sounds good thanks for fixing!

lineville (approved):

PR #733, opened 2026-08-19
rust(chore): add arm64 as a build target for sift-cli

evan-sift:


rust/crates/sift_cli/Cargo.toml
```diff
@@ -3,7 +3,7 @@ test-reports = ["sift_mcp/test-reports"]
 
 [package]
 name = "sift_cli"
-version = "0.4.3"
+version = "0.4.4-alpha.1"
```
solidiquis: Does this need to be a patch bump given no code changes and no CHANGELOG changes? If you want to kick off a release without the tag then I think it'd be better to add the ability to manually dispatch a workflow to the release GHA config. That way you can just merge this in without patch bump then do a manual release.

solidiquis (changes requested):

Brandon-Shippy (dismissed):

rust/crates/sift_cli/Cargo.toml
evan-sift: will revert.

Brandon-Shippy (approved):

solidiquis (approved):

PR #734, opened 2026-08-19
rust(feat): agent install works without client binaries

evan-sift:
## Description

`sift-cli agent install` no longer requires the client binaries. A client's config directory (`~/.claude`, `~/.codex`) now counts as detection, the skill installs via plain file IO, and an MCP registration that needs the client's own CLI is skipped with a warning instead of failing the install. `agent doctor` reports an uninspectable registration as a warning, not an error. Conflicts still block the install as before.

Adds `--path <DIR>` to install the skill into a fixed directory, even with zero detected clients, for image builds and packaging.

## Verification

- `cargo test -p sift_cli`: 210 passed, including end-to-end tests for the no-binary install and both `--path` cases.
- `cargo clippy -p sift_cli --all-targets` and `cargo fmt --check` clean.
- Manual: a fresh HOME with only `.claude/` and no client binaries on PATH installs the skill, prints the skip warning, and exits 0.

### DIFF
--- a/rust/crates/sift_cli/CHANGELOG.md
+++ b/rust/crates/sift_cli/CHANGELOG.md
@@ -16,6 +16,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   a client-CLI-backed MCP registration that cannot run is skipped with a
   warning instead of failing the install. `agent doctor` reports such a
   registration as a warning rather than an error.
+- Added an optional `--path <DIR>` to `sift-cli agent install` that installs
+  the agent skill into the named directory in addition to every detected
+  client, for image builds and other environments that package the skill at a
+  fixed location. The directory receives `SKILL.md` and `references/`
+  directly; rerun with the same `--path` to refresh it.
 
 ## [v0.4.1] - August 10, 2026
 
--- a/rust/crates/sift_cli/CHANGELOG.md
+++ b/rust/crates/sift_cli/CHANGELOG.md
@@ -20,7 +20,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   the agent skill into the named directory in addition to every detected
   client, for image builds and other environments that package the skill at a
   fixed location. The directory receives `SKILL.md` and `references/`
-  directly; rerun with the same `--path` to refresh it.
+  directly; rerun with the same `--path` to refresh it. Without `--path`, the
+  skill installs only to the default locations for detected clients.
 
 ## [v0.4.1] - August 10, 2026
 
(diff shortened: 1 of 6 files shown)


rust/crates/sift_cli/CHANGELOG.md
```diff
@@ -7,6 +7,18 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### What's New
 
+- `sift-cli agent install` now works in headless environments (container
+  builds, CI). A client's config directory counts as detection even when its
+  binary is not on PATH, the skill always installs (it is plain file IO), and
+  a client-CLI-backed MCP registration that cannot run is skipped with a
+  warning instead of failing the install. `agent doctor` reports such a
+  registration as a warning rather than an error.
+- Added an optional `--path <DIR>` to `sift-cli agent install` that installs
```
lineville, line 19: Maybe we should mention what the default behavior will be when `path` isn't provided

lineville (dismissed):

rust/crates/sift_cli/src/cmd/agent/mod.rs
```diff
@@ -719,24 +724,43 @@ fn install_environment(
...
+    let mut skipped = Vec::new();
     for harness in &environment.harnesses {
         let state = config::inspect(*harness, environment)?;
         match state {
-            config::State::Conflict(detail) | config::State::Unavailable(detail) => {
+            config::State::Conflict(detail) => {
                 blockers.push(format!("{} MCP registration: {detail}", harness.label()));
             }
-            _ => {}
+            config::State::Unavailable(detail) => {
```
Brandon-Shippy: `Unavailable` is not just a missing binary. For example, `inspect_json` returns it for a malformed, unreadable, or symlinked `mcp.json`.

Skipping the variant will create some false successful states.

rust/crates/sift_cli/src/cmd/agent/mod.rs
```diff
@@ -705,8 +709,9 @@ fn install_environment(
     environment: &Environment,
     verb: &str,
     registration: &Registration,
+    extra_skill_path: Option<&Path>,
 ) -> Result<ExitCode> {
-    if environment.harnesses.is_empty() {
+    if environment.harnesses.is_empty() && extra_skill_path.is_none() {
```
Brandon-Shippy, line 736: It should skip an unregistrable client the same way install now does

rust/crates/sift_cli/src/cmd/agent/mod.rs
```diff
@@ -719,24 +724,43 @@ fn install_environment(
...
+        && !targets.iter().any(|target| target.path == path)
+    {
+        targets.push(skill::Target {
+            path: path.to_path_buf(),
+            harnesses: Vec::new(),
+        });
+    }
     let mut blockers = Vec::new();
     for target in &targets {
         if skill::inspect(&target.path)? == skill::State::Conflict {
```
Brandon-Shippy, line 762: `--path` rejects a directory that already exists and is empty, which is what an image build hands it; `skill::inspect` should return `Missing` for an empty directory rather than `Conflict`.

lineville (approved):

PR #736, opened 2026-08-19
rust(feat): calculated channel MCP tools

evan-sift:
## Description

Adds a `calculated_channels` domain to the MCP server.

Tool surface: `list_calculated_channels`, `list_calculated_channel_versions`, `create_calculated_channel`, `update_calculated_channel`, `archive_calculated_channel`, `unarchive_calculated_channel`. Both list tools carry the crate's `fields` projection and report `count` / `has_more`.

Nested channel references ride in `expression_channel_references_json`, a documented JSON string parameter, keeping every other parameter flat. Update pre-reads the channel before writing: the update mask paths cover whole configuration messages, so a mask-only write would clear sibling fields the caller never mentioned.

Surface note: there is no dedicated archive RPC. Archive and unarchive go through the update mask on `is_archived`, matching the Python client, so the server owns the archive timestamp. They are exposed as separate tools so each can carry a destructive write gate, and the reported outcome is derived from the state the server returns rather than asserted.

Stack note: the bundled agent skill and the 0.5.0 changelog for this whole stack land at the tip (#763).

## Verification

- `cargo test -p sift_mcp` on this branch: 332 passed, 0 failed (pagination, limit clamp, mask construction, archive/unarchive mask shape, state-mismatch reporting, write gates, JSON parameter validation, fields projection and count).
- `cargo build -p sift_mcp --no-default-features`, `cargo fmt --all -- --check`, `cargo clippy -p sift_mcp --all-targets`: clean.
- MCP protocol check against a development org (2026-08-18, before the archive-mask rework): all six tools registered, live `list_calculated_channels` round trip, and all four write tools reject with the remediation message when the server runs without `--allow-create` / `--allow-destructive`. The archive mechanism was since aligned with the Python client; its request shape is pinned by tests.
- Agent tool-selection gates (list, versions, create-blocked, create, archive) pass against the full stack binary.


rust/crates/sift_mcp/src/tool/calculated_channels/mod.rs
```diff
@@ -0,0 +1,753 @@
...
+                `asset_name`, `tag_id`, `tag_name`, `units`, `calculated_channel_version_id`, `created_date`,
+                `modified_date`, `created_by_user_id`, `modified_by_user_id`, `is_archived`, `archived_date`.
+                Folder membership is filterable via `folders` and `activeFolders` (folder-id lists;
+                `activeFolders` excludes archived folders): `\"<folder_id>\" in folders` returns calculated
+                channels in a folder, `size(activeFolders) == 0` returns uncategorized ones.
+                When filtering or searching, use `name.matches(\"(?i)thrust\")`, not `==`. Use `==` only for an
+                exact value from a prior result. `contains`/`startsWith`/`endsWith` are case-SENSITIVE:
+                `contains(\"Thrust\")` silently misses `thrust_margin`. Calculated channel names can embed `.`,
+                a regex wildcard, so match a full literal name with `contains`, not `matches`.
+              - `order_by`: optional comma-separated `FIELD_NAME[ desc]` list. Orderable fields: `created_date`,
```
lineville, line 105: The backend accepts only supports ordering by created_date and modified_date, so I suspect that calling order_by with name, description etc will fail

rust/crates/sift_mcp/src/tool/calculated_channels/mod.rs
```diff
@@ -0,0 +1,753 @@
...
+                `modified_date`, `name`, `description`, `units`, `archived_date`. Default sort is
```
lineville: I think the actual default on the backend is version ascending

lineville (commented):
A couple small inconsistencies with the backend but other than that looks good!

rust/crates/sift_mcp/src/tool/calculated_channels/mod.rs
evan-sift: Fixed: the order_by doc now lists created_date and modified_date only, matching GetOrderableColumnNames in the model. The proto comment itself overpromises the same fields; correction is in sift-stack/azimuth#13870.

rust/crates/sift_mcp/src/tool/calculated_channels/mod.rs
evan-sift: Fixed for the versions tool: orderable fields documented as created_date/modified_date/version with default version ascending. The channels-list default (created_date desc) matches the model's GetDefaultOrdering, so that claim stands. Proto comments corrected in sift-stack/azimuth#13870.

lineville (approved):

PR #737, opened 2026-08-19
rust(feat): MCP GA batch: saved calculated channels in get_data, preview_rule, UDF tools, bulk annotations, sift-cli 0.5.0

evan-sift:
## Description

This PR now carries the remainder of the MCP tool expansion: the reviewed stack #738, #739, #752, and #763 merged into it after #736 (calculated channel management tools) landed on main. Per-area details and review threads live in those PRs.

**Saved calculated channels in `get_data`** (this PR's original scope): a name in `channel_names` with no raw-channel match resolves as an active saved calculated channel for the asset and run, whether the caller passed `asset_name` or `asset_id`. Lookup is asset-scoped; ambiguous names are reported with the colliding ids instead of silently picking one, and partial resolution is explicit via `unresolved_calculated_channels`. A raw channel wins a shared name; `channel_regex` stays raw-only.

**`preview_rule`** (#738): a read-only dry run of a saved rule (`rule_id` or `rule_name`, active rules only) or an ad-hoc `draft_rule_config` against a run. Persists nothing, needs no write flags.

**User-defined function tools** (#739): `list_user_defined_functions`, `list_user_defined_function_versions`, `create_user_defined_function`, `update_user_defined_function`, `archive_user_defined_function`, `unarchive_user_defined_function`. Mask-only updates create a new version; renames apply alone; names are validated client-side with the backend's rule; `user_notes` are pre-read and carried through updates and archive flips so they are never wiped.

**Bulk annotation updates** (#752): `update_annotation` takes a required `annotation_ids` list (1 to 1000, breaking change from `annotation_id`) and applies the same change set to every id. Archiving uses one batch request; unarchiving clears each annotation's delete date through per-annotation updates with bounded fan-out. Per-id failures, `not_attempted` reporting, and `isError` on partial failure.

**Agent skill refresh and 0.5.0 release prep** (#763): the bundled skill's tool inventory mirrors the server registry (49 tools), workflow guidance covers everything above, and `sift_cli` is bumped to 0.5.0 with a CHANGELOG entry covering the batch, including `get_data.asset_id` from #757.

The new list tools carry the crate's `fields` projection and `count`/`has_more` convention throughout.

## Verification

- At this tree: `cargo test -p sift_mcp` 434 passed, `cargo test -p sift_cli cmd::agent` 38 passed (including the skill/reference-bundle map test), `cargo fmt --all -- --check` clean, `cargo build -p sift_mcp --no-default-features` clean, `cargo clippy -p sift_mcp --all-targets` no new warnings.
- Live end-to-end against a development org (2026-08-26): an MCP client drove the built server through every new tool family — reads, write gates without flags, creates/updates/archives with flags, calculated-channel resolution through `get_data` returning real data, and bulk annotation archive/unarchive verified in both directions.
- Inventory cross-check: skill tool list, `tool_events.json` (49 keys), and the server registry agree.
- Review: each constituent PR was reviewed independently; reviewer findings were fixed on the branches before the merges (see the threads in #736, #739, #763).


rust/crates/sift_mcp/src/service/calculated_channels/mod.rs
```diff
@@ -244,6 +284,183 @@ impl CalculatedChannelService {
...
 
+    /// Resolves saved calculated channels, named by the caller, into concrete
+    /// expressions for a single asset, optionally narrowed to a run. Each name
+    /// is looked up among the active saved channels and then resolved by the
+    /// API, which is what decides whether the channel applies to the asset.
+    ///
+    /// Names with no saved channel, and channels the API cannot resolve for the
+    /// asset, are returned in `unresolved` rather than dropped, so the caller
+    /// can name them instead of returning a partial result silently.
+    pub async fn resolve_calculated_channels(
```
lineville, line 295: An ABAC denial on a single channel would fail the entire get_data call, not necessarily the wrong behavior but just want to check if that's the desired behavior. If someone made a call and had access to 4/5 channels the whole call would fail when the backend sees there's a mismatch in the number in the request vs response.

lineville (dismissed):
Looks good! Just the one non-blocking question on partial failures.

rust/crates/sift_mcp/src/service/calculated_channels/mod.rs
evan-sift: that is desired, and I thiiink that ABAC shouldn't even let them see the channel that they're requesting in the first place

lineville (approved):

PR #739, opened 2026-08-19
rust(feat): user defined function MCP tools

evan-sift:
## Description

Adds a `user_defined_functions` domain to the MCP server.

Tool surface: `list_user_defined_functions`, `list_user_defined_function_versions`, `create_user_defined_function`, `update_user_defined_function`, `archive_user_defined_function`, `unarchive_user_defined_function`.

Function inputs ride in `function_inputs_json`, a documented JSON string parameter. Updates are mask-only and create a new version; a `name` change applies alone per the API's semantics, so the tool rejects `name` combined with other fields.

Surface note: there is no dedicated archive RPC. The public API archives a user-defined function by setting the archive field through the update mask. `archive_user_defined_function` and `unarchive_user_defined_function` are exposed as separate tools anyway, so an agent does not have to know that, and so each can carry a destructive write gate.

The list tools carry the crate's `fields` projection and report `count` / `has_more`.

Stack note: the bundled agent skill and the 0.5.0 changelog for this whole stack land at the tip (#763).

## Verification

- `cargo test -p sift_mcp` on this branch: 411 passed, 0 failed (pagination, mask construction, no stale version echo, archive/unarchive shape, write gates, input validation, fields projection and count).
- MCP protocol check against a development org: all six tools registered, live list round trip, all four write gates reject without their flags.
- Agent tool-selection gates (list, versions, create-blocked, create, archive) pass against the full stack binary.


rust/crates/sift_mcp/src/service/user_defined_functions/mod.rs
lineville, line 1: Every update and every archive/unarchive will silently clear the user_notes on the new version. The go handler reads in user_notes and replaces it and we're passing it an empty value so I think we'll end up unintentionally wiping out those notes. I think we can solve this by passing through user_notes

rust/crates/sift_mcp/src/tool/user_defined_functions/mod.rs
lineville, line 1: The naming rules here are less strict that the ones on the go backend which might cause some conflict, we might want to mirror what the backend has defined for name constraints.

lineville (changes requested):
Looking good mostly, I think the one issue to be fixed is with the user_notes being potentially overwritten with empty values.

rust/crates/sift_mcp/src/service/user_defined_functions/mod.rs
evan-sift: Fixed: update and archive/unarchive now pre-read the function and carry user_notes through, read-modify-write like the calculated-channel update. Tests assert the outgoing request carries the pre-read notes on both paths.

rust/crates/sift_mcp/src/tool/user_defined_functions/mod.rs
evan-sift: Fixed: create and rename validate client-side with the backend rule (starts with a letter, then alphanumerics or underscore, max 253). One thing this surfaced: the backend's InvalidName message says names must end with a letter, but the regex allows trailing digits and underscores — message aligned in sift-stack/azimuth#13870.

lineville (approved):
Looks good!

PR #740, opened 2026-08-19
rust(feat): campaign list and review MCP tools

evan-sift:
## Description

Adds a read-only `campaigns` domain: `list_campaigns` and `review_campaigns`.

Listing stays cheap: the outgoing request always skips report summaries, and a unit test pins that flag. `review_campaigns` is the expensive counterpart, fetching review summaries for named campaigns. `list_campaigns` exposes `include_archived` because the API excludes archived campaigns from queries entirely by default; without it they would be unreachable.

Surface note: `create_campaign` and `update_campaign` are not exposed in this change. Both RPCs exist and work. They are deferred until there is usage evidence that agents need to author campaigns rather than review them.

## Verification

- `cargo test -p sift_mcp`: 17 new mock-backed tests, including the `skip_report_summaries` assertion on the wire request and review request construction, full crate green.
- Live round trips against a development org: list and review both return real campaign data.
- Agent tool-selection gates pass: a listing prompt used `list_campaigns` only; a review prompt reached `review_campaigns`.


evan-sift:
Closing: campaign tools are cut from this MCP release. The stack now runs user-defined functions -> bulk annotations -> skill refresh.

PR #741, opened 2026-08-19
rust(feat): tag and metadata discovery MCP tools

evan-sift:
## Description

Adds read-only discovery tools for tags and metadata: `list_tags`, `list_metadata_keys`, `list_metadata_values`, `list_metadata_usage`.

Surface notes: tags v2 exposes only CreateTag and ListTags, so there is no update or archive to wrap, and a `create_tag` tool would be redundant — updating an asset or run with a new tag name creates the tag implicitly. Metadata has no update RPC; change is create plus archive, and those write tools are deferred as a scope decision. The metadata delete RPCs are deliberately not exposed: hard-deleting taxonomy in use is unrecoverable, and archive covers the workflow.

## Verification

- `cargo test -p sift_mcp`: 31 new mock-backed tests across both domains (pagination, limit clamp, filter and order_by forwarding, error mapping), full crate green.
- MCP protocol check against a development org: all four tools registered, live `list_tags` and `list_metadata_keys` round trips.
- Agent tool-selection gates pass for tags, keys, values, and usage prompts.


evan-sift:
Closing: tag and metadata discovery tools are cut from this MCP release. The stack now runs user-defined functions -> bulk annotations -> skill refresh.

PR #758, opened 2026-08-25
rust(bug): report channels missing from get_data output

rowenwu:
## What

`get_data` could write a Parquet file with fewer channels than were requested and still report a clean success. Three paths in:

1. **A channel name that matches nothing is dropped.** The selection filter is `name in [...]`, which matches what it can and says nothing about the rest. A typo returned a narrower table with no warning.
2. **A channel that matched but has no samples in the window is dropped.** It never reaches the column map, so it has no column at all rather than an all-null one. This is the common case, since it fires whenever a real channel is quiet in the queried window.
3. **When every matched channel is empty, the error counted them instead of naming them.** Finding the offender took a bisection.

All three share a root cause: the two facts needed to detect a gap live in different layers, and neither was carried across. The tool knows the names the caller asked for. The service knows which channels produced samples.

## How

`DataService::get_data` now returns `DataOutput { empty_channels }`, derived from the channel ids already present as `ColumnName` keys, so none of the nine per-type decode branches change. The no-data error names those channels, capped at 20 with an "and N more" tail so a 200-channel selection stays readable.

The tool diffs the requested names against what resolved and reports both gaps as `unmatched_channel_names` and `empty_channels`, omitted when empty, matching the `unmatched_fields` convention on the `list_*` tools.

Both also appear in `next_step`. That is the load-bearing half. The structured fields alone do not reliably stop a calling model from describing a partial fetch as complete, because a channel with no column is indistinguishable from one that was never requested.

## Two deliberate choices

**A partially empty request stays a success.** The data that exists is still written and useful, and a wide channel set over a narrow window is a legitimate query. The defect is the silence, not the status code.

**Empty channels are not emitted as all-null columns.** That would make the file self-describing, but a channel with zero samples has no observed type, so the column type would have to come from the request-side channel definition through a new mapping. That creates a second source of truth for column typing and changes file shape for anyone chaining `sql`.

## Testing

`cargo test -p sift_mcp --all-features`: 310 passing, up from 303. `cargo clippy --all-features --all-targets` clean.

New coverage:

- Service: an empty channel is named while the other returns data; nothing is reported when every channel has samples; the no-data error names every channel; the 20-name cap holds.
- Tool: unmatched names are reported and appear in `next_step`; a regex selection reports neither field, since it carries no per-name expectation; a matched-but-empty channel is reported.

Also verified against a live server over the MCP protocol.

Requesting four real channels plus one that does not exist:

```
unmatched_channel_names: ["<the bad name>"]
next_step: "...The file does NOT have a column for every channel requested.
            1 matched no channel on this asset: <the bad name>. Name those
            channels to the user before presenting any analysis..."
```

Every channel empty:

```
no channel data for given input parameters: no samples in the queried window for
LC-1, PT-PC, TC-NW; the channels exist, so widen the time range or drop the run
scope before concluding the asset has no data
```

Partially empty, reproducible from scratch: import a CSV with two channels whose samples cover non-overlapping halves of the time range, then request both over a window that only the first one covers.

```
empty_channels: ["CH_LATE"]
columns: [timestamp_unix_nanos, CH_EARLY]
```

Before this change all three returned success or a bare count, with no way to tell which channel was missing.

### End to end, with a model driving the tool

The unit tests prove the fields are populated. They cannot show whether a calling model acts on them, which is the point of the change, so I also drove the server over stdio with an LLM agent (Claude Opus 4.8) and asked it to report what it got back.

The agent was restricted to `get_data` alone. Given `list_channels` it can spot a bad name before ever calling `get_data`, and given `sql` it can find a missing column by querying the file and reading the schema error — both are sensible, and both would mask what the tool itself reports.

Two tasks, each run against the released 0.4.3 binary and a build of this branch:

1. Fetch three channels over a window where every real channel has data. One of the three names is a misspelling of a real channel.
2. Fetch two real channels over a window where only one of them has samples.

Against 0.4.3 the agent could not answer either, and was explicit about why:

> the tool response doesn't itself confirm which specific channels matched — it just reports success

> the tool response doesn't explicitly tell me which channel actually contains samples versus which came back empty

In the first case it guessed from the spelling that the name looked like a typo, which is inference from the string, not knowledge from the tool. In the second it offered to run a follow-up query to find out.

Against this branch both answers were correct and direct: the first named the misspelled channel as matching nothing on the asset, the second reported that one channel returned no samples in that window while the other returned data. Neither needed a follow-up call.

To reproduce the second case: import a CSV with two channels whose values occupy non-overlapping halves of the time range (a blank cell is not a sample), then request both channels over a window covering only the first half.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


lineville (dismissed):

rust/crates/sift_mcp/src/tool/data/mod.rs
```diff
@@ -257,23 +285,78 @@ impl SiftMcpServer {
...
-        let mut result = CallToolResult::structured(serde_json::json!({
-            "output": output_str,
-            "next_step": next_step,
-        }));
+        let mut body = serde_json::Map::new();
+        body.insert("output".to_string(), Value::from(output_str));
+        body.insert("next_step".to_string(), Value::from(next_step.clone()));
+        if !unmatched_channel_names.is_empty() {
+            body.insert(
+                "unmatched_channel_names".to_string(),
```
evan-sift: we should make `empty_channels` and `unmatched_channel_names` both stable fields that always appear in results, even if they're empty.

rust/crates/sift_mcp/src/tool/data/mod.rs
```diff
@@ -257,23 +285,78 @@ impl SiftMcpServer {
             .context("failed to open output parquet file")
             .map_err(from_anyhow)?;
 
-        self.data_service
+        let data_output = self
```
evan-sift: from local code review: 

```
The all-empty error discards unmatched names. For ["pressure", "presure"], assume pressure exists but has no samples. The service errors before the tool builds its structured body. The error names pressure, but loses the computed presure mismatch. Return a typed no-data error and merge both arrays into ErrorData.data. Add a combined-case tool test.
```

rust/crates/sift_mcp/src/service/data/mod.rs
```diff
@@ -835,6 +845,30 @@ impl DataService {
...
+            .iter()
+            .filter_map(|input| match input {
+                ChannelInput::Raw(channel) => {
+                    (!produced.contains(channel.channel_id.as_str())).then(|| channel.name.clone())
+                }
+                // Calculated channels are keyed by their channel key, not by an
+                // id the caller supplied, so fall back to matching on name.
+                ChannelInput::Calculation { name, .. } => {
+                    (!columns.keys().any(|column| column.name() == name)).then(|| name.clone())
+                }
```
evan-sift: from local code review:
```
Calculated-channel detection compares the requested key against ColumnName::name(). The data proto says calculated keys return in metadata.channel.channel_id; name is optional. Valid calculated data can therefore be reported as empty. Compare against ColumnName::channel_id() instead. No production caller currently constructs Calculation, which limits present impact.
```

evan-sift:
Couple things to fix, and need to resolve merge conflict. But pretty close

rust/crates/sift_mcp/src/tool/data/mod.rs
rowenwu: Done. Both are now unconditional in the success body, and the description says so: two empty arrays mean every requested channel is in the file.

Agreed on the reasoning — an absent key leaves "everything arrived" and "this tool never checked" looking identical, and the first is the assurance a caller needs before trusting the file. That is a different situation from `unmatched_fields` on the list tools, where silence costs nothing, so this deliberately diverges from that convention rather than following it.

Live check against dev, all channels present:

```
unmatched_channel_names: []
empty_channels: []
```

rust/crates/sift_mcp/src/tool/data/mod.rs
rowenwu: Good catch, and the scenario is exactly right — I confirmed the early return at the `get_data` call site discarded the names computed above it.

Worth spelling out why it bites: the caller is told only that `pressure` came back empty, so they widen the window, retry, and are still carrying the typo that nothing has ever mentioned. The one response that most needs both halves was the one carrying neither.

Fixed as you described. `NoChannelData` is now a typed error carrying the empty channel names, the tool downcasts it, and both arrays go into `ErrorData.data`. `empty_channels` appears only for the no-data error, since a transport failure never got far enough to know it; `unmatched_channel_names` comes from the request itself so it rides along on any failure.

Live, requesting `["PT-PC", "PT-PCC"]` over an empty window where only `PT-PC` exists:

```
code: -32603
message: ... no samples in the queried window for PT-PC ...
data: {"empty_channels": ["PT-PC"], "unmatched_channel_names": ["PT-PCC"]}
```

Combined-case tool test added.

rust/crates/sift_mcp/src/service/data/mod.rs
rowenwu: Confirmed against the proto before changing it — `protos/sift/data/v2/data.proto:149-153`:

```protobuf
message Channel {
  // For channel queries, this will contain the requested backing channel id.
  // For calculated channel queries, this will contain the requested channel key.
  string channel_id = 1 [(google.api.field_behavior) = REQUIRED];
  string name = 2;
}
```

`channel_id` is REQUIRED and carries the key; `name` has no such guarantee. Matching on `name` would report a calculated channel as empty even when it returned data.

Switched to `channel_id`. It also simplifies the code: both arms now look up the identifier the request sent, so the branch collapses to picking which value to report the miss under. Added a service test that returns a calculated page with `channel_id` set to the key and `name` empty, which is the shape that was failing.

Agreed on impact — nothing constructs `Calculation` today, so this was latent rather than live.

rowenwu:
All three addressed, and the conflict is resolved — I merged your `main` merge rather than force-pushing over it, so your resolution is intact.

- **Stable fields.** `empty_channels` and `unmatched_channel_names` always appear now.
- **All-empty error.** `NoChannelData` is a typed error; both arrays reach `ErrorData.data`, so a failed call no longer drops the unmatched names. Combined-case test added.
- **Calculated channels.** Now matched on `channel_id`, per `data.proto:149-153`. Test added with `name` empty, which is the shape that was failing.

314 tests passing, clippy clean. Re-verified against dev over the MCP protocol: both keys come back empty on a fully successful pull, and the combined case returns `data: {"empty_channels": ["PT-PC"], "unmatched_channel_names": ["PT-PCC"]}`.

evan-sift (approved):

PR #742, opened 2026-08-19
rust(docs): update the Sift agent skill for the new MCP tools

evan-sift:
## Description

Updates the bundled Sift agent skill for the new MCP tools. The skill's tool list now mirrors the server's registry exactly (55 tools, including the previously missing `ping`), and the workflow guidance covers the six new domains: calculated channel management, saved calculated channels through `get_data`, `preview_rule` dry runs, user-defined function discovery and versioning, campaign review cost, and tag/metadata discovery before adding taxonomy.

## Verification

- `cargo test -p sift_cli cmd::agent`: 32 passed, including the skill/reference-bundle map test.
- `sift-cli agent install` into a clean home: installed `SKILL.md` is byte-identical to the source.
- Inventory cross-check: the skill's tool list, `tool_events.json`, and the live server's `tools/list` agree.

PR #763, opened 2026-08-26
rust(chore): agent skill refresh and sift-cli 0.5.0 release prep

evan-sift:
## Description

Replaces #742, which GitHub's stack lock prevented from being retargeted after the stack restructure.

Two changes that close out the MCP tool expansion:

Refreshes the bundled Sift agent skill for the tools introduced below this branch in the stack. The skill's tool inventory now mirrors the server registry exactly (49 tools), and the workflow guidance covers calculated channel management, saved calculated channels through `get_data`, `preview_rule` dry runs, user-defined function discovery and versioning, and bulk `update_annotation` semantics (required `annotation_ids` list, per-id failure reporting).

Preps the sift-cli 0.5.0 release: version bump in `sift_cli/Cargo.toml` and a `CHANGELOG.md` entry covering the new tool surface, in the same format as the 0.4.4 entry.

## Verification

- `cargo test -p sift_cli cmd::agent`: 38 passed, 0 failed, including the skill/reference-bundle map test.
- Inventory cross-check: every tool named in the skill exists in the server registry and `tool_events.json` (49 keys); the only names in the skill that are not registry tools are deliberate "there is no such tool" statements.
- At this branch (the stack tip): `cargo test -p sift_mcp` 428 passed, `cargo test -p sift_cli` 218 passed, `cargo fmt --all -- --check` clean, `cargo build -p sift_mcp --no-default-features` clean.

### DIFF
--- a/rust/crates/sift_cli/assets/skills/sift/SKILL.md
+++ b/rust/crates/sift_cli/assets/skills/sift/SKILL.md
@@ -1,20 +1,19 @@
 ---
 name: sift
 description: >-
-  Use when working with Sift: ingesting or importing time-series data,
-  querying assets/runs/channels/users, managing calculated channels, rules,
-  user-defined functions, exporting data, decimating or running SQL over data,
-  opening a view in the Sift Explore web
-  app, writing code that integrates with Sift, installing, updating, or
-  diagnosing the Sift agent integration, or looking up how Sift works in its
-  product and API documentation. Covers the Sift MCP server (started by
-  `sift-cli mcp`), the `sift-cli` itself, the Sift REST API over cURL, the Sift
-  Python library (`sift_client`), and the Sift Rust streaming library
+  Use for Sift tasks: ingesting or importing time-series data, querying
+  assets/runs/channels/users, managing calculated channels, rules, and
+  user-defined functions, exporting or decimating data, running SQL over data,
+  opening a view in Sift Explore, writing code that integrates with Sift,
+  installing, updating, or diagnosing the Sift agent integration, or looking
+  up how Sift works in its product and API documentation. Covers the Sift MCP
+  server (started by `sift-cli mcp`), `sift-cli`, the Sift REST API over cURL,
+  the Sift Python library (`sift_client`), and the Sift Rust streaming library
   (`sift_stream`).
   Triggers include phrases like "import this file into Sift", "stream data to
   Sift", "list assets/runs/channels", "runs I created", "runs a teammate
-  created", "export a run", "query Sift", "graph", "plot", "visualize", "open
-  in Explore", "write code to integrate with Sift", "how does X work in Sift",
+  created", "export a run", "query Sift", "graph", "plot", "visualize", "open in
+  Explore", "write code to integrate with Sift", "how does X work in Sift",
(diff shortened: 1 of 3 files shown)


rust/crates/sift_cli/assets/skills/sift/SKILL.md
```diff
@@ -81,8 +96,39 @@ exists.
...
+- **Query a derived channel.** `get_data` serves saved calculated channels as
+  well as raw ones: name the calculated channel in `channel_names` and it is
+  evaluated for the requested asset and run. Confirm the name with
+  `list_calculated_channels` filtered on the asset first. A raw channel wins a
+  name it shares with a calculated channel, and `channel_regex` matches raw
+  channels only, so name calculated channels explicitly. When the result
+  carries `unresolved_calculated_channels`, the file is missing those columns:
+  tell the user which channels did not resolve rather than reporting the file as
+  complete.
+- **Author or change a calculated channel.** A calculated channel is a SEL
```
lineville: It's `CEL` expression not SEL right?

rust/crates/sift_cli/assets/skills/sift/SKILL.md
```diff
@@ -81,8 +96,39 @@ exists.
 - **Attribute something to a person.** Resolve the person with `list_users`,
   then filter another list on `created_by_user_id`. For "runs I created", pass
   `me: true`. Never guess which listed user is the caller.
+- **Update annotations.** `update_annotation` takes a required
```
lineville, line 98: I think that only field changes such as name, expression, or units insert a version, so this might lead to an agent telling someone that archive is a versioned entry that can be reversed which it isn't. It might be more clear to say that update creates a version, and that archive sets an archive state that unarchive_calculated_channel clears.

rust/crates/sift_cli/assets/skills/sift/SKILL.md
lineville, line 1: Skills should be no more than 1024 chars, some clients may truncate the skill now that it exceeds this limit. We should try to trim this if possible.

rust/crates/sift_cli/assets/skills/sift/SKILL.md
evan-sift: Reworded per your framing: update creates a new version and leaves earlier ones intact; archiving sets an archived state that unarchive_calculated_channel clears, and is not a versioned change.

rust/crates/sift_cli/assets/skills/sift/SKILL.md
evan-sift: Trimmed the frontmatter description from 1049 to 1022 characters (the 1024 limit applies to the description field; the file body has no such cap). Trigger phrases preserved.

rust/crates/sift_cli/assets/skills/sift/SKILL.md
lineville: I think after this one is changed, looks good will approve 👍 
```suggestion
- **Author or change a calculated channel.** A calculated channel is a CEL
```

evan-sift: ah historically we actually did call it SEL (sift expression language), but that's a remnant of the past. I think it might still be in the types though.

PR #764, opened 2026-08-26
rust(feat): gate MCP tools with account feature flags at startup

evan-sift:
## Description

The MCP server now resolves the account's enabled feature flags at startup, using the configured API key, and excludes tools whose flag is not enabled. Most tools carry no flag and are always registered.

- New `sift_mcp::FeatureFlags`: fetched once at startup via `GET /api/v1/feature-flags/variants` (5s timeout, bearer auth). A flag is enabled when its variant value is non-empty and not `off`.
- A static registry in `feature_flags.rs` maps tool names to required flags; gated tools are removed from the router before serving, so they are absent from `tools/list` and rejected on call.
- Fail-closed: if the fetch fails, the server starts normally with flag-gated tools disabled and logs a warning with the full error chain. Flag changes apply on MCP restart.
- The fetch is functional traffic and still runs under `--disable-nonessential-traffic` (help text updated).
- Removed the temporary Cargo feature that gated a tool domain at build time; those tools are now always built and gate at runtime instead.

## Verification

- `cargo test -p sift_mcp`: 311 passed. New coverage: wire-level fetch test (request line, bearer header, trailing-slash URI, 500 handling, response parse), registry drift guard, client-event invariant with all flags enabled, wire-level `tools/list` and `tools/call` for both flag states, fail-closed default behavior, enabled-semantics unit tests.
- `cargo check -p sift_cli`, `cargo fmt --all --check`, `cargo clippy --all-features` (no new warnings) all pass.

with ff off: 

image

with ff on:

image


rust/crates/sift_cli/src/cmd/mcp.rs
```diff
@@ -48,6 +48,15 @@ pub async fn run(ctx: Context, args: McpArgs, app_uri: String) -> Result<ExitCod
     if client_event_config.is_none() {
         tracing::info!("non-essential traffic is disabled");
     }
+    let feature_flags = sift_mcp::FeatureFlags::fetch(&ctx.rest_uri, &ctx.api_key)
```
lineville, line 51: This does add a blocking network call on startup. There's not a great way around this, we could do it in parallel but then they might kick things off without some tools that won't arrive until after this call. Not a big deal, but might be worth seeing if this adds any significant delays to start up.

lineville (dismissed):
Looks good, I think this might warrant a changelog entry as well but other than that lgtm!

rust/crates/sift_cli/src/cmd/mcp.rs
evan-sift: eh i think its fine to have this, because MCPs async start up in the background when CC / codex opens anyway

lineville: sounds good, yeah don't think it's a big concern. Feel free to resolve this and merge 🚢

lineville (approved):

PR #767, opened 2026-08-31
python(feat): list and get data imports

wei-qlu:
## Summary

Adds read access to data imports in `sift_client`

- `client.data_import.list_()`, `get()`, and `find()`, including a `DataImport` type with the import's status, error, and warnings.
- `run.data_imports`: a property that lists imports for a run directly.
- `job.get_data_import()`: get the import from the job `import_from_path` returns.

## Testing

- Unit tests 
- Manually listed imports by run, fetched by ID, and verified `run.data_imports` returns the same set.

### DIFF
--- a/python/lib/sift_client/sift_types/data_import.py
+++ b/python/lib/sift_client/sift_types/data_import.py
@@ -1027,6 +1027,7 @@ ImportConfig = Union[
 class DataImportStatus(str, Enum):
     """Status of a data import."""
 
+    UNSPECIFIED = "UNSPECIFIED"
     PENDING = "PENDING"
     IN_PROGRESS = "IN_PROGRESS"
     SUCCEEDED = "SUCCEEDED"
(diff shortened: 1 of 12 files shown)


python/lib/sift_client/sift_types/data_import.py
```diff
@@ -1016,3 +1022,92 @@ def _from_proto(cls, proto: UlogConfigProto) -> UlogImportConfig:
...
+    FAILED = "FAILED"
+
+    def to_filter_str(self) -> str:
+        """Convert to the string used in CEL filters."""
+        return f"DATA_IMPORT_STATUS_{self.value}"
+
+    @classmethod
+    def from_proto(cls, proto_value: int) -> DataImportStatus:
+        """Create from proto enum value."""
+        mapping: dict[int, DataImportStatus] = {
```
nathan-sift, line 1043: I'm thinking we might want to add an UNSPECIFIED enum here. It shouldn't normally happen, but technically it's a valid proto response. And for a function like list_(), right now that'll return a ValueError if any DataImport had this issue.

nathan-sift (changes requested):

python/lib/sift_client/sift_types/data_import.py
wei-qlu: Good catch, added unspecified and set that as the default [1c98112](https://github.com/sift-stack/sift/pull/767/commits/1c9811255f110eeb8967fd97c6a88bfbd7ca84b7)

nathan-sift (approved):
Looks good!

PR #766, opened 2026-08-28
rust(feat): artifact tools

lineville:
Adds three MCP tools backed by the public `sift.artifacts.v1.ArtifactService`:

- `list_artifacts`: lists artifacts in the organization, optionally scoped to one conversation.
- `download_artifact`: returns the latest version, or a pinned `artifact_version_id`, plus a signed `download_url` when the version has uploaded bytes.
- `create_artifact`: creates an artifact, or appends a version to an existing one. Artifacts cannot be edited in place; every change is a new version.

Access: creating needs `--allow-create`. Appending a version to an existing artifact also needs `--allow-destructive`.

**Feature flag:** these tools are enabled per account by the agents feature flag, resolved when the MCP server starts. This matches the backend gate on the artifact handler. When the flag is off, the tools are absent from the tool list and calls return "not found". Enabling the flag requires an MCP restart.

**Client events:** `tool_events.json` maps the three tools to `CLIENT_EVENT_USER_CALLED_MCP_TOOL_LIST_ARTIFACTS`, `_DOWNLOAD_ARTIFACT`, and `_CREATE_ARTIFACT`. These values are not yet in the azimuth `ClientEvent` enum, so the backend discards them until a follow-up azimuth PR adds them.

Also adds the `sift.artifacts.v1` proto, its generated Rust code, mocks for the artifact and remote-file services, and a test that each feature flag routes only its own tools.

https://claude.ai/code/session_01MMTsS9mCT21LBymq5Xmaex

PR #768, opened 2026-08-31
rust(feat): create_artifact uploads file content

lineville:
## Summary

`create_artifact` gains `file_path`. The MCP server streams the file to `POST /api/v0/remote-files/upload` (entity type `artifact_versions`, entity id = the new version), then returns the refreshed artifact with `file_name`, `file_mime_type`, `remote_file_id`, and a signed `download_url`.

- The multipart part declares a mime type from the extension via `mime_guess`, with `application/octet-stream` as the fallback. The server's own table misses `.md`, `.py`, and `.parquet`.
- Client checks: a regular non-empty file up to 1 GiB, name up to 255 bytes.
- A failed upload names the created version and tells the agent not to retry the create. The success note claims a preview only when the response carries a `download_url`.
- `RestConfig` is plumbed CLI → lib → server → `ArtifactService`. Uploads are essential traffic and ignore `--disable-nonessential-traffic`.
- rmcp 3.2.0: a legacy `initialize` naming `2026-07-28` is now answered with `2025-11-25`. Two handshake tests updated.

Closes #766. Consumed by sift-stack/azimuth#13949.

## Verification

`cargo test -p sift_mcp`: 516 pass. Clippy is clean apart from two pre-existing warnings. Built as 0.5.0 and exercised end to end in the azimuth kind cluster.


rust/crates/sift_cli/assets/skills/sift/SKILL.md
```diff
@@ -140,10 +140,13 @@ exists.
   Send a rename on its own. The API applies a `name` change by itself and
   ignores every other field, so `update_user_defined_function` rejects `name`
   combined with anything else.
-- **Create an artifact.** `create_artifact` with a `title` / `summary`.
-  Pass `conversation_id` to link it to a chat, and `authoring_kind=agent` when
+- **Create an artifact.** `create_artifact` with a `title` / `summary`, and
```
evan-sift, line 143: given that we're FF-ing specific tool calls, I wonder how we want to filter them in/out of the skill at all

rust/crates/sift_mcp/src/startup.rs
```diff
@@ -96,7 +96,9 @@ mod tests {
         let mut response = String::new();
         reader.read_line(&mut response).await.unwrap();
         let init_response: Value = serde_json::from_str(&response).unwrap();
-        assert_eq!(init_response["result"]["protocolVersion"], "2026-07-28");
+        // `2026-07-28` has no `initialize` handshake, so rmcp negotiates a
+        // client that names it there down to the newest legacy version.
+        assert_eq!(init_response["result"]["protocolVersion"], "2025-11-25");
```
evan-sift, line 101: I don't think we want to do this

evan-sift: is there anyway that we can get around needing to fall back to a lower version? is `initialize` necessary for artifacts?

rust/crates/sift_cli/assets/skills/sift/SKILL.md
lineville: Hmm good question, for now I think this is something we want included for everyone and not start filtering per org or anything like that.

rust/crates/sift_mcp/src/startup.rs
lineville: to be clear, initialize isn't specifically for artifacts, the downgrade resulted form upgrading to rmcp 3.2 not anything related to the artifacts implementation. Only the legacy lifecylcle would really use this, so I don't think we really need any changes here since we don't need to consider backwards compatibility

evan-sift: hmm ok. might be fine then 
CleanShot 2026-09-02 at 15 47 26

evan-sift (approved):

PR #769, opened 2026-08-31
python(feat): support enum_types for imports

wei-qlu:
### What was changed

Before: 
- `ValueError: "CsvDataColumn" object has no field "enum_types"`
- Setting a channel to be an enum type would succeed but the downstream data displayed nothing

Enum channels were only supported on TDMS imports. `enum_types` now lives on `DataColumnBase`, so all import config data columns (CSV, Parquet, HDF5, ULog) accept a `{name: key}` mapping.

- Add `enum_types` to `DataColumnBase` with a validator requiring `ChannelDataType.ENUM`, remove the duplicate field on `TdmsDataColumn`
- Build `ChannelConfigProto` via a shared helper in every `_to_proto`, and read enums back in every `_from_proto`

### Testing

Unit tests and manual imports with enum configs, validated the output

### DIFF
--- a/python/lib/sift_client/sift_types/data_import.py
+++ b/python/lib/sift_client/sift_types/data_import.py
@@ -156,11 +156,14 @@ class DataColumnBase(BaseModel, ABC):
 
     @model_validator(mode="after")
     def _check_enum_types(self) -> DataColumnBase:
-        if self.enum_types and self.data_type != ChannelDataType.ENUM:
-            raise ValueError(
-                f"'enum_types' requires data_type ChannelDataType.ENUM "
-                f"({self.name} is {self.data_type.name})."
-            )
+        if self.enum_types:
+            if self.data_type != ChannelDataType.ENUM:
+                raise ValueError(
+                    f"'enum_types' requires data_type ChannelDataType.ENUM "
+                    f"({self.name} is {self.data_type.name})."
+                )
+            if len(set(self.enum_types.values())) != len(self.enum_types):
+                raise ValueError(f"'enum_types' contains duplicate keys ({self.name}).")
         return self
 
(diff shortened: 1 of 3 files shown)


python/lib/sift_client/sift_types/data_import.py
```diff
@@ -144,12 +144,41 @@ class DataColumnBase(BaseModel, ABC):
...
     """
 
     name: str
     data_type: ChannelDataType
     units: str = ""
     description: str = ""
+    enum_types: dict[str, int] | None = None
+
+    @model_validator(mode="after")
+    def _check_enum_types(self) -> DataColumnBase:
```
alexluck-sift, line 158: Can we add validation to ensure there is no duplicate integers defined? a quick `len(list(enum_types.values())) == len(set(enum_types.values())`

wei-qlu: updated [2d1b9b1](https://github.com/sift-stack/sift/pull/769/commits/2d1b9b12167e27706fff0af0c3b24baafda9ce00)

python/lib/sift_client/sift_types/data_import.py
```diff
@@ -144,12 +144,44 @@ class DataColumnBase(BaseModel, ABC):
...
+        if self.enum_types:
```
alexluck-sift: shouldn't we instead check if type is enum OR enum types? Then if enum type, validate that the mapping is provided?

wei-qlu: That's right, now checks both ways [9722753](https://github.com/sift-stack/sift/pull/769/commits/9722753a4a3990ef1a606f114516499298bcdb0d)

alexluck-sift (approved):

PR #770, opened 2026-09-01
rust(chore): warn on import retries and document column flags

rowenwu:
## What

This updates ingest guidance in two places:

- The `sift-cli` skill explains that CSV column overrides are balanced sets of `-c/-d/-u/-n`, and that default inference produces doubles for numeric columns and strings for nonnumeric columns.
- The CLI retry warning is scoped to imports that create a named run with `--run`. It distinguishes local validation failures from failures after the request may have reached the server.
- The MCP `upload_dataset` tool warns callers that a request with `run_name` creates a new run and must not be retried automatically after an ambiguous failure.

## Why

A named ingest is not idempotent at the run-creation boundary. Repeating the same request can create another run with the same name, even when the samples themselves would be upserted into an existing run. Agents need to check for an exact run-name match on the asset before retrying a request that may have reached the server.

The previous CSV guidance also overstated default inference: nonnumeric values are inferred as strings, not doubles. Explicit column flags are still important because the channel data type cannot be changed after import.

The workspace also enables `tinyvec`'s `std` feature explicitly. `unicode-normalization` otherwise selects its `alloc`-only path, which does not compile in `tinyvec 1.13.0` and broke fresh dependency resolution in Rust CI.

## Reproducing

Call `upload_dataset` twice with the same `asset` and `run_name`: both uploads target the same asset, but each call creates a distinct run. The equivalent CLI case is importing twice with the same `--asset` and `--run`.

For CSV inference, preview or import a file containing both numeric and nonnumeric columns without column overrides. Numeric columns are inferred as doubles and nonnumeric columns as strings, with no inferred units or descriptions.

## Notes

The CLI guidance remains part of the embedded skill bundle. The MCP warning is part of the tool description returned to clients.

Validation:

- `cargo fmt --check`
- `cargo check --all-features`
- `cargo clippy --all-features`
- `cargo test --all-features`
- `cargo test --offline -p sift_cli cmd::agent` (38 passed)
- `cargo test --offline -p sift_mcp tool::data` (15 passed)


rust/crates/sift_cli/assets/skills/sift/references/cli.md
lineville, line 1: --data-type can accepts `infer`. Might be worth a mention here, since this text says the type must be right at import but never says -d infer is allowed.

lineville (dismissed):
One minor suggestion, otherwise LGTM!

lineville (approved):

PR #771, opened 2026-09-02
Python(fix): adjusting gRPC limits and page size

adamtoth-dev:
Aligning gRPC max received response limit with the backend. With the new limits, we can increase the page size on getData resulting in a faster data retrieval.

### DIFF
--- a/python/lib/sift_client/transport/grpc_transport.py
+++ b/python/lib/sift_client/transport/grpc_transport.py
@@ -16,10 +16,12 @@ from urllib.parse import urlparse
 
 from sift_client._internal.grpc_transport.transport import (
     DEFAULT_REQUEST_TIMEOUT_SECONDS,
-    MAX_DECODING_MESSAGE_SIZE,
     SiftChannelConfig,
     use_sift_async_channel,
 )
+
+MAX_DECODING_MESSAGE_SIZE = 50 * 1024 * 1024
+"""Largest gRPC message the client will accept or send, in bytes (50 MiB)."""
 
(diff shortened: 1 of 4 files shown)


python/lib/sift_client/transport/grpc_transport.py
```diff
@@ -48,6 +48,8 @@ def __init__(
         cert_via_openssl: bool = False,
         metadata: dict[str, str] | None = None,
         request_timeout: float | None = DEFAULT_REQUEST_TIMEOUT_SECONDS,
+        max_receive_message_length: int = 50 * 1024 * 1024,
```
alexluck-sift: Is it appropriate to set this here _and_ have `MAX_DECODING_MESSAGE_SIZE` defined in the internal grpc transport?

Either way, we should use constants as the default value rather than hard coding it inline for the arg default

adamtoth-dev: Good point, moved the value into a constant value.

alexluck-sift: I see we are including the current value in the comment, but this is ripe for going stale - since we have a constant (which has a doc string), can we rely on that instead? The docsite should support cross linking and whatnot to find what the value is easily.

alexluck-sift (approved):

python/lib/sift_client/transport/grpc_transport.py
adamtoth-dev: Updated and fixed.

PR #772, opened 2026-09-02
rust(fix): use stable sift-cli installer and restore update notices

evan-sift:
## Description

- Use the stable Sift API installer endpoints in CLI update messages and installation docs.
- Wait for the initial MCP release check before startup instructions are generated.
- Repeat available update notices in every session until the installed version is current.

## Verification

- `cargo test -p sift_cli` (261 passed)
- `cargo test -p sift_mcp` (508 passed)
- `cargo build -p sift_cli`
- `cargo fmt --all -- --check`
- Confirmed the shell and PowerShell installer endpoints return HTTP 200.


rust/crates/sift_cli/assets/docs/src/getting-started/installation.md
```diff
@@ -5,27 +5,35 @@ build it from source.
...
+version:
+
+```sh
+curl -fsSL https://api.siftstack.com/install/sift-cli/<version> | sh
+```
+
+```powershell
+irm https://api.siftstack.com/install/sift-cli.ps1/<version> | iex
+```
+
```
bijan-sift, line 35: An example could be useful here since it looks like the endpoint needs to be `x.y.z` instead of `vx.y.z` or `sift_clivx.y.z`

bijan-sift (dismissed):
LGTM, just added a suggestion. Also not sure about those failing tests, looks unrelated...

lineville (dismissed):

rust/crates/sift_cli/assets/docs/src/getting-started/installation.md
evan-sift: fair feedback, I'll include that change.

bijan-sift (approved):

PR #776, opened 2026-09-03
Python(fix): cache boundary resolution to include ns

alexluck-sift:
- The channel data cache stored segment bounds as `datetime`, truncating the nanosecond timestamps that getData supports. Every get data produced a `UserWarning: Discarding nonzero nanoseconds in conversion` once per channel, and then incorrectly truncated the precision when stored in cache. Bounds are now `pd.Timestamp`, and `get_data` / `get_data_as_arrow` accept `pd.Timestamp` bounds that keep their nanoseconds through to the request.
- Folds two duplicate protobuf converters into `to_timestamp_pb`, so the export, run, rule and tag paths stop silently flooring to microseconds.


python/lib/sift_py/data/service.py
```diff
@@ -22,7 +22,7 @@
 
 from sift_py._internal.cel import cel_in
 from sift_py._internal.channel import channel_fqn
-from sift_py._internal.convert.timestamp import to_pb_timestamp
+from sift_py._internal.time import to_timestamp_pb
```
ian-sift: sift_py's version to_timestamp_pb  is missing the Pd.Timestamp branch

alexluck-sift: removed this change entirely since sift_py is no longer being maintained

ian-sift (approved):

PR #779, opened 2026-09-04
rust(feat): artifact tools for the generic artifact store

evan-sift:
## Description

Extends the artifact MCP tools for the generic artifact store exposed by `sift.artifacts.v1`.

- Syncs the updated `artifacts.proto` (storage class, created-via surface, JSON payload, metadata, versioned links, CEL `filter` and `order_by`) and regenerates the Rust bindings. The public proto carries the stable surface; the link and discovery RPCs stay internal for now.
- `list_artifacts` accepts `filter` (CEL), `order_by`, and `include_archived`, and documents the filterable fields, enum comparisons, `links.exists(...)` and `metadata["<key>"]` forms. Results are scoped to the caller's organization.
- `create_artifact` accepts `storage_class` (`file` | `structured` | `blob`), `created_via` (`agent` | `canvas` | `upload`), a structured JSON `payload`, `metadata`, and `links`. `created_via` is required by the server and defaults to `agent`, since the tool runs inside agent sessions. Local validation rejects payload/storage-class mismatches before any request. `storage_class` stays unset on append so existing artifacts keep their class.
- `list_artifact_versions` returns one artifact's history, newest first. It takes no `filter` or `order_by` because the RPC has neither and fixes the order, and it omits `payload` — read a version's body with `download_artifact`.
- `archive_artifact` and `unarchive_artifact` hide and restore an artifact without touching its versions, links, or files. Both need `--allow-destructive`.
- `download_artifact` documents that structured artifacts carry their payload inline.

Every artifact tool stays behind the agents feature flag resolved at server start. Existing callers of `list_artifacts`, `create_artifact`, and `download_artifact` keep working unchanged.

Depends on the matching server change to `sift.artifacts.v1` landing first.

## Verification

Drove the MCP server over stdio against a live deployment: created a throwaway artifact, appended a version, listed its history newest-first, then archived and unarchived it. Archiving drops it from default listings, `include_archived` finds it, and its versions survive. Without `--allow-destructive` both writes refuse.

`cargo fmt --all --check` and `cargo clippy -p sift_mcp -p sift_cli --all-targets` are clean apart from a pre-existing `too_many_arguments` warning.


rust/crates/sift_mcp/src/service/artifacts/mod.rs
lineville, line 1: list_artifacts claims that payload is included, which it isn't that description should be updated.

lineville, line 1: Did we want `filter` to be required now? That will result in an error if a caller doesn't provide a filter.

lineville, line 1: an agent that appends a new title with no payload passes local validation and fails at the server, we probably need to sync up that validation logic.

protos/sift/artifacts/v1/artifacts.proto
```diff
@@ -124,123 +129,229 @@ enum ArtifactAuthoringKind {
...
+  optional string title = 4 [(google.api.field_behavior) = OUTPUT_ONLY];
+  optional string summary = 5 [(google.api.field_behavior) = OUTPUT_ONLY];
+  optional string authoring_message_id = 6 [(google.api.field_behavior) = OUTPUT_ONLY];
+  repeated string source_tool_use_ids = 7 [(google.api.field_behavior) = OUTPUT_ONLY];
+  optional string remote_file_id = 8 [(google.api.field_behavior) = OUTPUT_ONLY];
+  google.protobuf.Timestamp created_date = 9 [(google.api.field_behavior) = OUTPUT_ONLY];
+  // From the version's remote_files record. Unset until bytes are uploaded.
+  optional string file_name = 10 [(google.api.field_behavior) = OUTPUT_ONLY];
+  optional string file_mime_type = 11 [(google.api.field_behavior) = OUTPUT_ONLY];
+  // Never set: ListArtifactVersions omits it. Read a STRUCTURED version's
```
lineville, line 225: The 'never set' bit conflicts with what the azimuth repo says which is 'set by GetArtifact'

rust/crates/sift_mcp/src/service/artifacts/mod.rs
evan-sift: Fixed in 63a540a9a.

evan-sift: Fixed in 63a540a9a.

protos/sift/artifacts/v1/artifacts.proto
evan-sift: Fixed in 63a540a9a.

rust/crates/sift_mcp/src/service/artifacts/mod.rs
evan-sift: yes, filter is required. It can pass in `""` for it to list everything, and the tool already describe that option

lineville (approved):

PR #781, opened 2026-09-04
Feat: Genericize the sift-cli config to be used by Python

alexluck-sift:
`SiftClient` now reads the same `sift.toml` profiles that `sift-cli --profile`
uses. If you configure an environment once for the CLI, the Python client uses
it with no arguments.

```python
client = SiftClient()                    # the CLI's default profile
client = SiftClient(profile="staging")   # a named profile
```

The pytest plugin gains `--sift-profile`, the `sift_profile` ini key, and
`SIFT_PROFILE`.

## Precedence

Highest first:

1. Arguments.
2. A profile that you name in code.
3. The `SIFT_*` environment variables.
4. The profile that `SIFT_PROFILE` names.
5. The config file's default table.

`client.credential_sources` reports the layer that supplied each value.

The pytest plugin reverses this. There the profile ranks below the plugin's own
surfaces, so a profile on the runner cannot replace a key that CI injects.

## Behavior changes

- An `http://` URL now connects without TLS instead of failing, because
  transport security follows the scheme of the gRPC URL. This also affects the
  explicit-argument path, but only for `http://`, which never worked before.
- `api_key` is now the canonical config key. `apikey` stays valid, so you do
  not need to migrate.

### DIFF
--- a/python/CHANGELOG.md
+++ b/python/CHANGELOG.md
@@ -7,6 +7,22 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### What's New
 
+#### Credentials from sift-cli profiles
+
+`SiftClient` now reads the same `sift.toml` profiles that `sift-cli --profile` uses, so an environment configured once for the CLI works from Python with no arguments.
+
+```python
+client = SiftClient()                      # the CLI's default profile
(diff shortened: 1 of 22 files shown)


wei-qlu (approved):
lgtm, nice QoL improvements

PR #789, opened 2026-09-14
python(feat): webhooks resource

Brandon-Shippy:
# Description
Adds a webhooks resource to `sift_client`, and lets a rule call a webhook when it is violated.

  - **`WebhooksAPIAsync`** with `get`, `list_`, `find`, `create`, `update`, `archive`, `unarchive`, `test`. Wired up as `client.webhooks` and `client.async_.webhooks`
  - **`WebhooksLowLevelClient`** wraps `WebhookService`; the high-level client has zero proto imports
  - **Pydantic models** in `sift_types/webhook.py`: `Webhook`, `WebhookCreate`, `WebhookUpdate`, plus `WebhookEventType`, `WebhookHttpHeader`, `WebhookTestResult`
  - **`RuleAction.webhook(...)`** attaches a webhook to a rule. `RuleAction` previously only expressed annotation actions, so a webhook could be created but never wired to anything
  - `http_headers` accepts a list of `WebhookHttpHeader` or a `{name: value}` mapping; `update` replaces the list rather than merging it
  - `test()` sends a real POST to `target_url` and returns the response
  - `update_webhook` fetches before it writes — `UpdateWebhook` validates the whole `Webhook` message, not just the masked fields, so it rejects a sparse patch.
 
# Validation
- Full pipeline tested against a local stack: created a webhook and a rule with a webhook action, streamed points breaking the rule, and confirmed the delivery in the receiving server


python/lib/sift_client/sift_types/rule.py
alexluck-sift, line 277: This is a pretty inconvenient type to work with and doesn't follow the user-friendly patterns we do with other types. That is using the to_proto to convert to the proto friendly Ids. Can we address this as par tof this PR instead of carrying the tech debt forward?

Brandon-Shippy: https://github.com/sift-stack/sift/pull/789/commits/2baa0a690945f535b667943742e2916e0ab915ff

python/lib/sift_client/sift_types/_refs.py
```diff
@@ -0,0 +1,62 @@
+"""References to other Sift resources.
```
wei-qlu: `Ref[T]` is just `str | T` with a new name to look up. Let's write the union inline like the other models do and drop this

python/lib/sift_client/sift_types/rule.py
```diff
@@ -280,34 +285,38 @@ class RuleAction(BaseType[RuleActionProto, "RuleAction"]):
...
+    def for_webhook(cls, webhook: Ref[Webhook]) -> RuleAction:
+        """Create a webhook action.
+
+        Args:
+            webhook: The Webhook or webhook ID to call when the rule is violated.
+        """
+        return cls(action_type=RuleActionType.WEBHOOK, webhook=webhook)
+
+    @classmethod
+    def for_annotation(
```
wei-qlu: Renaming `RuleAction.annotation` breaks existing callers. Let's keep it as is and name the new constructor `RuleAction.webhook`

in general, renaming existing fields is not backwards compatible and can cause unintentional/breaking changes (e.g., `rule.action.tags` now returns something else

python/CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### What's New
+
+#### Webhooks API
+
+New `client.webhooks` resource, and `RuleAction.for_webhook(...)` to attach a webhook to a rule.
```
wei-qlu: make sure the changelog is updated to reflect any changes

wei-qlu (changes requested):

python/lib/sift_client/sift_types/rule.py
wei-qlu: pre-existing, but lets allow asset objects here too (`list[str | Asset]`) so callers don't need `asset._id_or_error`

wei-qlu (changes requested):

python/lib/sift_client/sift_types/_refs.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/789/commits/a6ddb678645b4674553366c978159079d6667e80

python/lib/sift_client/sift_types/rule.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/789/commits/a6ddb678645b4674553366c978159079d6667e80

python/CHANGELOG.md
Brandon-Shippy: https://github.com/sift-stack/sift/pull/789/commits/a6ddb678645b4674553366c978159079d6667e80

python/lib/sift_client/sift_types/rule.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/789/commits/a6ddb678645b4674553366c978159079d6667e80

python/lib/sift_client/sift_types/rule.py
```diff
@@ -282,6 +290,17 @@ class RuleAction(BaseType[RuleActionProto, "RuleAction"]):
     annotation_type: RuleAnnotationType | None = None
     tags_ids: list[str] | None = None
     default_assignee_user: str | None = None
+    webhook_id: str | None = None
```
alexluck-sift: `RuleAction.webhook(wh)` accepts an object, but the field it writes only takes a string, so anyone building a `RuleAction` directly is back to `webhook._id_or_error`. That's the thing my rule.py:277 thread asked to remove, and `asset_ids` gained `list[str | Asset]` 113 lines up in this same commit.

I'd take `str | Webhook` here and resolve it in `_to_update_request`. Worth settling with @wei-qlu first, since dropping the `Ref` alias was right but it took this with it. Same applies to `tags_ids` and `default_assignee_user` above, which are pre-existing so out of scope for this PR.

alexluck-sift (commented):
Audit pass on the webhooks surface. Nine comments, five of them nits. The first one is the only thing I'd call blocking, and it's a reviewer conflict rather than anything you did wrong.

python/lib/sift_client/sift_types/webhook.py
```diff
@@ -0,0 +1,199 @@
...
+
+    # Required fields
+    name: str
+    organization_id: str
+    target_url: str
+    event_type: WebhookEventType
+    http_headers: list[WebhookHttpHeader]
+    created_date: datetime
+    modified_date: datetime
+    created_by_user_id: str
```
alexluck-sift, line 80: `created_by_user_id`, `modified_by_user_id` and `organization_id` have no accessor, so finding out who made a webhook is a second call the caller assembles. `Rule.assets` and `Channel.asset` are the shape: an ID field plus a property that resolves it.

Add `created_by` and `modified_by` backed by `client.users`. Heads up that the seven existing `created_by`/`organization` properties in `sift_types/` all raise `NotImplementedError`, written before the Users resource existed, so they look like precedent and aren't.

python/lib/sift_client/resources/webhooks.py
```diff
@@ -0,0 +1,213 @@
...
+        Args:
+            webhook_id: The ID of the webhook.
+
+        Returns:
+            The Webhook.
+        """
+        webhook = await self._low_level_client.get_webhook(webhook_id=webhook_id)
+        return self._apply_client_to_instance(webhook)
+
+    async def list_(
```
alexluck-sift, line 49: No date ranges and no `created_by`/`modified_by`, so "webhooks added last week" isn't expressible. `ListWebhooksRequest` only filters on `webhook_id`, `name`, `event_type`, `is_archived` and `archived_date`, and CONTRIBUTING says a proto that can't serve the standard filter set should be changed rather than worked around.

Add `created_date`, `modified_date`, `created_by_user_id` and `modified_by_user_id` to the `ListWebhooks` filter fields, then expose the standard arguments here.

python/lib/sift_client/_internal/low_level_wrappers/webhooks.py
```diff
@@ -0,0 +1,194 @@
...
+
+        Returns:
+            The created Webhook.
+        """
+        request_proto = create.to_proto()
+        response = await self._grpc_client.get_stub(WebhookServiceStub).CreateWebhook(request_proto)
+        grpc_webhook = cast("CreateWebhookResponse", response).webhook
+        return Webhook._from_proto(grpc_webhook)
+
+    async def update_webhook(self, update: WebhookUpdate) -> Webhook:
```
alexluck-sift, line 143: Two concurrent `webhooks.update` calls each send a full `Webhook` here, so the second one reverts the first one's field. This is the only `MergeMessage` in `low_level_wrappers/`, and the proto declares a sparse masked patch with seven maskable fields named.

I'd send the patch and mask straight through. If the service genuinely rejects a sparse `Webhook`, file that as a server bug and link it from here, because the workaround costs a round trip and loses concurrent writes.

python/lib/sift_client/resources/webhooks.py
```diff
@@ -0,0 +1,213 @@
...
+
+        Args:
+            webhook: The Webhook or webhook ID to unarchive.
+
+        Returns:
+            The unarchived Webhook.
+        """
+        return await self.update(webhook, WebhookUpdate(is_archived=False))
+
+    async def test(
```
alexluck-sift: nit: this sends a real HTTP request to a customer endpoint, and `test` reads like the opposite to anyone scanning autocomplete, on top of colliding with test-suite vocabulary. `webhook` is also optional and positional, which fixes the parameter position permanently.

I'd go with `send_test_request` and move `webhook` behind the `*`. The rename touches `sift_types/webhook.py` and the sync stubs, so it needs doing by hand.

python/lib/sift_client/sift_types/webhook.py
```diff
@@ -0,0 +1,199 @@
...
+)
+
+if TYPE_CHECKING:
+    from sift_client.client import SiftClient
+
+
+class WebhookEventType(Enum):
+    """Enum for the events that trigger a webhook."""
+
+    UNSPECIFIED = WebhookEventTypeProto.WEBHOOK_EVENT_TYPE_UNSPECIFIED  # 0
```
alexluck-sift: nit: `WEBHOOK_EVENT_TYPE_UNSPECIFIED` is the proto's zero value and never something a caller should pass. No other enum in `sift_types/` exposes its unspecified member.

Drop it and map an unset `event_type` to `None` in `_from_proto`.

python/lib/sift_client/sift_types/webhook.py
```diff
@@ -0,0 +1,199 @@
...
+class WebhookTestResult(BaseModel):
+    """The response the target URL returned for a test request.
+
+    Attributes:
+        http_response_code: The HTTP status code the target returned.
+        http_response_body: The raw response body the target returned.
+    """
+
+    http_response_code: int
+    http_response_body: bytes
```
alexluck-sift, line 61: nit: `bytes` is the right call for the field, since the body comes from an arbitrary endpoint and may not be text at all, but it leaves every caller writing the same decode.

Add a `text` property returning `self.http_response_body.decode(errors="replace")`, matching `AnnotationCommentElement.text` in #792.

python/lib/sift_client/sift_types/webhook.py
```diff
@@ -0,0 +1,199 @@
...
+
+    name: str
+    value: str
+
+    @classmethod
+    def _from_proto(cls, proto: WebhookHttpHeaderProto) -> WebhookHttpHeader:
+        return cls(name=proto.name, value=proto.value)
+
+
+class WebhookTestResult(BaseModel):
```
alexluck-sift: nit: this is the library's first struct returned from a method, and #793 is adding a second one the same week with a different suffix, `UserDefinedFunctionValidation`.

No strong view on which wins, but the two PRs should agree. If `*Result` stays here, #793 wants `UserDefinedFunctionValidationResult`.

python/lib/sift_client/sift_types/webhook.py
```diff
@@ -0,0 +1,199 @@
...
+    @field_validator("http_headers", mode="before")
+    @classmethod
+    def _normalize_http_headers(cls, value):
+        """Accept a ``{name: value}`` mapping in place of a list of headers."""
+        if isinstance(value, dict):
+            return [WebhookHttpHeader(name=name, value=header) for name, header in value.items()]
+        return value
+
+    @model_validator(mode="after")
+    def _validate_target_url(self):
```
alexluck-sift: nit: this validator sits on `WebhookBase`, which doesn't declare `target_url`, so it reads through `getattr(self, "target_url", None)`. Rename the field on either subclass and validation silently passes everything, with a bad URL only failing at the server.

Declare `target_url: str | None = None` on the base so the attribute access is checked.

python/lib/sift_client/sift_types/rule.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/789/commits/38c50f7078d656d88817f257e29b2231638ae2be

`RuleAction.webhook_id` now just takes `str | Webhook`

python/lib/sift_client/sift_types/webhook.py
Brandon-Shippy: added `Webhook.created_by` and `modified_by` https://github.com/sift-stack/sift/pull/789/commits/38c50f7078d656d88817f257e29b2231638ae2be

python/lib/sift_client/resources/webhooks.py
Brandon-Shippy: renamed `test` to `send_test_request` https://github.com/sift-stack/sift/pull/789/commits/38c50f7078d656d88817f257e29b2231638ae2be

python/lib/sift_client/sift_types/webhook.py
Brandon-Shippy: dropped and maps properly now https://github.com/sift-stack/sift/pull/789/commits/38c50f7078d656d88817f257e29b2231638ae2be

python/lib/sift_client/sift_types/webhook.py
Brandon-Shippy: added `WebhookTestResult.text` https://github.com/sift-stack/sift/pull/789/commits/38c50f7078d656d88817f257e29b2231638ae2be

python/lib/sift_client/sift_types/webhook.py
Brandon-Shippy: the URL check is `field_validator` on each subclass

python/lib/sift_client/sift_types/_refs.py
Brandon-Shippy: i removed it union is now inline

python/lib/sift_client/sift_types/rule.py
Brandon-Shippy: I kept `RuleAction.annotation`, `tags_ids`, `default_assignee_user`, and the tags property

python/CHANGELOG.md
Brandon-Shippy: added logs for user-visible changes

python/lib/sift_client/sift_types/rule.py
Brandon-Shippy: `asset_ids` now accepts `list[str | Asset]`

python/lib/sift_client/sift_types/webhook.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/789/commits/dc8d34e3b436ba96579f45791ffab464f1c4be12

Decided on dropping the suffix

python/lib/sift_client/_internal/low_level_wrappers/webhooks.py
Brandon-Shippy: Server-side bug. Making a ticket and will triage. UpdateWebhook cannot accept a partial webhook. So, it reads the webhook before writing the update. Concurrent webhooks will overwrite each other.

python/lib/sift_client/resources/webhooks.py
Brandon-Shippy: ListWebhooks only supports `webhook_id`, `name`, `event_type`, `is_archived`, `archived_date`. Cannot filter this currently. I will make a ticket to fix this

PR #790, opened 2026-09-15
rust(feat): sift_mcp presents Sift entities as named Markdown links

lineville:
## Description

Companion to azimuth ENG-14734. The chat renders a Markdown link to an entity page as a chip with the link text as its label, so tool output must hand the model a named link.

- `url_clause` takes the entity kind and name and asks for `[name](url)`. Without a name it asks the model to use the entity's name, never a bare id or "link".
- `upload_dataset` returns `asset_url` and `run_url` and names both in `next_step`. Before, it returned bare ids with no link.
- `explore_url` asks for descriptive link text.
- List tool descriptions say to present each row as a named link with `url` as the target.
- Changelog entry under Unreleased.

## Verification

- `cargo test -p sift_mcp`: 528 passed.
- `cargo clippy` and `cargo fmt` clean for the crate; remaining warnings are pre-existing.


rust/crates/sift_mcp/src/tool/rules/mod.rs
```diff
@@ -61,8 +61,8 @@ impl SiftMcpServer {
                 `archived_date`, `is_live_evaluation_enabled`, `organization_id`, `conditions`, `rule_version`,
                 `current_version_id`, `asset_configuration` (asset_ids + tag_ids), `contextual_channels`,
                 `metadata`, and timestamps, plus an added `url` field with the rule's Sift web link
-                (`<host>/rules/<rule_id>`). `url` is omitted when the host can't be derived. Surface these links
-                to the user when presenting rules.
+                (`<host>/rules/<rule_id>`). `url` is omitted when the host can't be derived. Present each of these rules to the user as a Markdown link with its name as the text and `url` as the
+                target.
```
evan-sift: [nit] line length spacing could be cleaned up a bit in this PR, runs/mod.rs and reports/mod.rs has the same thing too

rust/crates/sift_mcp/src/tool/common/mod.rs
```diff
@@ -39,9 +39,22 @@ pub struct MetadataEntry {
...
 /// not be derived (e.g. self-hosted deployments without an `api.` subdomain) —
 /// so URL derivation never fails an operation.
-pub(crate) fn url_clause(url: Option<&str>) -> String {
-    url.map(|u| format!(" View it in Sift: {u}"))
-        .unwrap_or_default()
+/// The sentence that tells the model how to present an entity's Sift web link.
+/// The Sift chat renders a Markdown link to an entity page as a chip whose text
+/// is the link text, so the text must be the entity's own name: `[name](url)`.
+/// When the caller has no name in scope, the model is told to supply it. Empty
+/// when there is no link to present (host not derivable).
```
evan-sift: the doubled up comments here might be weird

lineville: done in [3db33e1](https://github.com/sift-stack/sift/pull/790/commits/3db33e1e350d20b3a263e1bc2bcc05bb46410d86)

Btw, I did test that links still work fine in the CLI. I think we could add a `--link-style markdown|plain` option, but I think I'll leave that out of this

rust/crates/sift_mcp/src/tool/rules/mod.rs
lineville: done in [3db33e1](https://github.com/sift-stack/sift/pull/790/commits/3db33e1e350d20b3a263e1bc2bcc05bb46410d86)

evan-sift (approved):

PR #792, opened 2026-09-17
python(feat): annotations resource

Brandon-Shippy:
# Description
  Adds an annotations resource to `sift_client`, covering data reviews, phases, and the review history.

  - **`AnnotationsAPIAsync`** with the usual `get`, `list_`, `find`, `create`, `update`, `archive`, `unarchive`, plus batch archive. Wired up as `client.annotations`
  - **`AnnotationLogsAPIAsync`** nested as `client.annotations.logs`
  - **`AnnotationsLowLevelClient`** wraps `AnnotationService`; the high-level client has zero proto imports
  - **Pydantic models** in `sift_types/annotation.py`
  - **`create_review(...)` and `create_phase(...)`** alongside `create`. Passing `channels` derives the asset
  - **`resolve()`, `flag()`, `reopen()`, `assign()`, `comment()`** on the resource and the `Annotation`

  # Validation
  - Verified against a local stack, including the annotation rendering in Explore


python/lib/sift_client/resources/annotations.py
```diff
@@ -0,0 +1,657 @@
...
+            ),
+        ]
+        if annotation_ids:
+            filter_parts.append(cel.in_("annotation_id", annotation_ids))
+        if annotation_type:
+            filter_parts.append(cel.equals("annotation_type", annotation_type.to_filter_str()))
+        if state:
+            filter_parts.append(cel.equals("state", state.to_filter_str()))
+        if assigned_to:
+            filter_parts.append(cel.equals("assignee", assigned_to))
```
wei-qlu, line 282: `assignee` filters on the user's name (azimuth maps it to users.name) but the docstring says user id

python/lib/sift_client/resources/annotations.py
```diff
@@ -0,0 +1,657 @@
...
+        log = await self._low_level_client.create_annotation_log(
+            annotation_id=annotation._id_or_error
+            if isinstance(annotation, Annotation)
+            else annotation,
+            kind=AnnotationLogKind.COMMENT,
+            comment=body,
+        )
+        return self._apply_client_to_instance(log)
+
+    async def record_assignment(self, annotation: str | Annotation, user: str) -> AnnotationLog:
```
wei-qlu: Drop `record_assignment` and `record_state`. They write a "resolved" or "assigned" entry to the feed without changing the annotation, and the server already writes those entries itself whenever resolve/flag/reopen/assign run, so the only thing they add is history that disagrees with the annotation itself 

Comments stay since that's intended

python/lib/sift_client/resources/annotations.py
```diff
@@ -0,0 +1,657 @@
...
+        log = await self._low_level_client.create_annotation_log(
+            annotation_id=annotation._id_or_error
+            if isinstance(annotation, Annotation)
+            else annotation,
+            kind=AnnotationLogKind.STATE_UPDATE,
+            state=state,
+        )
+        return self._apply_client_to_instance(log)
+
+    async def delete(self, annotation: str | Annotation, log: str | AnnotationLog) -> None:
```
wei-qlu: Drop `logs.delete`, nothing in Sift supports deleting logs, and we shouldn't expose it to users when it's own test says it doesn't work.

python/lib/sift_client/resources/annotations.py
```diff
@@ -0,0 +1,657 @@
...
+
+        Returns:
+            The created Annotation.
+        """
+        if isinstance(create, dict):
+            create = AnnotationCreate.model_validate(create)
+        created = await self._low_level_client.create_annotation(create=create)
+        return self._apply_client_to_instance(created)
+
+    async def create_review(
```
wei-qlu: Let's fold `create_review` and `create_phase` into `create`. Under the hood both just build an `AnnotationCreate` and call create with `annotation_type` set, but with renamed fields along the way (assign_to vs assign_to_user_id, channels vs linked_channels, run vs run_id), so users end up having to know two names for everything

```
client.annotations.create(AnnotationCreate(name="Spike", start_time=start, end_time=end,
    assets=["Nostromo"], linked_channels=[AnnotationLinkedChannel(channel_id=c.id_) for c in chans],
    assign_to_user_id=bob))
client.annotations.create_review("Spike", start, end, channels=chans, assign_to=bob)
client.annotations.create_phase("Burn", start, end, channels=chans)
```

Would become:

```
client.annotations.create(AnnotationCreate(name="Spike", start_time=start, end_time=end,
    linked_channels=chans, assign_to_user_id=bob))
client.annotations.create({"name": "Burn", "start_time": start, "end_time": end,
    "linked_channels": chans, "annotation_type": AnnotationType.PHASE})
```
Let `AnnotationCreate` accept `Channel` and `Asset` objects and move the asset-from-channels lookup from `_create_typed` into `create`

python/lib/sift_client/sift_types/annotation.py
```diff
@@ -0,0 +1,507 @@
...
+
+        Args:
+            user: The user ID to assign to.
+
+        Returns:
+            The updated annotation.
+        """
+        return self.update({"assigned_to_user_id": user})
+
+    def resolve(self) -> Annotation:
```
wei-qlu: Convention is that instance methods forward to the resource and `_update` (see archive right above). assign/resolve/flag/reopen restate the logic instead and duplicate `_set_state`, so let's have them call `self.client.annotations.resolve(annotation=self)` etc., which also deletes Annotation._set_state.

python/lib/sift_client/_internal/low_level_wrappers/annotations.py
```diff
@@ -0,0 +1,339 @@
...
+
+        Args:
+            annotation_id: The ID of the annotation to archive.
+
+        Returns:
+            The archived Annotation.
+        """
+        request = ArchiveAnnotationRequest(annotation_id=annotation_id)
+        await self._grpc_client.get_stub(AnnotationServiceStub).ArchiveAnnotation(request)
+        return await self.get_annotation(annotation_id)
```
wei-qlu: `ArchiveAnnotationRequest` and `UnarchiveAnnotationRequest` already return the updated annotation so the extra `get_annotation` here is redundant

wei-qlu (changes requested):

python/lib/sift_client/sift_types/annotation.py
```diff
@@ -0,0 +1,507 @@
...
+    start_time: datetime
+    end_time: datetime
+    annotation_type: AnnotationType
+    organization_id: str
+    created_date: datetime
+    modified_date: datetime
+    created_by_user_id: str
+    modified_by_user_id: str
+    tags: list[str]
+    asset_ids: list[str]
```
alexluck-sift, line 122: Add props that return the Assets and Run objects, created/updated user

python/lib/sift_client/sift_types/annotation.py
```diff
@@ -0,0 +1,507 @@
...
+    metadata: dict[str, str | float | bool]
+    is_archived: bool
+    pending: bool
+
+    # Optional fields
+    state: AnnotationState | None
+    run_id: str | None
+    assigned_to_user_id: str | None
+    created_by_rule_condition_version_id: str | None
+    legend_config: str | None
```
alexluck-sift, line 133: this is not user-useful, drop this field

python/lib/sift_client/sift_types/annotation.py
```diff
@@ -0,0 +1,507 @@
...
+    tags: list[str]
+    asset_ids: list[str]
+    linked_channels: list[AnnotationLinkedChannel]
+    metadata: dict[str, str | float | bool]
+    is_archived: bool
+    pending: bool
+
+    # Optional fields
+    state: AnnotationState | None
+    run_id: str | None
```
alexluck-sift: same as above comment, can this be a Run?

python/lib/sift_client/sift_types/annotation.py
alexluck-sift: The enum names do not actually correlated to what they are referred to in the application. Use the application nomenclature instead.

python/lib/sift_client/sift_types/annotation.py
```diff
@@ -0,0 +1,507 @@
...
+        Returns:
+            The updated annotation.
+        """
+        return self.update({"assigned_to_user_id": user})
+
+    def resolve(self) -> Annotation:
+        """Close out the review as resolved."""
+        return self._set_state(AnnotationState.RESOLVED)
+
+    def flag(self) -> Annotation:
```
alexluck-sift: A better pattern would be to call these `set_<state>` or something similar. That way they are more discoverable with autocomplete and grouped together in docs, etc.

python/lib/sift_client/sift_types/annotation.py
```diff
@@ -0,0 +1,507 @@
...
+        """Move to a review state, skipping the call if already there.
+
+        The server rejects a redundant state change with INVALID_ARGUMENT, so calling
+        `resolve` twice would fail without this.
+        """
+        if self.state is state:
+            return self
+        return self.update({"state": state})
+
+    def comment(self, text: str | list[AnnotationCommentElement]) -> AnnotationLog:
```
alexluck-sift: `add_comment`

python/lib/sift_client/sift_types/annotation.py
```diff
@@ -0,0 +1,507 @@
...
+    """Create model for Annotation.
+
+    Note that `assets` takes asset names, not IDs, and `tags` takes tag names.
+    """
+
+    name: str
+    start_time: datetime
+    end_time: datetime
+    annotation_type: AnnotationType = AnnotationType.DATA_REVIEW
+    assets: list[str] | None = None
```
alexluck-sift: asset and run should accept both Object and Id

python/lib/sift_client/sift_types/annotation.py
```diff
@@ -0,0 +1,507 @@
...
+    assign_to_user_id: str | None = None
+    organization_id: str | None = None
+
+    def _get_proto_class(self) -> type[CreateAnnotationRequestProto]:
+        return CreateAnnotationRequestProto
+
+    @model_validator(mode="after")
+    def _validate_state(self):
+        """Phase annotations have no review state; the server rejects one."""
+        if self.annotation_type is AnnotationType.PHASE and self.state is not None:
```
alexluck-sift: It may be nicer to have different models for the different types since they are very different in usage.

python/lib/sift_client/sift_types/annotation.py
```diff
@@ -0,0 +1,507 @@
...
+
+    Attributes:
+        text: The literal text, when this element is text.
+        user_id: The mentioned user's ID, when this element is a mention.
+        user_email: The mentioned user's email, when this element is a mention.
+    """
+
+    text: str | None = None
+    user_id: str | None = None
+    user_email: str | None = None
```
alexluck-sift: not a fan of this pattern - we have a User resource now. Can we instead use `user_id` and a user prop?

python/lib/sift_client/sift_types/annotation.py
```diff
@@ -0,0 +1,507 @@
...
+
+class AnnotationLog(BaseType[AnnotationLogProto, "AnnotationLog"]):
+    """One entry in an annotation's history: an assignment, a state change, or a comment."""
+
+    annotation_id: str
+    kind: AnnotationLogKind
+    created_date: datetime
+    modified_date: datetime
+    created_by_user_id: str
+    created_by_user_name: str
```
alexluck-sift: Same as above - use id and user resource instead of name, email, etc.

python/lib/sift_client/sift_types/annotation.py
```diff
@@ -0,0 +1,507 @@
...
+    OPEN = AnnotationStateProto.ANNOTATION_STATE_OPEN  # 1
+    FLAGGED = AnnotationStateProto.ANNOTATION_STATE_FLAGGED  # 2
+    RESOLVED = AnnotationStateProto.ANNOTATION_STATE_RESOLVED  # 3
+
+    def to_filter_str(self) -> str:
+        """Convert to the string used in CEL filters."""
+        return f"ANNOTATION_STATE_{self.name}"
+
+
+class AnnotationLinkedChannel(BaseModel):
```
alexluck-sift: I don't think we should use a new type for this. Simply reference the existing types.

`linked_channels: list[Channel | CalcualtedChannel | ....]`

python/lib/sift_client/resources/annotations.py
```diff
@@ -0,0 +1,657 @@
...
+        """
+        super().__init__(sift_client)
+        self._low_level_client = AnnotationsLowLevelClient(grpc_client=self.client.grpc_client)
+
+    async def list_(
+        self,
+        *,
+        annotation: str | Annotation | None = None,
+        # self ids
+        annotation_log_ids: list[str] | None = None,
```
alexluck-sift: Should be `annotation_logs: list[str | AnnotationLog]`

Do we think there is really a need for this without filtering by an annotation first?

python/lib/sift_client/resources/annotations.py
```diff
@@ -0,0 +1,657 @@
...
+            if isinstance(annotation, Annotation)
+            else annotation,
+            query_filter=query_filter or None,
+            order_by=order_by,
+            max_results=limit,
+            **({"page_size": page_size} if page_size is not None else {}),
+        )
+        return self._apply_client_to_instances(logs)
+
+    async def comment(
```
alexluck-sift: Same comment as on the type - `add_comment`

python/lib/sift_client/resources/annotations.py
alexluck-sift: +1

python/lib/sift_client/resources/annotations.py
```diff
@@ -0,0 +1,657 @@
...
+
+    async def list_(
+        self,
+        *,
+        name: str | None = None,
+        names: list[str] | None = None,
+        name_contains: str | None = None,
+        name_regex: str | re.Pattern | None = None,
+        # self ids
+        annotation_ids: list[str] | None = None,
```
alexluck-sift, line 183: should be `Annotation | str`

python/lib/sift_client/resources/annotations.py
```diff
@@ -0,0 +1,657 @@
...
+        tags: list[str] | list[Tag] | None = None,
+        metadata: dict[str, Any] | None = None,
+        # annotation specific
+        annotation_type: AnnotationType | None = None,
+        state: AnnotationState | None = None,
+        assigned_to: Any | str | None = None,
+        pending: bool | None = None,
+        assets: list[Asset] | list[str] | None = None,
+        runs: list[Run] | list[str] | None = None,
+        rule_ids: list[str] | None = None,
```
alexluck-sift: support Rule and Report objects

python/lib/sift_client/resources/annotations.py
```diff
@@ -0,0 +1,657 @@
...
+            The Annotation.
+        """
+        annotation = await self._low_level_client.get_annotation(annotation_id=annotation_id)
+        return self._apply_client_to_instance(annotation)
+
+    async def list_(
+        self,
+        *,
+        name: str | None = None,
+        names: list[str] | None = None,
```
alexluck-sift: We should combine these into a single arg (not just here but everywhere). 

e.g. `name: str | list[str] | None`

python/lib/sift_client/resources/annotations.py
```diff
@@ -0,0 +1,657 @@
...
+    async def batch_unarchive(self, annotations: list[str | Annotation]) -> None:
+        """Unarchive many annotations in one call.
+
+        Args:
+            annotations: The Annotations or annotation IDs to unarchive.
+        """
+        ids = [a._id_or_error if isinstance(a, Annotation) else a for a in annotations]
+        await self._low_level_client.batch_unarchive_annotations(annotation_ids=ids)
+
+    async def assign(self, annotation: str | Annotation, user: str) -> Annotation:
```
alexluck-sift: user should be `User | str`

naming: `assign_to_user` is much more precise

python/lib/sift_client/resources/annotations.py
```diff
@@ -0,0 +1,657 @@
...
+        """Move an annotation to a review state, skipping the call if already there.
+
+        The server rejects a redundant state change with INVALID_ARGUMENT. The state is
+        only known without a fetch when an Annotation was passed rather than an ID.
+        """
+        if isinstance(annotation, Annotation) and annotation.state is state:
+            return annotation
+        return await self.update(annotation, AnnotationUpdate(state=state))
+
+    async def resolve(self, annotation: str | Annotation) -> Annotation:
```
alexluck-sift: Same comments as on the type regarding function naming

python/lib/sift_client/resources/annotations.py
```diff
@@ -0,0 +1,657 @@
...
+        # tags and metadata
+        tags: list[str] | list[Tag] | None = None,
+        metadata: dict[str, Any] | None = None,
+        # annotation specific
+        annotation_type: AnnotationType | None = None,
+        state: AnnotationState | None = None,
+        assigned_to: Any | str | None = None,
+        pending: bool | None = None,
+        assets: list[Asset] | list[str] | None = None,
+        runs: list[Run] | list[str] | None = None,
```
alexluck-sift, line 200: Consider adding `annotations()` property on Run

Brandon-Shippy: https://github.com/sift-stack/sift/pull/792/commits/0b4f654371ad40de7fd61015dbdc95d1bc0a816f

python/lib/sift_client/resources/annotations.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/792/commits/0b4f654371ad40de7fd61015dbdc95d1bc0a816f

python/lib/sift_client/resources/annotations.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/792/commits/0b4f654371ad40de7fd61015dbdc95d1bc0a816f

python/lib/sift_client/resources/annotations.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/792/commits/0b4f654371ad40de7fd61015dbdc95d1bc0a816f

python/lib/sift_client/resources/annotations.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/792/commits/0b4f654371ad40de7fd61015dbdc95d1bc0a816f

python/lib/sift_client/resources/annotations.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/792/commits/0b4f654371ad40de7fd61015dbdc95d1bc0a816f

python/lib/sift_client/resources/annotations.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/792/commits/0b4f654371ad40de7fd61015dbdc95d1bc0a816f

python/lib/sift_client/resources/annotations.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/792/commits/0b4f654371ad40de7fd61015dbdc95d1bc0a816f

python/lib/sift_client/sift_types/annotation.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/792/commits/0b4f654371ad40de7fd61015dbdc95d1bc0a816f

python/lib/sift_client/sift_types/annotation.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/792/commits/0b4f654371ad40de7fd61015dbdc95d1bc0a816f

python/lib/sift_client/sift_types/annotation.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/792/commits/0b4f654371ad40de7fd61015dbdc95d1bc0a816f

python/lib/sift_client/sift_types/annotation.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/792/commits/0b4f654371ad40de7fd61015dbdc95d1bc0a816f

python/lib/sift_client/sift_types/annotation.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/792/commits/0b4f654371ad40de7fd61015dbdc95d1bc0a816f

python/lib/sift_client/sift_types/annotation.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/792/commits/0b4f654371ad40de7fd61015dbdc95d1bc0a816f

python/lib/sift_client/sift_types/annotation.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/792/commits/0b4f654371ad40de7fd61015dbdc95d1bc0a816f

python/lib/sift_client/sift_types/annotation.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/792/commits/0b4f654371ad40de7fd61015dbdc95d1bc0a816f

python/lib/sift_client/sift_types/annotation.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/792/commits/0b4f654371ad40de7fd61015dbdc95d1bc0a816f

python/lib/sift_client/sift_types/annotation.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/792/commits/0b4f654371ad40de7fd61015dbdc95d1bc0a816f

python/lib/sift_client/_internal/low_level_wrappers/annotations.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/792/commits/0b4f654371ad40de7fd61015dbdc95d1bc0a816f

python/lib/sift_client/sift_types/annotation.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/792/commits/0b4f654371ad40de7fd61015dbdc95d1bc0a816f

python/lib/sift_client/resources/annotations.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/792/commits/0b4f654371ad40de7fd61015dbdc95d1bc0a816f

python/lib/sift_client/resources/annotations.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/792/commits/0b4f654371ad40de7fd61015dbdc95d1bc0a816f

python/lib/sift_client/resources/annotations.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/792/commits/0b4f654371ad40de7fd61015dbdc95d1bc0a816f

python/lib/sift_client/resources/annotations.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/792/commits/0b4f654371ad40de7fd61015dbdc95d1bc0a816f

Brandon-Shippy: docstring for `assigned_to` now says it filters on the user's name

python/lib/sift_client/resources/annotations.py
Brandon-Shippy: dropped `logs.record_assignment` and `logs.record_state`

python/lib/sift_client/resources/annotations.py
Brandon-Shippy: folded `create_review` and `create_phase` into one `create()` that takes either `AnnotationCreate` or `PhaseCreate`

python/lib/sift_client/resources/annotations.py
Brandon-Shippy: i killed it

python/lib/sift_client/sift_types/annotation.py
Brandon-Shippy: instance now forwards to the resource, then _update. Annotation._set_state is deleted

python/lib/sift_client/_internal/low_level_wrappers/annotations.py
Brandon-Shippy: removed the `get_annotation` call

python/lib/sift_client/sift_types/annotation.py
Brandon-Shippy: added `assets`, `run`, `created_by`, `modified_by` and `assigned_to`

python/lib/sift_client/sift_types/annotation.py
Brandon-Shippy: its gone

python/lib/sift_client/sift_types/annotation.py
Brandon-Shippy: renamed to match open / failed / accepted

python/lib/sift_client/sift_types/annotation.py
Brandon-Shippy: renamed them to set_<>, i.e `set_open` `set_failed` `set_accepted`

python/lib/sift_client/sift_types/annotation.py
Brandon-Shippy: done

python/lib/sift_client/sift_types/annotation.py
Brandon-Shippy: assets takes str | Asset
run_id takes str | Run

python/lib/sift_client/sift_types/annotation.py
Brandon-Shippy: They are now split into both `AnnotationCreate` and `PhaseCreate` but are folded under `create()`

python/lib/sift_client/sift_types/annotation.py
Brandon-Shippy: makes sense, done

python/lib/sift_client/sift_types/annotation.py
Brandon-Shippy: email etc is gone now

python/lib/sift_client/sift_types/annotation.py
Brandon-Shippy: AnnotationLinkedChannel deleted.
writes do `list[Channel | CalculatedChannel]`

python/lib/sift_client/resources/annotations.py
Brandon-Shippy: annotation is now required on `logs.list_` (`annotation_log_ids` is now `annotation_logs`)

python/lib/sift_client/resources/annotations.py
Brandon-Shippy: updated

python/lib/sift_client/resources/annotations.py
Brandon-Shippy: done

python/lib/sift_client/resources/annotations.py
Brandon-Shippy: rule_ids and report_ids became rules and replorts which both accept objects or ids

python/lib/sift_client/resources/annotations.py
Brandon-Shippy: all have been collapsed

i noticed a pattern on resources wehere we have asset and assets. are these things we want merged as well. it is in the asset resource

python/lib/sift_client/resources/annotations.py
Brandon-Shippy: i agree, did it

python/lib/sift_client/resources/annotations.py
Brandon-Shippy: now matches the others

python/lib/sift_client/resources/annotations.py
Brandon-Shippy: add ed `Run.annotations` property

python/lib/sift_client/sift_types/annotation.py
```diff
@@ -0,0 +1,546 @@
...
+
+    `assets` takes Assets or asset names, and `tags` takes tag names. Omit `assets` and
+    the asset is taken from `linked_channels`.
+    """
+
+    name: str
+    start_time: datetime
+    end_time: datetime
+    assets: list[str | Asset] | None = None
+    run_id: str | Run | None = None
```
alexluck-sift: should be `run`

python/lib/sift_client/resources/annotations.py
alexluck-sift: Doesn't look fixed?

---
_Generated by [Claude Code](https://claude.ai/code)_

python/lib/sift_client/sift_types/annotation.py
alexluck-sift: I meant legend config

---
_Generated by [Claude Code](https://claude.ai/code)_

python/lib/sift_client/sift_types/annotation.py
alexluck-sift: Create works, but `AnnotationUpdate(linked_channels=...)` is silently dropped: `exclude=True` (annotation.py:301-303) keeps it out of the mask, and `update_annotation` never adds it back, though the `update` docstring says it's replaced. Add it to the proto and mask in `to_proto_with_mask`, as `ReportTemplateUpdate` does (report_template.py:285-290). On read, line 224 passes calculated-channel version IDs as `calculated_channel_id`, so resolve them with one `GetCalculatedChannelVersions` call instead, and fix test_annotations.py:300, since `Channel` has no `channel_id`.

---
_Generated by [Claude Code](https://claude.ai/code)_

python/lib/sift_client/sift_types/annotation.py
alexluck-sift: The split is in, but `annotation_type` is still settable on both models (annotation.py:373, 399), so `AnnotationCreate(annotation_type=AnnotationType.PHASE, state=...)` gets all the way to the server, which rejects a phase with a state. Add a pydantic check so it fails validation instead: pin the type per class (`Literal[AnnotationType.PHASE]` on `PhaseCreate`) or bring back a `_validate_state` model validator. Then move `test_create_phase` to `PhaseCreate` and fix `test_phase_model_has_no_usable_state`, which raises KeyError at head.

---
_Generated by [Claude Code](https://claude.ai/code)_

python/lib/sift_client/sift_types/annotation.py
alexluck-sift: `assets` now takes `Asset` objects, but a string is still read as an asset name (annotation.py:336-353), while `list_(assets=...)` reads it as an ID (resources/annotations.py:285-287). Strings should be asset IDs here, like for any other Sift resource, since a name can change. Resolve the IDs to names in `create()` with the same `assets.list_(asset_ids=...)` call that `_assets_for_channels` already makes.

---
_Generated by [Claude Code](https://claude.ai/code)_

python/lib/sift_client/sift_types/annotation.py
alexluck-sift: `user_email` is gone, but there's still no user prop: `AnnotationCommentElement` (annotation.py:448-483) is a plain `BaseModel`, so reading a mention means calling `client.users.get` by hand, and writing one rejects a `User`. Accept `str | User` for `user_id`, and resolve mentions through a property on `AnnotationLog`, which has the client.

---
_Generated by [Claude Code](https://claude.ai/code)_

python/lib/sift_client/sift_types/annotation.py
alexluck-sift: The enums read right now. `test_record_state` (test_annotations.py:248-253) still uses `AnnotationLogState.FLAGGED` and the removed `logs.record_state`, so it fails the first time anyone runs it; delete it. The `set_*` docstrings at resources/annotations.py:469-494 still say resolve, flag and reopen, so reword them to accepted, failed and open and regenerate the stubs.

---
_Generated by [Claude Code](https://claude.ai/code)_

python/lib/sift_client/resources/annotations.py
alexluck-sift: The objects work now. The `list_` docstring still lists `rule_ids` and `report_ids` at resources/annotations.py:237-238, and the stub repeats them, so hover docs show two arguments that don't exist. Drop both lines and regenerate the stubs.

---
_Generated by [Claude Code](https://claude.ai/code)_

python/lib/sift_client/resources/annotations.py
alexluck-sift: Yes, and let's make it library-wide: please file a ticket to move every `list_` (and `_build_name_cel_filters`) onto the single `name: str | list[str]` shape and settle pairs like `asset`/`assets` the same way, and link it here. Also, a tuple or set in `name` falls through both checks at resources/annotations.py:256-257 and silently drops the filter, so I'd move that split into `_build_name_cel_filters` and treat any non-string iterable as a list.

---
_Generated by [Claude Code](https://claude.ai/code)_

python/lib/sift_client/sift_types/annotation.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/792/commits/a0c97fcc41eea72847a973a078142c85c0390d7f

python/lib/sift_client/sift_types/annotation.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/792/commits/a0c97fcc41eea72847a973a078142c85c0390d7f

yeah that makes more sense here

python/lib/sift_client/resources/annotations.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/792/commits/a0c97fcc41eea72847a973a078142c85c0390d7f

should be fixed, the docstring was still listed

python/lib/sift_client/sift_types/annotation.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/792/commits/a0c97fcc41eea72847a973a078142c85c0390d7f

now sets and adds linked_channels to the mask

Brandon-Shippy: `list_versions` filters on `calculated_channel_version_id` now

Brandon-Shippy: test needed to use `channel_id` like you said

python/lib/sift_client/sift_types/annotation.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/792/commits/a0c97fcc41eea72847a973a078142c85c0390d7f

a mismatched type fails validation now

python/lib/sift_client/sift_types/annotation.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/792/commits/a0c97fcc41eea72847a973a078142c85c0390d7

create() resolves them to names using `_asset_names`

python/lib/sift_client/sift_types/annotation.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/792/commits/a0c97fcc41eea72847a973a078142c85c0390d7f

`AnnotationLog.mentioned_users` resolves both str and User

python/lib/sift_client/sift_types/annotation.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/792/commits/a0c97fcc41eea72847a973a078142c85c0390d7f

deleted `test_record_state`

`set_*` docstrings use the proper words: Accepted/Failed/Open

python/lib/sift_client/resources/annotations.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/792/commits/a0c97fcc41eea72847a973a078142c85c0390d7f

now any non-string iterable can be treated as a list

PR #793, opened 2026-09-17
python(feat): udf resource

Brandon-Shippy:
# Description
Adds a user-defined functions resource to `sift_client`. A function is a reusable, versioned expression that calculated channels and rules call by name.

- **`UserDefinedFunctionsAPIAsync`** with `get`, `list_`, `find`, `create`, `update`, `archive`, `unarchive`. Wired up as `client.user_defined_functions`
- **`UserDefinedFunctionVersionsAPIAsync`** nested as `client.user_defined_functions.versions`
- **`UserDefinedFunctionsLowLevelClient`** wraps `UserDefinedFunctionService`; the high-level client has zero proto imports
- **Pydantic models** in `sift_types/user_defined_function.py`
- **`validate(...)`** checks an expression without saving it
- **`dependents(...)`** returns the rules and calculated channels that would break on a change
- **`sync(...)`** creates or updates a list of definitions by name, so a definitions file is rerunnable

  # Validation
  - Verified against a local stack, including a calculated channel and a rule that call a function by name


python/lib/sift_client/resources/user_defined_functions.py
```diff
@@ -0,0 +1,330 @@
...
+        function_id = (
+            function._id_or_error if isinstance(function, UserDefinedFunction) else function
+        )
+        if isinstance(update, dict):
+            update = UserDefinedFunctionUpdate.model_validate(update)
+        update.resource_id = function_id
+        updated = await self._low_level_client.update_function(update)
+        return self._apply_client_to_instance(updated)
+
+    async def sync(
```
wei-qlu, line 241: I'd drop `sync`, or if you want to keep it, rename it to something like `create_or_update` and look up by names instead of listing the whole org, skip archived matches, carry `user_notes` through, and document that the server blocks input, output type, and name changes on in-use functions, so a rerun can't promise Sift will match the file

I think we should drop it, since anyone who needs it can write `find` then `create` or `update` in a few lines and see exactly where a partial failure stopped

python/lib/sift_client/sift_types/user_defined_function.py
```diff
@@ -0,0 +1,253 @@
...
+        updated = self.client.user_defined_functions.unarchive(function=self)
+        self._update(updated)
+        return self
+
+
+class UserDefinedFunctionBase(ModelCreateUpdateBase):
+    """Base class for UserDefinedFunction create and update models."""
+
+    description: str | None = None
+    function_inputs: list[FunctionInput] | None = None
```
wei-qlu, line 173: The server rejects an empty `function_inputs` on create and validate, so this should be required on `UserDefinedFunctionCreate` and `validate()`

python/lib/sift_client/resources/user_defined_functions.py
```diff
@@ -0,0 +1,330 @@
...
+
+        Returns:
+            Whether the expression compiles, and its output type or error.
+        """
+        return await self._low_level_client.validate_function(
+            expression=expression, function_inputs=function_inputs
+        )
+
+    async def dependents(
+        self, function: str | UserDefinedFunction, *, version_id: str | None = None
```
wei-qlu, line 289: if `version_id` is passed, `function` gets ignored. Is this intentional?

python/lib/sift_client/_internal/low_level_wrappers/user_defined_functions.py
```diff
@@ -0,0 +1,349 @@
...
+        else:
+            request_kwargs["user_defined_function_version_id"] = version_id
+
+        request = GetUserDefinedFunctionDependentsRequest(**request_kwargs)
+        response = await self._grpc_client.get_stub(
+            UserDefinedFunctionServiceStub
+        ).GetUserDefinedFunctionDependents(request)
+        response = cast("GetUserDefinedFunctionDependentsResponse", response)
+
+        return FunctionDependents(
```
wei-qlu, line 345: Was there a reason to reduce these to IDs? The response returns full messages and `_from_proto` exists for all three types

python/lib/sift_client/resources/user_defined_functions.py
```diff
@@ -0,0 +1,330 @@
...
+        Args:
+            function_id: The ID of the function.
+
+        Returns:
+            The UserDefinedFunction.
+        """
+        function = await self._low_level_client.get_function(function_id=function_id)
+        return self._apply_client_to_instance(function)
+
+    async def list_(
```
wei-qlu: Let's add `names` next to `name` and a `metadata` dict, same as the other resources, so people can fetch a known set of functions in one call and filter by labels

python/lib/sift_client/resources/user_defined_functions.py
```diff
@@ -0,0 +1,330 @@
...
+            create = UserDefinedFunctionCreate.model_validate(create)
+        created = await self._low_level_client.create_function(create=create)
+        return self._apply_client_to_instance(created)
+
+    async def update(
+        self,
+        function: str | UserDefinedFunction,
+        update: UserDefinedFunctionUpdate | dict,
+    ) -> UserDefinedFunction:
+        """Update a function. This creates a new version.
```
wei-qlu, line 223: Two things here. Archives return the current version, so something like `Changes to the expression, inputs, description, or metadata create a new version.` would be more accurate. And let's add a `user_notes` kwarg like calculated channels have, threaded into the wrapper and set on the proto after `to_proto_with_mask()`. The server reads it from the body, and right now there's no way to leave a note on a version

wei-qlu: One catch for `user_notes`, the server treats an empty note as a change, so when none is given, send the current version's note rather than empty, otherwise every update clears it and bumps the version. `Rules` update already fetches the object when passed an ID, so the same pattern works here

wei-qlu (changes requested):

python/lib/sift_client/resources/user_defined_functions.py
```diff
@@ -0,0 +1,330 @@
...
+        Returns:
+            The function at that version.
+        """
+        version = await self._low_level_client.get_version(version_id=version_id)
+        return self._apply_client_to_instance(version)
+
+    async def list_(
+        self,
+        *,
+        function: str | UserDefinedFunction | None = None,
```
alexluck-sift, line 58: missing a lot of other filters we should support here.

python/lib/sift_client/resources/user_defined_functions.py
```diff
@@ -0,0 +1,330 @@
...
+        return self._apply_client_to_instance(function)
+
+    async def list_(
+        self,
+        *,
+        name: str | None = None,
+        name_contains: str | None = None,
+        name_regex: str | re.Pattern | None = None,
+        # self ids
+        function_ids: list[str] | None = None,
```
alexluck-sift, line 142: object and Id support.

A lot of missing filters here?

python/lib/sift_client/resources/user_defined_functions.py
alexluck-sift: I agree we should remove this. It is not idiomatic with the rest of the types.

python/lib/sift_client/resources/user_defined_functions.py
```diff
@@ -0,0 +1,330 @@
...
+        for definition in definitions:
+            current = existing.get(definition.name)
+            if current is None:
+                results.append(await self.create(definition))
+                continue
+            changes = definition.model_dump(exclude_unset=True, exclude={"name"})
+            results.append(await self.update(current, changes))
+        return results
+
+    async def validate(
```
alexluck-sift, line 272: `validate_expression` - more descriptive naming

python/lib/sift_client/resources/user_defined_functions.py
```diff
@@ -0,0 +1,330 @@
...
+            function_inputs: The inputs the expression refers to.
+
+        Returns:
+            Whether the expression compiles, and its output type or error.
+        """
+        return await self._low_level_client.validate_function(
+            expression=expression, function_inputs=function_inputs
+        )
+
+    async def dependents(
```
alexluck-sift, line 288: `get_where_used` is more clear IMO

python/lib/sift_client/resources/user_defined_functions.py
alexluck-sift, line 289: `version: UserDefinedFunctionVersion | str`

Also since version is fully defined, it should make passing the function itself optional

alexluck-sift, line 289: nit on naming: `user_defined_function` instead of `function`

python/lib/sift_client/sift_types/user_defined_function.py
```diff
@@ -0,0 +1,253 @@
...
+
+    Attributes:
+        identifier: How the expression refers to this input, such as `$1`.
+        data_type: The type the input accepts.
+        constant: Whether the input is a literal rather than a channel.
+    """
+
+    identifier: str
+    data_type: FunctionDataType = FunctionDataType.NUMERIC
+    constant: bool = False
```
alexluck-sift, line 54: This naming is confusing. It isn't a constant since that would imply it doesn't change, rather this is a value provided by the caller. `scalar` or something may make more sense.

python/lib/sift_client/sift_types/user_defined_function.py
```diff
@@ -0,0 +1,253 @@
...
+            archived_date=(
+                proto.archived_date.ToDatetime(tzinfo=timezone.utc)
+                if proto.HasField("archived_date")
+                else None
+            ),
+            _client=sift_client,
+        )
+
+    @property
+    def versions(self) -> list[UserDefinedFunction]:
```
alexluck-sift, line 139: Please add a `latest_version`

python/lib/sift_client/sift_types/user_defined_function.py
```diff
@@ -0,0 +1,253 @@
...
+
+    Each save produces a new version. `id_` identifies the function; `version_id`
+    identifies this particular version.
+    """
+
+    # Required fields
+    name: str
+    description: str
+    expression: str
+    version: int
```
alexluck-sift, line 86: This is pretty confusing and repeats some of the same sins of say Channels.

Instead of a single object, we should have two:
* User Defined Function
* User Defined Function Version

The User Defined Function should contain a reference to the latest version, but be version agnostic

python/lib/sift_client/sift_types/user_defined_function.py
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+
+class UserDefinedFunctionCreate(
+    UserDefinedFunctionBase, ModelCreate[CreateUserDefinedFunctionRequestProto]
+):
+    """Create model for UserDefinedFunction."""
+
+    name: str
+    expression: str
```
alexluck-sift, line 196: Should be documented that this is a cel expression

python/lib/sift_client/sift_types/user_defined_function.py
```diff
@@ -0,0 +1,253 @@
...
+    user_notes: str | None = None
```
alexluck-sift, line 197: Rename this "change_notes" or something more clear what it is doing

python/lib/sift_client/sift_types/user_defined_function.py
```diff
@@ -0,0 +1,253 @@
...
+class UserDefinedFunctionValidation(BaseModel):
+    """The result of checking an expression before saving it.
+
+    Attributes:
+        valid: Whether the expression compiles.
+        error: Why it failed, when it is not valid.
+        output_type: The type the expression returns, when it is valid.
+    """
+
+    valid: bool
```
alexluck-sift, line 232: `is_valid`

python/lib/sift_client/sift_types/user_defined_function.py
```diff
@@ -0,0 +1,253 @@
...
+    error: str | None = None
+    output_type: FunctionDataType | None = None
+
+
+class FunctionDependents(BaseModel):
+    """What would break if a function changed.
```
alexluck-sift, line 238: Is that what this represents? Why would things necessarily break?

python/lib/sift_client/sift_types/user_defined_function.py
```diff
@@ -0,0 +1,253 @@
...
+
+    Attributes:
+        function_ids: IDs of other functions that call it.
+        calculated_channel_ids: IDs of calculated channels that use it.
+        rule_ids: IDs of rules that use it.
+    """
+
+    function_ids: list[str] = []
+    calculated_channel_ids: list[str] = []
```
alexluck-sift, line 247: add properties to return the objects

alexluck-sift (changes requested):

python/lib/sift_client/resources/user_defined_functions.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/793/commits/2546dd6bb8d3ea0be987eab449bb6127e08aaca9

dropped

python/lib/sift_client/sift_types/user_defined_function.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/793/commits/2546dd6bb8d3ea0be987eab449bb6127e08aaca9

made it required on `UserDefinedFunctionCreate` and `validate_expression`

python/lib/sift_client/resources/user_defined_functions.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/793/commits/2546dd6bb8d3ea0be987eab449bb6127e08aaca9

nope. now raises if you pass neither or both

python/lib/sift_client/_internal/low_level_wrappers/user_defined_functions.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/793/commits/2546dd6bb8d3ea0be987eab449bb6127e08aaca9

`FunctionDependants` turned into `FunctionUsage` now returns the real objects. `UserDefinedFunction`, `CalculatedChannel`, etc

python/lib/sift_client/resources/user_defined_functions.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/793/commits/2546dd6bb8d3ea0be987eab449bb6127e08aaca9

docstring update. kwarg added

reads the function first then carries forward the current version notee when `change_notes` omitted, an update should no longer clear it

python/lib/sift_client/resources/user_defined_functions.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/793/commits/2546dd6bb8d3ea0be987eab449bb6127e08aaca9

renamed

python/lib/sift_client/resources/user_defined_functions.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/793/commits/2546dd6bb8d3ea0be987eab449bb6127e08aaca9

renamed

python/lib/sift_client/resources/user_defined_functions.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/793/commits/2546dd6bb8d3ea0be987eab449bb6127e08aaca9

now optional, and what it takes updated

Brandon-Shippy: https://github.com/sift-stack/sift/pull/793/commits/2546dd6bb8d3ea0be987eab449bb6127e08aaca9

renamed

python/lib/sift_client/sift_types/user_defined_function.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/793/commits/2546dd6bb8d3ea0be987eab449bb6127e08aaca9

renamed

python/lib/sift_client/sift_types/user_defined_function.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/793/commits/2546dd6bb8d3ea0be987eab449bb6127e08aaca9

added it onto `UserDefinedFunction`

python/lib/sift_client/sift_types/user_defined_function.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/793/commits/2546dd6bb8d3ea0be987eab449bb6127e08aaca9

split it up

python/lib/sift_client/sift_types/user_defined_function.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/793/commits/2546dd6bb8d3ea0be987eab449bb6127e08aaca9

updated the documentation

python/lib/sift_client/sift_types/user_defined_function.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/793/commits/2546dd6bb8d3ea0be987eab449bb6127e08aaca9

renamed to `change_notes`

python/lib/sift_client/sift_types/user_defined_function.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/793/commits/2546dd6bb8d3ea0be987eab449bb6127e08aaca9

renamed

python/lib/sift_client/sift_types/user_defined_function.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/793/commits/2546dd6bb8d3ea0be987eab449bb6127e08aaca9

added the properties

python/lib/sift_client/resources/user_defined_functions.py
Brandon-Shippy: same as above, will be address server-side in a ticket

python/lib/sift_client/resources/user_defined_functions.py
Brandon-Shippy: Making a ticket for this, blocked on the proto

python/lib/sift_client/sift_types/user_defined_function.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/793/commits/2546dd6bb8d3ea0be987eab449bb6127e08aaca9

changed the framing to be more like "where is a function used"

PR #795, opened 2026-09-17
python(feat): campaigns resource

Brandon-Shippy:
# Description
Adds a campaigns resource to `sift_client`. A campaign groups the reports from a series of runs, so a whole test program can be reviewed at once.

- **`CampaignsAPIAsync`** with the usual `get`, `list_`, `find`, `create`, `update`, `archive`, `unarchive`, plus `get_or_create`. Wired up as
`client.campaigns`
- **`CampaignsLowLevelClient`** wraps `CampaignService`; the high-level client has zero proto imports
- **Pydantic models** in `sift_types/campaign.py`
- **`add_runs(...)` and `add_reports(...)`** grow a campaign without replacing its report list
- **`report_summaries(...)`** returns the per-report rule rollup; `Campaign.report_summaries()` and `Campaign.runs` read it back from the campaign
- **`RunCreate.create_default_report`** so a run generates the report a campaign holds

# Validation
- Verified against a local stack, including a multi-run test program collected into one campaign


python/lib/sift_client/resources/campaigns.py
```diff
@@ -0,0 +1,342 @@
...
+        merged = existing_ids + [
+            report_id
+            for report_id in (self._report_id(r) for r in reports)
+            if report_id not in existing_ids
+        ]
+        return await self.update(
+            current, CampaignUpdate(reports=[CampaignReport(report_id=r) for r in merged])
+        )
+
+    async def add_runs(self, campaign: str | Campaign, runs: list[Run] | list[str]) -> Campaign:
```
wei-qlu: `create(runs=...)` adds every report a run has, but `add_runs` adds only the default report and raises without one, so users will assume they match and silently lose reports. List reports with `cel.in_("run_id", run_ids)` and pass them to `add_reports` (return early on an empty `run_ids`, since an empty `cel.in_` matches everything), then drop the `create_default_report` wording

python/lib/sift_client/sift_types/campaign.py
```diff
@@ -0,0 +1,251 @@
...
+        num_annotations: Annotations across the report's rules.
+        num_passed_rules: Rules that never triggered.
+        num_accepted_rules: Rules whose annotations are all accepted.
+        num_failed_rules: Rules with any failed annotation.
+        num_open_rules: Rules with open annotations and none failed.
+    """
+
+    report_id: str
+    report_name: str = ""
+    num_annotations: int = 0
```
alexluck-sift, line 50: can we make this cleaner by moving this into a structure like:

`rule_statistics: passed_count: .., open_count}` etc

python/lib/sift_client/sift_types/campaign.py
```diff
@@ -0,0 +1,251 @@
...
+
+def _campaign_report_to_proto(**kwargs) -> CampaignReportProto:
+    """Convert a campaign report dict (from model_dump) into its proto form.
+
+    Only `report_id` is sent; the name and counts are output only.
+    """
+    return CampaignReportProto(report_id=kwargs["report_id"])
+
+
+def tag_names(tags: list[str] | list[Tag] | None) -> list[str]:
```
alexluck-sift, line 77: Seems like an odd place for this function

python/lib/sift_client/sift_types/campaign.py
```diff
@@ -0,0 +1,251 @@
...
+            archived_date=(
+                proto.archived_date.ToDatetime(tzinfo=timezone.utc)
+                if proto.HasField("archived_date")
+                else None
+            ),
+            _client=sift_client,
+        )
+
+    @property
+    def report_ids(self) -> list[str]:
```
alexluck-sift, line 139: I don't think we need this - a user can do the list comprehension themselves if they really need this.

python/lib/sift_client/sift_types/campaign.py
```diff
@@ -0,0 +1,251 @@
...
+        """The IDs of the reports in this campaign."""
+        return [r.report_id for r in self.reports]
+
+    @property
+    def resolved_reports(self) -> list[Report]:
```
alexluck-sift, line 144: Should this just be `reports` and the existing `reports` to `report_summaries`

python/lib/sift_client/sift_types/campaign.py
```diff
@@ -0,0 +1,251 @@
...
+from sift_client.sift_types.tag import Tag
+from sift_client.util.metadata import metadata_dict_to_proto, metadata_proto_to_dict
+
+if TYPE_CHECKING:
+    from sift_client.client import SiftClient
+    from sift_client.sift_types.report import Report
+    from sift_client.sift_types.run import Run
+
+
+class CampaignReport(BaseModel):
```
alexluck-sift, line 33: `CampaignReportSummary` would be more descriptive

python/lib/sift_client/sift_types/campaign.py
```diff
@@ -0,0 +1,251 @@
...
+        """Fetch the Runs behind this campaign's reports. Reports with no run are skipped."""
+        run_ids = [r.run_id for r in self.resolved_reports if r.run_id]
+        if not run_ids:
+            return []
+        return self.client.runs.list_(run_ids=run_ids)
+
+    def report_summaries(self) -> list[CampaignReport]:
+        """Get this campaign's reports with their rule counts, ordered to match `report_ids`.
+
+        The service returns them in no fixed order, so this would otherwise vary.
```
alexluck-sift, line 161: can't we just reorder them when converting from the proto? That seems better than having two "sources" of the same info.

python/lib/sift_client/sift_types/campaign.py
```diff
@@ -0,0 +1,251 @@
...
+class CampaignUpdate(CampaignBase, ModelUpdate[CampaignProto]):
+    """Update model for Campaign.
+
+    `reports` replaces the campaign's report list. Prefer `campaigns.add_reports` or
+    `campaigns.add_runs`, which read the current list first. `tags` is unordered, so it
+    may come back in a different order.
+    """
+
+    name: str | None = None
+    reports: list[CampaignReport] | None = None
```
alexluck-sift, line 233: this seems like it should be `list[Report] | list[str]`. The `CampaignReport` is return only and is intended to be used for provide stats and avoid pagination/grpc message limit size issues only.

python/lib/sift_client/sift_types/run.py
```diff
@@ -217,6 +217,8 @@ class RunCreate(RunBase, ModelCreate[CreateRunRequestProto]):
     name: str
     client_key: str | None = None
     organization_id: str | None = None
+    create_default_report: bool | None = None
```
alexluck-sift, line 220: Do we really want to support this? Feels like a bodge and not first class. Would prefer a different method.

python/lib/sift_client/sift_types/campaign.py
```diff
@@ -0,0 +1,251 @@
...
+    """
+    return CampaignReportProto(report_id=kwargs["report_id"])
+
+
+def tag_names(tags: list[str] | list[Tag] | None) -> list[str]:
+    """Reduce Tags or tag names to plain names."""
+    return [tag.name if isinstance(tag, Tag) else tag for tag in tags or []]
+
+
+class Campaign(BaseType[CampaignProto, "Campaign"]):
```
alexluck-sift, line 82: Some helpers around "add_reports`, etc. would be nice.

python/lib/sift_client/resources/campaigns.py
```diff
@@ -0,0 +1,342 @@
...
+            return campaigns[0]
+        return None
+
+    async def create(
+        self,
+        create: CampaignCreate | dict,
+        *,
+        reports: list[Report] | list[str] | None = None,
+        runs: list[Run] | list[str] | None = None,
+        from_campaign: str | Campaign | None = None,
```
alexluck-sift, line 172: can this just be "campaign"?

python/lib/sift_client/resources/campaigns.py
```diff
@@ -0,0 +1,342 @@
...
+            from_run_ids=(
+                [r._id_or_error if isinstance(r, Run) else r for r in runs] if runs else None
+            ),
+            from_campaign_id=(
+                from_campaign._id_or_error if isinstance(from_campaign, Campaign) else from_campaign
+            ),
+        )
+        return self._apply_client_to_instance(created)
+
+    async def get_or_create(self, create: CampaignCreate | dict) -> Campaign:
```
alexluck-sift, line 202: This is generally something we should avoid. Any specific reason we have this for campaigns?

python/lib/sift_client/resources/campaigns.py
```diff
@@ -0,0 +1,342 @@
...
+            The updated Campaign.
+        """
+        campaign_id = campaign._id_or_error if isinstance(campaign, Campaign) else campaign
+        if isinstance(update, dict):
+            update = CampaignUpdate.model_validate(update)
+        update.resource_id = campaign_id
+        updated = await self._low_level_client.update_campaign(update)
+        return self._apply_client_to_instance(updated)
+
+    async def add_reports(
```
alexluck-sift, line 244: `add_reports_to_campaign`... etc

alexluck-sift (changes requested):

python/lib/sift_client/sift_types/campaign.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/795/commits/6d84852edd4936d3cb4e77e7be112ae84f2712a8

added RuleStatistics with annotations, passed, accepted, failed, open

python/lib/sift_client/sift_types/campaign.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/795/commits/6d84852edd4936d3cb4e77e7be112ae84f2712a8

moved to be next to the models

python/lib/sift_client/sift_types/campaign.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/795/commits/6d84852edd4936d3cb4e77e7be112ae84f2712a8

removed

python/lib/sift_client/sift_types/campaign.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/795/commits/6d84852edd4936d3cb4e77e7be112ae84f2712a8

renamed

python/lib/sift_client/sift_types/campaign.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/795/commits/6d84852edd4936d3cb4e77e7be112ae84f2712a8

renamed

python/lib/sift_client/sift_types/campaign.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/795/commits/6d84852edd4936d3cb4e77e7be112ae84f2712a8

re-sorted it

python/lib/sift_client/sift_types/campaign.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/795/commits/6d84852edd4936d3cb4e77e7be112ae84f2712a8

CampaignUpdate.reports takes Reports or IDs now

python/lib/sift_client/sift_types/campaign.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/795/commits/6d84852edd4936d3cb4e77e7be112ae84f2712a8

Added Campaign.add_reports and Campaign.add_runs

python/lib/sift_client/resources/campaigns.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/795/commits/6d84852edd4936d3cb4e77e7be112ae84f2712a8

renamed

python/lib/sift_client/resources/campaigns.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/795/commits/6d84852edd4936d3cb4e77e7be112ae84f2712a8

nope youre right, we should avoid this fully

i got rid of it

python/lib/sift_client/resources/campaigns.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/795/commits/66cfceba17130973a54aef27b7ce56702105f84f

changed

python/lib/sift_client/sift_types/run.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/795/commits/2acf5af558d61fb67777223fe6417a4eb2775cbf

This is deleted

python/lib/sift_client/sift_types/campaign.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/795/commits/2acf5af558d61fb67777223fe6417a4eb2775cbf

i shrank it down and just kept add_runs due to the relation of campaigns not storing runs

python/lib/sift_client/resources/campaigns.py
Brandon-Shippy: https://github.com/sift-stack/sift/pull/795/commits/2acf5af558d61fb67777223fe6417a4eb2775cbf

added `add_reports_to_campaign`

PR #797, opened 2026-09-21
rust(feat): update_artifact and composed artifact reads

evan-sift:
## Description

Adds an `update_artifact` tool and reads artifacts as a container plus a version.

`update_artifact` changes only the fields it is given; everything else carries forward, so a title edit no longer has to resend the payload. `create_artifact` now only creates, and drops `authoring_kind` in favour of `created_via`.

Artifact tags are renumbered while the API still has no callers to keep compatible.

## Verification

n/a


lineville (approved):
comments are a bit verbose, but otherwise lgtm

rust/crates/sift_mcp/src/service/artifacts/mod.rs
```diff
@@ -255,80 +311,151 @@ impl ArtifactService {
...
+            created,
+            uploader,
+            file_path,
+            "created",
+            "create the artifact again",
+        )
+        .await
+    }
+
+    pub async fn update_artifact(
```
bijan-sift, line 348: Does this need to be in [TOOL_FEATURE_FLAGS](https://github.com/sift-stack/sift/blob/9cd170e6323a12a8691cca2a0c9c28c4b312c23d/rust/crates/sift_mcp/src/feature_flags.rs#L9)?

